PR #48, opened 2024-09-06
feat: pcb_keepout added

imrishabh18:
Preview:
https://pcb-viewer-dahpoxxe0-tscircuit.vercel.app/?path=/story/keepout--keepout-rect-example


lib/components/primitive-components/Keepout.ts
```diff
@@ -0,0 +1,56 @@
...
+    if (props.shape === "circle") {
+      pcb_keepout = db.pcb_keepout.insert({
+        layer: ["top"],
+        shape: "circle",
+        // @ts-ignore: no idea why this is triggering
+        radius: props.radius,
+        center: {
+          x: position.x,
+          y: position.y,
+        },
```
seveibar, line 38: the rect vs circle should not have their centers defined differently, we should fix this in soup before encoding it into core

seveibar: https://github.com/tscircuit/circuit-json/issues/40

seveibar (changes requested):
looks good, but we should resolve the soup definition issues. IMO it is always worth having API repos like @tscircuit/soup and @tscircuit/props PRs reviewed because as soon as we realize something is wrong in the api design we generally shouldn't merge anything that uses the the deprecated api design

lib/components/primitive-components/Keepout.ts
```diff
@@ -0,0 +1,56 @@
+import { PrimitiveComponent } from "../base-components/PrimitiveComponent"
+import { pcbKeepoutProps } from "@tscircuit/props"
+import type { RenderPhaseFn } from "../base-components/Renderable"
+import type { PCBKeepout } from "@tscircuit/soup"
+import { decomposeTSR } from "transformation-matrix"
+
+export class KeepOut extends PrimitiveComponent<typeof pcbKeepoutProps> {
```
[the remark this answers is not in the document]
seveibar: (this is the google convention vs the microsoft convention for camelCase)

imrishabh18:
@seveibar For keepout the layer value can be both `top` and `bottom`, right?

seveibar (approved):

PR #85, opened 2024-09-13
now supports fabrication note path/text color prop

ShiboSoftwareDev:


seveibar (approved):

ShiboSoftwareDev:
@seveibar what's with the size check?

seveibar:
No idea, but this pr shouldnt change thr size anyway. It seems like it could be an issue with the workflow not supporting whatever we're supplying as the GITHUB_TOKEN

Its worth fixing this in the future but we can merge this one for now.

PR #119, opened 2024-09-24
fix: Schematic capacitor & resistor symbol fix

imrishabh18:


imrishabh18:
@seveibar Can you review this please? this part of the `core` is a bit confusing for me. Also, I am not able to add the svg snapshot for you to visualise in test, have tried few things but still the json for the schematic-symbol for `capacitor_horz` is getting undefined even though I can see that in the build. Will have to look more into it

What this PR does is, that the prop name of symbol_name is matched with the schematic-symbol list and then getting added into the circuit-json.

If this is okay, then will similarly update for other normal-components which are failing in tests. I hope I am making sense

seveibar (approved):

lib/components/base-components/NormalComponent.ts
```diff
@@ -189,7 +189,7 @@ export class NormalComponent<
     const { schematicSymbolName } = this.config
     if (!schematicSymbolName) return
     // TODO switch between horizontal and vertical based on schRotation
-    const symbol_name = `${this.config.schematicSymbolName}_horz`
+    const symbol_name = `${this.config.schematicSymbolName}`
```
seveibar, line 192: Hmm i think you need this

seveibar (approved):

seveibar (commented):
Going to take a closer look in ~1 hr, saw something a but weird

seveibar (commented):
Going to take a closer look in ~1 hr, saw something a but weird

lib/components/base-components/NormalComponent.ts
seveibar: this is why you're getting the type error, symbols always have _horz or _vert as part of their name

tests/readme.test.tsx
```diff
@@ -12,8 +12,8 @@ it("should create soup with various elements", () => {
   })
   project.add(board)
 
-  const R1 = new Resistor({ name: "R1", resistance: "10k", footprint: "0402" })
-  board.add(R1)
+  const R1 = <resistor name="R1" resistance="10k" footprint="0402" />;
+  board.add(R1);
```
seveibar, line 16: biome issue here fwiw, not sure if i have formatbot enabled on this repo

seveibar:
Ok sorry for the many reviews, the tests are correctly throwing, every symbol has an orientation attached to it e.g. `_horz`, but generally the approach seems fine to me

lib/components/base-components/NormalComponent.ts
seveibar: this is probably fine

lib/components/base-components/NormalComponent.ts
```diff
@@ -189,7 +189,7 @@ export class NormalComponent<
...
 
     const symbol = (symbols as any)[symbol_name] as SchSymbol | undefined
```
seveibar, line 194: the errors in this file are basically just saying you need to go refactor every component that defined the symbol without orientation, which is fine.

tests/readme.test.tsx
imrishabh18: The biome is enabled, will update the file

PR #133, opened 2024-10-02
Fix: chip design

imrishabh18:


imrishabh18:
@seveibar The chip still needs more work regarding the port spacing for the pins. Will create an issue for that and work

tests/examples/example4-kicad.test.tsx
```diff
@@ -46,6 +47,11 @@ it("example 4: kicad theme demo", async () => {
       />
     </board>,
   )
+  const fs = require('node:fs');
+  const circuitJson = project.getCircuitJson();
+  fs.writeFileSync('circuit.json', JSON.stringify(circuitJson, null, 2)); 
 
   project.render()
+
+//   expect(project).toMatchSchematicSnapshot(import.meta.path)
```
seveibar, line 56: Curious why we dont keep the snapshot

seveibar (approved):

tests/examples/example4-kicad.test.tsx
imrishabh18: Cause it's a circle, the new `circuit-to-svg` needs to be published for this snapshot to work

imrishabh18: And for that to work, I need to publish this `core` 😆 

@seveibar

PR #136, opened 2024-10-05
new component pcb_solder_paste created from smtpads

ShiboSoftwareDev:


circuit.json
```diff
@@ -0,0 +1,889 @@
+[
```
seveibar, line 1: we should remove this file, we can add to gitignore as well

lib/components/primitive-components/SmtPad.ts
```diff
@@ -26,6 +27,35 @@ export class SmtPad extends PrimitiveComponent<typeof smtPadProps> {
     }
   }
 
+  createSolderPasteComponent(): void {
```
seveibar, line 30:
```suggestion
  _addInferredSolderPasteComponent(): void {
```

lib/components/primitive-components/SmtPad.ts
```diff
@@ -26,6 +27,35 @@ export class SmtPad extends PrimitiveComponent<typeof smtPadProps> {
...
+        shape: "rect",
+        height: this._parsedProps.height * 0.7,
+        width: this._parsedProps.height * 0.7,
+        layer: this._parsedProps.layer,
+        pcbX: this._parsedProps.pcbX,
+        pcbY: this._parsedProps.pcbY,
+      })
+      solderPaste.pcb_smtpad_id = this.pcb_smtpad_id
+      this.add(solderPaste)
+      solderPaste.doInitialPcbPrimitiveRender()
```
seveibar, line 55: i think for simplicity, we shouldn't have pcb primitives contain other pcb primitives, so what we can do here is not use the `SolderPaste` class at all in this class, we'll just insert the `pcb_solder_paste` json inside of our rendering function.

You could remove the `SolderPaste` class completely in the short term for simplicity, it's just part of smtpad rendering (when someone needs it we can introduce the class)

seveibar (approved):

PR #139, opened 2024-10-06
fix: Chip placement

imrishabh18:
Had to update some snapshots, cause it's a chicken and egg problem as you know 😂. This is how it will look after the `circuit-to-svg` has this updated core in it.

Screenshot 2024-10-07 at 12 59 11 AM


seveibar (approved):
looking good!

PR #141, opened 2024-10-07
chore: update with latest footprinter version without lodash

imrishabh18:


imrishabh18:
@seveibar Removing the lodash dependency in footprinter decreased the bundle by 0.4mb

seveibar (approved):
LOVE IT

seveibar:
@imrishabh18 soup-util is ridiculously big considering it has super simple functions. That dep has no business being 2.6mb right now, it should be <100kb

PR #143, opened 2024-10-07
chore: update package to smaller bundle size

imrishabh18:


imrishabh18:
@seveibar Damn this is nice! The `soup-util` was having unused `bun/types` package

imrishabh18:
Looking in `circuit-json` if I can find something

seveibar:
Very nice, still suspiciously high # of files for soup-util (i think there should only be dist/index.js) but looking good!!

imrishabh18:
There is only one `index.js` file

Screenshot 2024-10-08 at 12 23 23 AM

seveibar:
Why does the bundle size analysis show 31 files?

seveibar:
Im actually not sure. Maybe not an issue

imrishabh18:
Even I am not sure, most of the packages are just only the `index.js` in the dist folder but still the bundle analysis is showing a lot of files

Screenshot 2024-10-08 at 12 28 56 AM

PR #153, opened 2024-10-10
Added silkscreen circle and rect

imrishabh18:
- Added test file for Via
- Added Silkscreen Circle and Rect component


lib/components/primitive-components/Via.ts
```diff
@@ -53,7 +53,6 @@ export class Via extends PrimitiveComponent<typeof viaProps> {
       x: newCenter.x,
       y: newCenter.y,
     })
-    this.matchedPort?._setPositionFromLayout(newCenter)
```
seveibar, line 56: This is kind of weird, just noting it

seveibar (approved):

lib/components/primitive-components/Via.ts
seveibar: I guess a via never has a matched port

imrishabh18: Yeah, that's why removed it. By mistake I added it in my last PR

PR #174, opened 2024-10-16
feat: symbol name added to schematic_component with width and height

imrishabh18:


lib/components/normal-components/Led.ts
```diff
@@ -15,15 +15,45 @@ export class Led extends NormalComponent<
...
+    )
+  }
+
+  doInitialSchematicComponentRender() {
+    const { db } = this.root!
+    const { _parsedProps: props } = this
+
+    const schematic_component = db.schematic_component.insert({
+      source_component_id: this.source_component_id!,
+      symbol_name: this.props.symbolName ?? ("led_horz" as BaseSymbolName),
```
seveibar, line 48: these are sort of supposed to come from the `config`, we're repeating the code a lot here. I think people will copy this pattern for new components and we'll have a lot of repeated code, is it possible to just use the config and have `doInitialSchematicComponentRender` be part of `NormalComponent`?

lib/components/normal-components/Led.ts
```diff
@@ -15,15 +15,45 @@ export class Led extends NormalComponent<
...
+      center: { x: props.pcbX ?? 0, y: props.pcbY ?? 0 },
+      rotation: props.pcbRotation ?? 0,
+      size: {
+        width: Number(props.schWidth ?? "1"),
+        height: Number(props.schHeight ?? "0.4"),
```
seveibar, line 53: likewise for these, ideally they come from the schematic symbol

seveibar (commented):
can we move the `doInitialSchematicComponentRender` code to `NormalComponent`? I think we're potentially going to make the code hard to change if we repeat it in each component

seveibar:
Ideally the code for each of these basic components is mostly just a config without any functions defined (except maybe for simulation later)

imrishabh18:
Yeah, but I was confused in the line of that we can possible have these height, width, center etc. different for `schematic` and different value for `pcb`. So, we will need to keep both the values, right?

lib/components/normal-components/Capacitor.ts
```diff
@@ -67,4 +79,22 @@ export class Capacitor extends NormalComponent<
...
+
+  doInitialSchematicComponentRender() {
+    const { db } = this.root!
+    const { _parsedProps: props } = this
+
+    const schematic_component = db.schematic_component.insert({
+      source_component_id: this.source_component_id!,
+      symbol_name:
+        this.props.symbolName ?? ("capacitor_horz" as BaseSymbolName),
+      center: { x: props.pcbX ?? 0, y: props.pcbY ?? 0 },
```
seveibar, line 91: oh this should be `schX`/`schY` too

imrishabh18: Yeah, I missed it

seveibar:
> Yeah, but I was confused in the line of that we can possible have these height, width, center etc. different for `schematic` and different value for `pcb`. So, we will need to keep both the values, right?

hmm not sure i understand, the sizes and positions can be totally different for the pcb vs schematic

lib/components/base-components/NormalComponent.ts
```diff
@@ -226,7 +226,10 @@ export class NormalComponent<
     const schematic_component = db.schematic_component.insert({
       center: { x: this.props.schX ?? 0, y: this.props.schY ?? 0 },
       rotation: this.props.schRotation ?? 0,
-      size: symbol.size,
+      size: {
```
imrishabh18, line 229: @seveibar Stupid me didn't check that I could have added it here 🤦‍♂️

seveibar (changes requested):
the schematic width and height come from the symbol, the pcbWidth and height come from the footprint, we should not allow them to be set as props

PR #187, opened 2024-10-21
feat: add Port support for `-size` suffix port arrangements

imrishabh18:


seveibar (approved):

seveibar:
looking good! just need to note that in the screenshot the pin order is still reversed so we don't get too used to it 😁

PR #194, opened 2024-10-23
Schematic Debug Object Rendering

seveibar:
- Update to latest circuit-to-svg with support for schematic_debug_object
- Add `_schDebugObjectsEnabled` prop to subcircuits to allow drawing debugging boxes
- Switch from IJumpAutorouter to MultilayerAutorouter which is the current best and probably has less bugs


lib/components/primitive-components/Trace.ts
```diff
@@ -629,19 +629,36 @@ export class Trace extends PrimitiveComponent<typeof traceProps> {
 
     const simpleRouteJsonInput: SimpleRouteJson = {
       minTraceWidth: 0.1,
-      obstacles: [], // leave empty to force traces to draw
+      // obstacles: [], // leave empty to force traces to draw
```
seveibar, line 632:
```suggestion
```

PR #206, opened 2024-10-28
Adding the missing trace space between the ports and route

imrishabh18:


lib/components/primitive-components/Trace.ts
```diff
@@ -623,18 +610,26 @@ export class Trace extends PrimitiveComponent<typeof traceProps> {
...
-        ...port._getGlobalSchematicPositionAfterLayout(),
-        ...projectPointInDirection(
-          port._getGlobalSchematicPositionAfterLayout(),
-          port.facingDirection!,
-          0.1501,
-        ),
-        layer: "top",
-      })
+    // Get port positions for later use
+    const portData = ports.map(({ port }) => ({
```
seveibar, line 614: PortsWithPosition


(Im anti-“data”, its basically the same as portsVariable)

lib/components/primitive-components/Trace.ts
```diff
@@ -674,25 +654,75 @@ export class Trace extends PrimitiveComponent<typeof traceProps> {
...
+      case "right":
+        firstStub.from.x -= STUB_LENGTH
+        break
+      case "up":
+        firstStub.from.y += STUB_LENGTH
+        break
+      case "down":
+        firstStub.from.y -= STUB_LENGTH
+        break
+    }
```
seveibar, line 695: Use projectPointInDirection

seveibar (approved):
Looking good but lets remove the verbose projection math into isolated functions, projectPointInDirection already exists iirc and projectPointInOppositeDirection could be introduced.

Issue #207, opened 2024-10-28
schPortArrangement could be undefined


imrishabh18:
Handle the case where the schPortArrangement is null, so uniformly distribute the ports on both the sides of the `<Chip />`

PR #211, opened 2024-10-29
fix: get the ports from footprint

imrishabh18:
Close - #207


lib/components/normal-components/Chip.ts
```diff
@@ -52,7 +52,7 @@ export class Chip<PinLabels extends string = never> extends NormalComponent<
       schPinSpacing: pinSpacing,
       schPinStyle: props.schPinStyle,
 
-      pinCount,
+      pinCount: pinCount === 0 ? this.getPortsFromFootprint().length : pinCount,
```
seveibar, line 55: can you make the `pinCount` accurate on line 41? Having the condition there makes more sense because we know to trust the `pinCount` variable to represent the `pinCount` at all times

seveibar (approved):
nice looking better!

lib/components/normal-components/Chip.ts
```diff
@@ -38,11 +38,12 @@ export class Chip<PinLabels extends string = never> extends NormalComponent<
     const { db } = this.root!
     const { _parsedProps: props } = this
 
-    const pinCount =
+    const totalPinCount =
```
seveibar, line 41: this var name is actually misleading
```suggestion
    const pinCountFromSchArrangement =
```

PR #231, opened 2024-11-04
WIP fix: Removing fatal error from missing props

imrishabh18:
Closes - #225 

@seveibar What do you think about this approach, Is it fine?


seveibar (changes requested):
no the defaults should be inside @tscircuit/props. The correct approach is to remove the component entirely and insert a new error (which will need to be defined in circuit json) indicating that the component was removed because it had invalid props.

seveibar:
@imrishabh18 generally with our errors they should always go to circuit json, if they don't go to circuit json then there's no way for the AI or the user to see them (the user doesn't look in the console on snippets, and they shouldn't need to)

Issue #216, opened 2024-10-31
Add `useDiode` hook


seveibar:

seveibar:
CC @andrii-balitskyi possibly good getting started issues

PR #242, opened 2024-11-06
Test useDiode hook

andrii-balitskyi:
https://github.com/tscircuit/core/issues/216 issue is about adding useDiode hook though it was already added in https://github.com/tscircuit/core/pull/217, so I just added a test.


seveibar (approved):
Nice!!! We could do the useLed hook whoch is similar?

Issue #191, opened 2024-10-21
Support using pin aliases instead of pin numbers in `schPortArrangement`


seveibar:
[...]

PR #246, opened 2024-11-07
feat: Add pins as string in portArrangement

imrishabh18:
Closes - #191


tests/components/normal-components/chip-pin-alias.test.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+  circuit.add(
+    <board width="10mm" height="10mm">
+      <chip
+        name="U1"
+        schPinSpacing={0.75}
+        schPortArrangement={{
+        leftSide: {
+          direction: "top-to-bottom",
+          pins: ["B1", "B2"],
```
seveibar, line 16: I think there's a bug in your implementation, it looks like you assume `pin1` is `B1` because of the integer in the string- we shouldn't do this. In fact, usually in circuits with `B1` there is also an `A1`. Can you modify this test 
```suggestion
          pins: ["B1", "B2"],
        },
        rightSide: { direction: "top-to-bottom", pins: ["A1", "A2"] }
```

Then check the output in the test to ensure that there are four schematic_port.pin_number


My hunch is that a lot of this PR will need to be redone because I think you might have changed code that specifically handled aliases, but I'm not completely sure

seveibar (changes requested):

tests/components/normal-components/chip-pin-alias.test.tsx
imrishabh18: @seveibar Actually, I am not assuming the `pin1` is `B1` because of the integer in the string. I am just incrementing the pinNumbers as new ports are getting added. The `B1` here is just an alias for that pin, so now I have updated the pin label as the alias, or default to pin number (will have to update the circuit_json with the variable name). Take a look at the updated test file and snap.

seveibar: I don't understand- `pin_number` should always be a number. `B1`/`A1` is a pin label

imrishabh18: Yeah, that's what I will update in circuit_json as well. Just wanted to confirm that the logic is not taking the string's Integer as pin number

lib/components/base-components/NormalComponent.ts
```diff
@@ -98,20 +98,23 @@ export class NormalComponent<
...
+            if(pinNumberInt) {
+              portsToCreate.push(
+                new Port({ pinNumber: pinNumberInt }),
+              )
+            } else {
+              portsToCreate.push(
+                new Port({ pinNumber: pinNumber, aliases: [pin] }),
+              )
+            }
+            pinNumber++
```
seveibar, line 117: This logic isn't sound in many ways

- doesn't handle `"pin1"`
- doesn't handle many number parsing cases properly (use `/\d+/.match` to determine if a string is a number). As is, stuff like 1.1 will cause issues.
- depending on the order in which things are defined, duplicate numbers are added

lib/utils/schematic/getAllDimensionsForSchematicBox.ts
```diff
@@ -183,21 +183,21 @@ export const getAllDimensionsForSchematicBox = (
...
   for (let sideIndex = 0; sideIndex < sidePinCounts.bottomSize; sideIndex++) {
-    const pinNumber =
+    const pinNumberOrAlias =
       params.schPortArrangement &&
       isExplicitPinMappingArrangement(params.schPortArrangement)
         ? params.schPortArrangement.bottomSide?.pins[sideIndex]!
         : truePinIndex + 1
     const pinStyle =
-      params.schPinStyle?.[`pin${pinNumber}`] ?? params.schPinStyle?.[pinNumber]
+      params.schPinStyle?.[`pin${pinNumberOrAlias}`] ?? params.schPinStyle?.[pinNumberOrAlias]
```
seveibar, line 192: This change is wrong, it ALWAYS be `pin${pinNumber}`

```suggestion
      params.schPinStyle?.[`pin${pinNumberOrAlias}`] ?? params.schPinStyle?.[pinNumberOrAlias]
```

seveibar: if there is an alias, we can check for `params.schPinStyle[alias]`

seveibar:
@imrishabh18 before doing anything with the implementation, can we build a really extensive test for this? Instead of just testing one example, can we put together a test with a bunch of hard examples?

imrishabh18:
Yeah, I think that would be better. I will write one test file with possible edge cases I can think of and will share

lib/components/base-components/NormalComponent.ts
seveibar: @imrishabh18 copy this function into core and export it, it's a useful function that handles virtually every case properly. We can pre-process `pins` with it.

When you copy over this function can you also copy it's tests, we can then patch easyeda-converter to import it when the function gets merged into core

https://github.com/tscircuit/easyeda-converter/blob/main/lib/normalize-pin-labels.ts

PR #253, opened 2024-11-08
Feat: error handling for missing footprint

### LINKED ISSUE #249: `missing_footprint` error if a `NormalComponent` doesn't
- may require defining `missing_footprint_error` as an error type in circuit json
- when rendering `NormalComponent`s create an error if they don't have any footprint representation and add to the db

/bounty $20

DhairyaMajmudar:
-> Adding `MissingFootprintError` class which extends `Error` class for error handling
-> Updating `NormalComponent` class.

closes #249 
/claim #249 

cc: @seveibar


seveibar (changes requested):
I modified the description to be more clear, this issue requires introducing a new error type to circuit json and adding that error to the db, not throwing

image

lib/components/base-components/NormalComponent.ts
```diff
@@ -375,6 +395,12 @@ export class NormalComponent<
       rotation: props.pcbRotation ?? 0,
       source_component_id: this.source_component_id!,
     })
+    if (this._footprintError) {
+      const source_component = db.pcb_missing_footprint_error.insert(
+        props._footprintError,
+      )
+      this.source_component_id = source_component.source_component_id
+    }
```
seveibar, line 403: this doesn't make sense, `pcb_missing_footprint_error.insert` returns an error object not a source_component, you shouldn't be setting that

lib/components/base-components/NormalComponent.ts
```diff
@@ -363,6 +374,15 @@ export class NormalComponent<
     this.schematic_component_id = schematic_component.schematic_component_id
   }
 
+  _missingFootprintError() {
```
seveibar, line 377: we don't need this function (functions that only do a simple thing, should just have the simple thing inline at the original function all site)

seveibar (changes requested):
make sure to introduce a test that creates a missing footprint error. This code actually inserts errors into every circuit, but it's difficult to know that because there isn't a test to make sure it doesn't.

You should introduce 2 tests, one that makes sure the error appears when a footprint is missing, and one that makes sure that there is no error if the footprint is provided.

seveibar (approved):
nice work

PR #257, opened 2024-11-08
Implement source_trace.subcircuit_connectivity_map_key

seveibar:
- **update circuit-to-svg, add reproduction of bad junction and bad trace push**
- **subcircuit_connectivity_map_key implementation**
- **add test for subcircuit_connectivity_map_key implementatio**


seveibar:
@ShiboSoftwareDev core now computes connectivity map keys and puts them in `source_trace.subcircuit_connectivity_map_key`, so we should be able to remove connectivity map calculations from pcb viewer in favor of this key eventually. I'll say one reason not to do it immediately is we might want to wait until subcircuit support is a bit more stable, since keys are not necessarily unique across subcircuits

PR #263, opened 2024-11-10
Normalized pin labels

imrishabh18:


lib/components/base-components/NormalComponent.ts
```diff
@@ -96,24 +97,18 @@ export class NormalComponent<
...
-        }
+      const normalizedPinLabels = this._getNormalizedPinLabels()
+      for (const pinLabels of normalizedPinLabels) {
+        const pinNumber = parseInt(pinLabels[0].replace("pin", ""))
+        portsToCreate.push(
+          new Port({
+            pinNumber,
+            name: pinLabels[1],
+            aliases: pinLabels,
+          }),
```
seveibar, line 110: we should add the originDescription, it's very useful for debugging

lib/components/base-components/NormalComponent.ts
```diff
@@ -639,6 +634,23 @@ export class NormalComponent<
     return this._parsedProps.schPortArrangement
   }
 
+  _getNormalizedPinLabels(): string[][] {
```
seveibar, line 637:
```suggestion
  _getNormalizedPinLabelsFromSchPortArrangement(): string[][] {
```

seveibar: i think otherwise this is a bit confusing because it doesn't look at the footprint or other sorts of pin labels (even in a later phase it could look at all the initialized ports, so depending on the context it's unclear what the source is)

lib/utils/schematic/getAllDimensionsForSchematicBox.ts
```diff
@@ -60,7 +60,7 @@ interface Params {
   >
   pinCount?: number
   schPortArrangement?: PortArrangement
-  pinLabels?: Record<string, string>
+  pinLabels?: string[][]
```
seveibar, line 63: let's not change this

lib/utils/schematic/normalizePinLabels.ts
```diff
@@ -0,0 +1,123 @@
...
+          result[i].push(label)
+        } else {
+          incrementalLabelCounts[label] = (incrementalLabelCounts[label] ?? 0) + 1
+          result[i].push(`${label}${incrementalLabelCounts[label]}`)
+        }
+      }
+    }
+
+  return result
+}
```
seveibar, line 123: we should copy over the associated test suite for this file (it's a really good test, and we'll refactor easyeda to import this function later so we'll want to keep the test inside core rather than over at the easyeda converter)

tests/components/normal-components/chip-pin-alias.test.tsx
```diff
@@ -0,0 +1,115 @@
...
+        }}
+      />
+    </board>,
+  )
+
+  circuit.render()
+
+  const fs = require("fs")
+  fs.writeFileSync("circuit.json", JSON.stringify(circuit.getCircuitJson(), null, 2))
+  expect(circuit.getCircuitJson()).toMatchSchematicSnapshot(import.meta.path)
```
seveibar, line 42: 👏 👏 👏 👏  this is a really good sample. We should remember to remove the file writing but looks good (ikik this is a draft) but looks good 👍

seveibar: this test should _not_ rely on snapshots however!!!! Load in all the `source_port`s, sort them, and confirm in code that they match expectations. 

FWIW i think this test in particular points to a very confusing behavior from the normalizer. `B11` and `B112` should probably be `B1_1` and `B1_2`!!! We can fix it in a future PR and look for these edge cases with the normalizer- but really this test reproduces a bug and we should leave a comment in the test that indicates that it's a known issue that should be fixed

seveibar (commented):
some small comments on naming and changing `getAllDimensionsForSchematicBox` (I don't think we need to change this file) but the tests are good. These tests _should not_ just be snapshot tests- we should test for the specific pin label/number matching that we believe is correct.

When I looked at the snapshot it seemed ok, but then when I read the code I realized the **snapshot actually showed a bug and I didn't notice**. Future editors of this code also won't notice the bug even if the snapshot changes!!! We need to have a concise bit of code in the test that checks that we're getting what we expect/what it currently outputs and a note that there's a bug.

The bug is `B1` turning into `B11` and `B12` instead of `B1_1` and `B1_2`- we don't have to fix it in this PR but we should improve the test so that it's easy for the next person to understand and fix the behavior

lib/components/base-components/NormalComponent.ts
```diff
@@ -115,6 +122,26 @@ export class NormalComponent<
           }
         }
       }
+      // pin label are string
+      if (normalizationNeeded) {
```
seveibar, line 126: not a big deal, but by convention i try to always have a prefix `is*` `are*` etc.

```suggestion
      if (isNormalizationNeeded) {
```

lib/components/base-components/NormalComponent.ts
```diff
@@ -115,6 +122,26 @@ export class NormalComponent<
...
+                name: pinName.toString(),
+                aliases: [pinLabel, pinName],
+              },
+              {
+                originDescription: `schPortArrangement:pin${pinNumber}`,
+              },
+            ),
+          )
+        }
+      }
```
seveibar, line 144: this code seems to create duplicate ports. You're them before you know if normalization is needed

seveibar (changes requested):
this PR is taking too long, is too big and important code is becoming too complicated- in particular `getAllDimensionsForSchematicBox` and `NormalComponent.ts`. Revert changes to those files and just PR the skip tests, then create issues for each of the tests to unskip. The other files we should keep.

If there is complex code being introduced, it should be in a small PR. I'm spending too much time lately fixing bugs and reviewing, the solution to this in my experience is to mandate small PRs where it's easy to understand what they do.

PR #271, opened 2024-11-11
Pins are normalised when the labels are string

imrishabh18:
Pins are normalised for both the string and number scenario


seveibar (changes requested):
we should not change tests to skip tests to avoid regressions

seveibar:
@imrishabh18 the only time you can do skip tests is when they're first introduced- when this code is merged it's going to break snippets. I'm also not really sure what this code does yet for schematic port arrangements, it seems like it's going to break it

imrishabh18:
@seveibar I will push the update which fixes pins for numbers, but then that will make the PR huge again.

seveibar (changes requested):
The code is just too complex. That giant ternary statement.... its too hard to tell what ot does already and of we build on it itll become unmaintainable

lib/utils/schematic/getAllDimensionsForSchematicBox.ts
```diff
@@ -185,7 +186,8 @@ export const getAllDimensionsForSchematicBox = (
   for (let sideIndex = 0; sideIndex < sidePinCounts.bottomSize; sideIndex++) {
     const pinNumber =
       params.schPortArrangement &&
-      isExplicitPinMappingArrangement(params.schPortArrangement)
+      isExplicitPinMappingArrangement(params.schPortArrangement) &&
+      typeof params.schPortArrangement.bottomSide?.pins[sideIndex] !== "string"
```
seveibar, line 190: `"pin1": string` === `1:number`

PR #273, opened 2024-11-11
Port.ts getRelativeDirection refactored

ShiboSoftwareDev:
if there is no side defined get facingDirection based on nearest point in container

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@tscircuit/core",
   "type": "module",
-  "version": "0.0.179",
+  "version": "0.0.180",
   "types": "dist/index.d.ts",
   "main": "dist/index.js",
   "module": "dist/index.js",


lib/components/primitive-components/Port.ts
```diff
@@ -273,8 +275,22 @@ export class Port extends PrimitiveComponent<typeof portProps> {
...
+    const nearestPointInContainer = {
+      x: Math.min(
+        Math.max(portCenter.x, containerCenter.x - containerWidth / 2),
+        containerCenter.x + containerWidth / 2,
+      ),
+      y: Math.min(
+        Math.max(portCenter.y, containerCenter.y - containerHeight / 2),
+        containerCenter.y + containerHeight / 2,
+      ),
+    }
```
seveibar, line 287: i'm a bit confused by this logic. Can you break it into a couple variables? It looks like you're computing the `containerTopY`, `containerBottomY` etc. so that could help. But I don't see how this will properly always work intuitively, we could refactor. the code to make it more obvious.

It would also help to have a test that demonstrates that this works, but `localPortInfo.side` is usually defined so i'm not sure how to recreate it, you could break the functionality out into a testable utility function `getDirectionFromPointToNearestSideOfContainer`

seveibar (changes requested):
we'd want to make sure this is tested, if we can't do a render test, a couple unit tests against a utility function would help confirm it works

Issue #281, opened 2024-11-12
Net label anchor for capacitors not correct


imrishabh18:
https://tscircuit.com/editor?snippet_id=d2c1c529-97b0-41de-96fa-2cc0c84ba342

![Image](https://github.com/user-attachments/assets/4c68fdda-d43e-4cad-bcdc-19aa96ac92bc)

PR #291, opened 2024-11-13
fix: anchor position for netalias

imrishabh18:
Closes - #281


seveibar (approved):
this is ok but it doesn't fix 99% of net label issues, generally people do not define <netalias /> when creating net labels, it would be better to fix the case where somebody says something like <trace from="net.gnd" to=".R1 > .pin1" />

lib/utils/schematic/getEnteringEdgeFromDirection.ts
```diff
@@ -6,8 +6,8 @@ export const getEnteringEdgeFromDirection = (
   return (
     (
       {
-        up: "bottom",
-        down: "top",
+        up: "top",
+        down: "bottom",
```
seveibar, line 10: this is wrong, if you're going down you enter from the top

seveibar (changes requested):
broke a utility function

seveibar:
@imrishabh18 i pushed to this PR to get your tests to pass, i just re-reran snapshots to fix conflicts, but then i discovered that there was a broken utility function

seveibar:
There is still some good stuff in this PR, but we can mix it with this PR and we've fixed (i think) 100% of cases https://github.com/tscircuit/core/pull/296

PR #300, opened 2024-11-14
Feat: potentiometer

imrishabh18:
Close - #171


lib/components/normal-components/Potentiometer.ts
```diff
@@ -0,0 +1,50 @@
...
+      zodProps: potentiometerProps,
+      sourceFtype: "simple_potentiometer" as Ftype,
+    }
+  }
+
+  initPorts() {
+    super.initPorts({
+      additionalAliases: {
+        pin1: ["anode", "pos", "left"],
+        pin2: ["cathode", "neg", "right"],
```
seveibar, line 31:
```suggestion
        pin1: [ "pos", "left"],
        pin2: ["neg", "right"],
```

lib/components/normal-components/Potentiometer.ts
```diff
@@ -0,0 +1,50 @@
...
+      },
+    })
+  }
+
+  _getSchematicSymbolDisplayValue(): string | undefined {
+    return `${formatSiUnit(this._parsedProps.maxResistance)}Ω`
```
seveibar, line 37: nice 👍

tests/components/normal-components/potentiometer.test.tsx
```diff
@@ -0,0 +1,17 @@
...
+  circuit.add(
+    <board width="10mm" height="10mm">
+      <potentiometer name="P1" maxResistance="10k" />
+    </board>,
+  )
+
+  circuit.render()
+
+  expect(circuit.db.schematic_component.list()).toHaveLength(1)
+  expect(circuit.db.schematic_port.list()).toHaveLength(2)
```
seveibar, line 16: can you throw in a snapshot here?

seveibar (commented):
looking good! can we add a snapshot?

imrishabh18:
@seveibar Couple of things I noticed from the snapshots are, that if the `schRotation` is 0 the component is flipped to what it is in the schematic-symbols. And when `180` rotation is also not rotating it

seveibar (approved):
nice find

Issue #308, opened 2024-11-15
Silkscreen Paths cause extra pins in schematic view


seveibar:
/bounty $10

![Image](https://github.com/user-attachments/assets/c50d8ad6-b86a-4222-b142-92bac140993c)

PR #311, opened 2024-11-15
Fix silkscreen paths causing extra pins in schematic view

vishwamartur:
Related to #308

Add method to prevent silkscreen paths from being interpreted as pins in schematic view.

* **SilkscreenPath.ts**
  - Add `doInitialSchematicComponentRender` method to `SilkscreenPath` class to prevent silkscreen paths from being interpreted as pins in the schematic view.
* **silkscreenpath.test.tsx**
  - Add test case to verify that silkscreen paths are not interpreted as pins in the schematic view.

### DIFF
--- a/lib/components/primitive-components/FabricationNotePath.ts
+++ b/lib/components/primitive-components/FabricationNotePath.ts
@@ -48,4 +48,8 @@ export class FabricationNotePath extends PrimitiveComponent<
     this.fabrication_note_path_id =
       fabrication_note_path.pcb_fabrication_note_path_id
   }
+
+  doInitialSchematicComponentRender(): void {
+    // This method intentionally left blank to prevent fabrication note paths from being interpreted as pins in the schematic view
+  }
 }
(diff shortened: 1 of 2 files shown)


tests/components/primitive-components/silkscreenpath.test.tsx
```diff
@@ -36,3 +36,27 @@ test("SilkscreenPath rendering", () => {
...
+        layer="top"
+      />
+    </board>,
+  )
+
+  project.render()
+
+  const schematicPins = project.db.schematic_pin.list()
+
+  expect(schematicPins.length).toBe(0)
```
seveibar, line 61: this test is good

seveibar: hmm actually the repro calls for a `<chip />`

seveibar (changes requested):
fix doesn't make sense, test doesn't test `<chip footprint=... />`

seveibar:
The fix is probably in this code

image

but you should get a good repro first

PR #313, opened 2024-11-16
Feat: adding use-pinheader hook and test case

### LINKED ISSUE #312: Implement `usePinHeader` (accept pin labels as argument similar to chip)
See `useChip` for example

/bounty $5

DhairyaMajmudar:
-> Creating use-pinheader hook which accepts a string array of pinlabels as argument.
-> Creating a test case for it.

closes #312 
/claim #312 

cc: @seveibar

### DIFF
--- a/lib/hooks/use-pinheader.tsx
+++ b/lib/hooks/use-pinheader.tsx
@@ -1,8 +1,15 @@
-import type { PinHeaderProps } from "@tscircuit/props"
-import { createUseComponent } from "./create-use-component"
+import type { PinHeaderProps } from "@tscircuit/props";
+import { createUseComponent } from "./create-use-component";
 
-export const usePinHeader = (pinLabels: string[]) =>
+export const usePinHeader = (
+  name: string,
+  options: {
+    pinLabels: PinHeaderProps["pinLabels"];
+  }
+) =>
   createUseComponent(
-    (props: PinHeaderProps) => <pinheader pinLabels={pinLabels} {...props} />,
-    pinLabels,
-  )
+    (props: PinHeaderProps) => (
+      <pinheader {...props} name={name} pinLabels={options.pinLabels} />
+    ),
+    options.pinLabels!
+  );
--- a/lib/hooks/use-pinheader.tsx
+++ b/lib/hooks/use-pinheader.tsx
@@ -1,15 +1,15 @@
-import type { PinHeaderProps } from "@tscircuit/props";
-import { createUseComponent } from "./create-use-component";
+import type { PinHeaderProps } from "@tscircuit/props"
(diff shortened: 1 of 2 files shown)


lib/hooks/use-pinheader.tsx
```diff
@@ -0,0 +1,8 @@
+import type { PinHeaderProps } from "@tscircuit/props"
+import { createUseComponent } from "./create-use-component"
+
+export const usePinHeader = (pinLabels: string[]) =>
+  createUseComponent(
+    (props: PinHeaderProps) => <pinheader pinLabels={pinLabels} {...props} />,
```
seveibar, line 6: Needs to have name prop as first argument, second argument should be { pinLabels } (an options style object). Use the pinlabelspec or PinHeaderProps[“pinLabels”] to accept all forms of pinlabels including object variants

seveibar (changes requested):
Very close

lib/hooks/use-pinheader.tsx
DhairyaMajmudar: Done the changes 👍🏻

lib/hooks/use-pinheader.tsx
```diff
@@ -0,0 +1,15 @@
+import type { PinHeaderProps } from "@tscircuit/props"
+import { createUseComponent } from "./create-use-component"
+
+export const usePinHeader = (
+  name: string,
+  options: {
+    pinLabels: PinHeaderProps["pinLabels"]
+  },
```
seveibar, line 8: Looking at your code vs other examples like useChip, i think “options” should actually be props: PinHeaderPrips

tests/hooks/use-pinheader.test.tsx
```diff
@@ -0,0 +1,31 @@
...
+  const circuit = new Circuit()
+
+  const useHeader = usePinHeader("Header", {
+    pinLabels: ["VCC", "GND", "TX", "RX"],
+  })
+
+  const Header = useHeader("Header", {
+    pinCount: 4,
+    pinLabels: ["VCC", "GND", "TX", "RX"],
+  })
```
seveibar, line 15: The usage is wrong, see the useChip or useResistor examples.


const J1 = usePinHeader(“J1”, { pinLabels })

J1.myPin // string

<J1 />

seveibar (changes requested):

tests/hooks/use-pinheader.test.tsx
DhairyaMajmudar: gotcha, I've fixed the variable names 👍🏻

seveibar (changes requested):
a hook shouldn't return another hook, only use the `use*` prefix for hooks

Issue #310, opened 2024-11-15
some schematic snapshots SVG's aren't rendering the symbols correctly.


Abse2001:
start with the battery test.

![Image](https://github.com/user-attachments/assets/a7bdb9cd-d18d-40e2-9acb-e62acde34393)

ShiboSoftwareDev:
fixed in #314

PR #314, opened 2024-11-16
fixed battery symbol not rendering

ShiboSoftwareDev:
/claim #310


imrishabh18 (commented):
Looks good! Just remove the lockfile

PR #317, opened 2024-11-17
FIX: Type error fix from `create-use-component-2-number-inference.test.tsx` file

DhairyaMajmudar:
-> Fixing incorrect object parameter call
-> Removed type-error ignore 

cc: @seveibar


seveibar (changes requested):
this isn't a bug, this test is testing that there is an error. They are important to keep to make sure that those lines DO throw errors

DhairyaMajmudar:
> this isn't a bug, this test is testing that there is an error. They are important to keep to make sure that those lines DO throw errors

Oh! That one is intentionally kept for errors, cool then let me close this PR 👍🏻

PR #318, opened 2024-11-17
`schPinStyle` should accept pin labels

### LINKED ISSUE #301: `schPinStyle` should accept pin labels as keys (not just pin numbers)
Try to keep the solution not super hacky. `schPortArrangement` already has similar functionality where you can specify pin numbers or labels and there's some kind of utility function for it.

/bounty $10

DhairyaMajmudar:
-> `schPinStyle` should accept pin labels along with pin numbers
-> Example usage: 

```tsx
const params = {
  schPinStyle: {
    pin1: { leftMargin: 10 },
    2: { rightMargin: 5 },
    "3": { topMargin: 15 },
    VCC: { bottomMargin: 20 },
  },
};
```

closes #301 
/claim #301

cc: @seveibar


seveibar (commented):
Can you show that it works a new visual snapshot test. Try to make it very obvious in the snapshot that the margins are applying by making the margins large

DhairyaMajmudar:
> you need a usage test that uses `schPinStyle`

@seveibar I've added a new test using `schPinStyle` along with its snapshot. I hope now we are good to go : )

seveibar (changes requested):
from the visual snapshot, i can't confirm that `schPinStyle` is working. Looking at `WRITING_TESTS.md` for the basic format for creating a usage test

lib/utils/schematic/getAllDimensionsForSchematicBox.ts
```diff
@@ -194,8 +194,13 @@ export const getAllDimensionsForSchematicBox = (
       truePinIndex,
     })
 
+    const pinLabel =
+      params.pinLabels?.[`pin${pinNumber}`] ?? params.pinLabels?.[pinNumber]
```
seveibar, line 198: this doesn't handle the case where `pinLabels: Record<string, string[]>`

DhairyaMajmudar:
> from the visual snapshot, i can't confirm that `schPinStyle` is working. Looking at `WRITING_TESTS.md` for the basic format for creating a usage test

Got your point thanks!! adding a test for that

lib/utils/schematic/getAllDimensionsForSchematicBox.ts
DhairyaMajmudar: adding a case for that 👍🏻

DhairyaMajmudar:
@seveibar I've updated the PR with required test 👍🏻

lib/utils/schematic/getAllDimensionsForSchematicBox.ts
DhairyaMajmudar: added a case handle their

tests/examples/example8-schPinStyle-accepts-pinLabels.test.tsx
```diff
@@ -0,0 +1,22 @@
...
+        schPinStyle={{ PWR: { bottomMargin: 1 }, GND: { bottomMargin: 0.5 } }}
+        pcbX={0}
+        pcbY={0}
+      />
+    </board>,
+  )
+
+  circuit.render()
+
+  await expect(circuit).toMatchSchematicSnapshot(import.meta.path)
```
seveibar, line 21: Can you add a pinlabels array test here as well? Generally you should try to test every code path, i suspect there may be a bug with pin label arrays based on your implementation

DhairyaMajmudar: sure thing : )

seveibar: For your test, try to test (in a single test) variations on the usage, eg the pin label could be the first entry)

i also suspect you may not handle the case where schPinStyle contains a pin label that doesnt exist, but this can be a future issue.

based on your code, i think there would be a bug if schPinStyle has a substring in the key, eg if it had a key “GND” but the pinlabels array listed “GND1” , “GND2”

DhairyaMajmudar: Gotcha will update for the same

PR #319, opened 2024-11-17
introduce <pushbutton /> component

Abse2001:


seveibar (approved):
The symbol is upside down, but i dont think you caused that, i think its an issue woth a schematic symbols update or circuit to svg update, will look into it separately

great work!

Abse2001:
@seveibar I talked to @imrishabh18  on discord about it there is already an issue for it and Im looking into it I dont understand circuit-to-svg that good yet so anyone that want to fix it feel free to do so.

Issue #320, opened 2024-11-17
CI: Failing bundle-size reporter workflow


DhairyaMajmudar:
-> It has been noticed that whenever a new PR is opened by non-tscircuit org. member bundle-size reporter workflow fails to do possible permission issue

-> Error logs:-
```
/home/runner/work/_actions/marocchino/sticky-pull-request-comment/v2/node_modules/@actions/core/lib/core.js:106
        throw new Error(`Input required and not supplied: ${name}`);
^
Error: Input required and not supplied: GITHUB_TOKEN
    at Object.getInput (/home/runner/work/_actions/marocchino/sticky-pull-request-comment/v2/node_modules/@actions/core/lib/core.js:106:1)
    at Object.88 (/home/runner/work/_actions/marocchino/sticky-pull-request-comment/v2/lib/config.js:67:1)
    at __nccwpck_require__ (/home/runner/work/_actions/marocchino/sticky-pull-request-comment/v2/webpack/bootstrap:21:1)
    at Object.3109 (/home/runner/work/_actions/marocchino/sticky-pull-request-comment/v2/lib/main.js:37:1)
    at __nccwpck_require__ (/home/runner/work/_actions/marocchino/sticky-pull-request-comment/v2/webpack/bootstrap:21:1)
    at /home/runner/work/_actions/marocchino/sticky-pull-request-comment/v2/webpack/startup:4:1
    at Object.<anonymous> (/home/runner/work/_actions/marocchino/sticky-pull-request-comment/v2/dist/index.js:32978:12)
    at Module._compile (node:internal/modules/cjs/loader:1358:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1416:10)
    at Module.load (node:internal/modules/cjs/loader:1208:32)
```

-> Possible Fix:-

Errors are coming from `Update comment step` https://github.com/tscircuit/core/blob/main/.github/workflows/bundle-size.yml#L86-L106, This step can be by passed

AkashJana18:
/attempt #320

1. Ensure `GITHUB_TOKEN` is passed to the `sticky-pull-request-comment` action in `bundle-size.yml`.
2. Define necessary permissions (e.g., `pull-requests: write`, `issues: write`) in the workflow file.
3. Optionally, add a conditional check to skip the "Update comment" step for non-members.
4. Test the workflow with a PR from a non-member to confirm the fix works.

PR #327, opened 2024-11-18
Fix failing bundle-size reporter workflow

vishwamartur:
Related to #320

Fix the bundle-size reporter workflow for PRs from non-tscircuit org members.

* **README.md**
  - Add a note about the `GITHUB_TOKEN` requirement for the bundle-size reporter workflow.

* **.github/workflows/bundle-size.yml**
  - Ensure `GITHUB_TOKEN` is passed to the `sticky-pull-request-comment` action.
  - Define necessary permissions (`pull-requests: write`, `issues: write`) in the workflow file.
  - Add a conditional check to skip the "Update comment" step for non-members.

/claim #320


seveibar:
You didn't fix it you disabled it...

PR #333, opened 2024-11-19
feat: cache route

imrishabh18:
Is this how it should be cached, if we are re-rendering?

Or save this cacheRoute in a separate file to persist?


lib/components/primitive-components/Trace/Trace.ts
```diff
@@ -262,14 +257,26 @@ export class Trace
...
+    }
+
+    // Check for cached route
+    const cachedRoute = subcircuit._parsedProps.pcbRouteCache?.pcbTraces
+    if (cachedRoute) {
+      const pcb_trace = db.pcb_trace.insert({
+        route: cachedRoute.flatMap(trace => trace.route),
+        source_trace_id: this.source_trace_id!,
+      })
+      this.pcb_trace_id = pcb_trace.pcb_trace_id
```
seveibar, line 275: yes this is basically how it should work

seveibar (approved):
yes that is basically it.

PR #340, opened 2024-11-22
create rotated_rects on chip rotation

ShiboSoftwareDev:


seveibar (changes requested):
This looks pretty good but we want to still use regular rects for any 90 degree (-90, 90, 180, 270 etc) rotation

Regular rects (axis aligned rects) have properties that make them simpler to work with, it also makes downstream viewers easier to implement

PR #342, opened 2024-11-22
Fixed manufacturer Part Number should go below refdes for four-sided pin layouts,handled three pin layouts and added tests for it.

### LINKED ISSUE #341: Manufacturer Part Number should go below refdes for four-sided pin layouts
(switch location)

Image

Abse2001:
/closes #341


lib/components/normal-components/Chip.ts
```diff
@@ -32,8 +32,8 @@ export class Chip<PinLabels extends string = never> extends NormalComponent<
       supplier_part_numbers: props.supplierPartNumbers,
     })
     const dimensions = this._getSchematicBoxDimensions()
-    const hasOnlyLeftAndRightPins =
-      props.schPortArrangement?.topSide !== undefined &&
+    const hasTopAndBottomPins =
```
seveibar, line 35:
```suggestion
    const hasTopOrBottomPins =
```

seveibar (approved):
EXTREMELY GOOD TEST hats off to you

@imrishabh18 i think we should have a name for this snapshot pattern where it's like a grid of all the possible ways to do configure a visual thing

@anas-sarkez please tweak the variable name before merging! See comment

Abse2001:
@seveibar Ty I will tweak the variable name

Issue #180, opened 2024-10-18
Support `<crystal />`


seveibar:
e.g.

![Image](https://github.com/user-attachments/assets/d4f3da90-e829-49b5-b9ac-a044f11f73a5)

imrishabh18:
@seveibar Our symbol represents only 2 ports though. Is this fine?

![Image](https://github.com/user-attachments/assets/5d83d52f-968f-4117-a264-d73b580f855c)

seveibar:
Hmm we need both variants, I think the one I have linked there is a special symbol that represents when there's a crystal that has capacitors included in it (which is common and a good thing to have because it makes routing a lot easier)

seveibar:
I would give claude both versions of the crystal and ask it what the one with capacitors is called

PR #349, opened 2024-11-24
feat: add crystal

imrishabh18:
Close - #180


seveibar (approved):
Looks good, note that there are also 4 pin crystals

seveibar:
We can do four pin crystals in a future issue

Issue #353, opened 2024-11-26
name prop being passed to the imported modules are not added to the source_component of circuit json


imrishabh18:
In this snippet - https://tscircuit.com/editor?snippet_id=6d278271-c6d9-4a5f-889e-905477176178

The name prop being passed to all the `<RedLed />` is not getting added into the source_component. All of them are having the same name in the circuit json.

![Image](https://github.com/user-attachments/assets/cb1d44e7-8f09-4d4d-8d6b-ca181203f0a4)

In place of LED it should have been `led1`

PR #355, opened 2024-11-26
preserve explicit component names in createUseComponent

mrudulpatil18:
Issue: #353 

Suspected Problem: 
`const RedLed = useRedLed("LED")` takes preference over `<RedLed name="led1" .../>`

Changes ensure:
- If a name is passed directly when using the component, it will be used
- If no name is passed, it falls back to the default name from useRedLed
- added a test with code similar to [reference](https://tscircuit.com/editor?snippet_id=6d278271-c6d9-4a5f-889e-905477176178) to verify `name` parameter

/claim #353


seveibar (changes requested):
so this is a great PR, I actually like this approach quite a bit, but I think it creates an ambiguous situation for users because a name is specified twice. In addition, the "typed selectors" will be wrong (e.g. `RedLedComp.pin1` will not work properly). Can we instead error if the name prop is provided and is different from the one specified in the hook?

mrudulpatil18:
@seveibar 

Changes:

- Throw error if name from hook and name from prop don't match
- added a test file to check this behaviour

seveibar (approved):

PR #357, opened 2024-11-26
Add Cachebust Origin to avoid cors issues

seveibar:


seveibar:
CC @Abse2001 this may be of interest

PR #363, opened 2024-11-27
fix: bundle-size.yml file doesn't run on pr from forks

### LINKED ISSUE #360: size analysis doesn't run on pr from forks
I'll do this once I have time

Anshgrover23:
Fixes #360 
/claim #360

I changed the workflow to use `pull_request_target` instead of `pull_request`

### DIFF
--- a/.github/workflows/bundle-size.yml
+++ b/.github/workflows/bundle-size.yml
@@ -1,7 +1,7 @@
 name: Size
 
 on:
-  pull_request:
+  pull_request_target:
     branches:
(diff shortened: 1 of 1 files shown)


seveibar:
@Anshgrover23 have you verified that this works?

seveibar:
We already fixed this on the snippets repo, it doesnt require a PAT, there is a workflow on snippets

ShiboSoftwareDev:
use the proxy

Issue #366, opened 2024-11-28
Usb c snippet, where there are "touching" VCC1 and VCC2 pins, doesnt route. Create reproduction


seveibar:
![image](https://github.com/user-attachments/assets/06a148d8-2aac-4c13-a606-660549ee5273)

When creating reproductions, you can either do a skip test or a snapshot (in this case, snapshot is preferable)

You are welcome to simplify the reproduction, look at snippets to find the usb c module

/bounty $10

PR #367, opened 2024-11-29
repro USB-C GND & VBUS route issue

Abse2001:
/claim #366


seveibar (changes requested):
Only issue is test name, then its good to go, extension is off

seveibar (approved):
Niceee

PR #368, opened 2024-11-29
draw normal rect smtpads when rotation is 90 degrees

ShiboSoftwareDev:


seveibar (approved):

seveibar:
@ShiboSoftwareDev this could be improved/stabilized by doimg a repro test or adding a case, but good to have the proper fix

ShiboSoftwareDev:
> @ShiboSoftwareDev this could be improved/stabilized by doimg a repro test or adding a case, but good to have the proper fix

Yeah I can create a test for it

PR #370, opened 2024-11-29
Implemented `setImmediate` to yield while rendering when renderUntilSettled is used to help fixes website freeze up issue

### LINKED ISSUE #346: `Circuit` should return partial renders and use `setImmediate` to defer full renders when `renderUntilSettled` is used
We're seeing the website freeze up because we don't yield during a render, let's use setImmediate to yield while rendering when renderUntilSettled is used

/bounty $20

This is a bit of a difficult implementation

anas-sarkez:
/claim #346
/closes #346

### DIFF
--- a/lib/Circuit.ts
+++ b/lib/Circuit.ts
@@ -81,7 +81,9 @@ export class Circuit {
     )
   }
 
-  render() {
+  async render(
+    partial = false,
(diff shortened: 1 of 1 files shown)


seveibar (changes requested):
this issue isn't actually really possible, sorry about that.

You're not allowed to change `render()` to not be synchronous or it will break React, I'm surprised this didn't break the tests

PR #372, opened 2024-11-30
add: Support root.schematicDisabled to skip schematic rendering #371

### LINKED ISSUE #371: Support `root.schematicDisabled` to skip schematic rendering
This prop will disable most phases related to schematic rendering. It should be allowed on the root circuit, we might allow it on subcircuits but i'd prefer it just as a root property in the short term

/bounty $15

Anshgrover23:
## Description:
- added schematicDisabled prop to Circuit class to automatically apply to its instance root.
## Reference Issue:
Fixes #371 
/claim #371

### DIFF
--- a/lib/Circuit.ts
+++ b/lib/Circuit.ts
@@ -19,6 +19,7 @@ export class Circuit {
   db: SoupUtilObjects
   root: Circuit | null = null
   isRoot = true
+  schematicDisabled: boolean = false // Add the schematicDisabled property
 
   _hasRenderedAtleastOnce = false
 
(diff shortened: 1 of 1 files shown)


imrishabh18 (changes requested):
I don't think this is how it should be implemented. The prop should be added first in the https://github.com/tscircuit/props repo. It is an optional prop which will be passed to the `<board />` or `<group />` . And if it's present, then in the renderPhase skip the phase which has a prefix of `Schematic`. It should be implement something in this line.

Correct me here, if I am missing something.
cc: @seveibar

seveibar:
@imrishabh18 in this case i recommended it be implemented on the root circuit instead of as a subcircuit prop, although a subcircuit prop is acceptable too. My reason for this is it's more of a platform parameter than a circuit parameter, e.g. if someone selects a schematic or pcb view, the platform (snippets) should be able to say "only render schematics"

that said, @Anshgrover23 this is sort of a a low quality PR, you didn't include any testing. I'm closing this PR. Please don't re-attempt because based on the code here you didn't understand the issue and we've spent a lot of time helping you on PRs, be more careful in future issues. When you open a PR, make sure you understand the code and create tests that demonstrate you completed the issue.

PR #376, opened 2024-12-02
Support `root.schematicDisabled` to skip rendering all schematic components. With testing

### LINKED ISSUE #371: Support `root.schematicDisabled` to skip schematic rendering

anas-sarkez:
/closes #371 
/claim #371


lib/components/base-components/NormalComponent.ts
```diff
@@ -373,6 +375,7 @@ export class NormalComponent<
   }
 
   _doInitialSchematicComponentRenderWithSymbol() {
+    if (this.root?.schematicDisabled) return
```
imrishabh18, line 378: I think rather checking the `schematicDisabled` in each and every function. You can check it in the `runRenderPhase` function,  where all the function having the prefix as `Schematic` can be skipped.

cc: @seveibar

imrishabh18 (changes requested):
I think rather than checking the

lib/components/base-components/NormalComponent.ts
seveibar: @imrishabh18 so this makes sense but i actually like the approach where we do it for each function. The reason why is we’re eventually transitioning to a “dependency graph”-style render pipeline, so id prefer to keep Renderable simple until we make that transition

imrishabh18 (approved):

lib/components/normal-components/PinHeader.ts
```diff
@@ -62,20 +62,22 @@ export class PinHeader extends NormalComponent<typeof pinHeaderProps> {
...
+        text: props.name ?? "",
+        schematic_component_id: source_component.source_component_id,
+        anchor: "left",
+        rotation: 0,
+        position: {
+          x: (props.schX ?? 0) - (schematic_box_width ?? 0) / 2,
+          y: (props.schY ?? 0) + (schematic_box_height ?? 0) / 2 + 0.13,
+        },
+        color: "#006464",
+      })
```
seveibar, line 79: This code shouldn't even be here, this is supposed to be in a schematic rendering phase, making a note because it should be a separate issue- should be moved to schematic component render phase

tests/examples/example9-shcematicDisabled.test.tsx
```diff
@@ -0,0 +1,41 @@
+import { expect, test } from "bun:test"
```
seveibar, line 1: test name has a typo

PR #378, opened 2024-12-02
Moved Schematic Text Logic to `SchematicComponentRenderWithSchematicBoxDimensions`

### LINKED ISSUE #377: Move Schematic Text Logic to SchematicComponentRender phase
This code shouldn't even be here, this is supposed to be in a schematic rendering phase, making a note because it should be a separate issue- should be moved to schematic component render phase

_Originally posted by @seveibar in https://github.com/tscircuit/core/pull/376#discussion_r1865635399_

anas-sarkez:
/closes #377 
/claim #377


seveibar (approved):

anas-sarkez:
> @seveibar @anas-sarkez it is also present in [jumper.ts](https://github.com/tscircuit/core/blob/9dbb2c896a5074a542e95d284ad1914b0d62d8e8/lib/components/normal-components/Jumper.ts#L42) file i think we also have to moved it from there also. image

@Anshgrover23 Yes you are right, I. Thanks, I Will create a quick pr for that

PR #396, opened 2024-12-07
fixing JSX type  module for React 19 types

Abse2001:
@seveibar is this the right approach I couldn't test it on snippet and after updating react types and fixing the type error I couldn't retest it even if I discard all changes I made


seveibar (commented):
This is the right approach, but if you look at that react-three PR they declare 3 different modules (eqch one is relevant for different tsconfig modes)

You cant test locally? Have you tried cmd shift p then "restart ts server"?

Abse2001:
@seveibar on core it looks like it works no more errors but I cant use yalc and overwrite the url in snippet ig

lib/fiber/intrinsic-jsx.ts
```diff
@@ -1,53 +1,60 @@
 import type * as Props from "@tscircuit/props"
 
-declare global {
+export interface TScircuitElements {
```
seveibar, line 3: Casing, so we use google style single word casing and tscircuit we generally consider a single word
```suggestion
export interface TscircuitElements {
```

seveibar (approved):

seveibar:
Yea with ATA testing is extremely difficult

Issue #387, opened 2024-12-05
Automatic schematic net labeling for passive-chip traces


seveibar:
We have a trace with a passive on one side and a chip on the other side we it's eligible to be automatically turned into a net label style schematic trace it will be automatically converted if there are more than 5 complex trace elements: junction, crossings or turns

The net label to use is the label of the pin on the chip

e.g. C15 should "auto net label" to "ADC_VDDIO"

![Image](https://github.com/user-attachments/assets/ea6ef02c-c2e6-46ae-8970-0c5bb626b227)

/bounty $30

AnasSarkiz:
![Image](https://github.com/user-attachments/assets/42837861-3d09-4dc9-97ac-1c3b3644f948)

seveibar:
That looks right!!! @AnasSarkiz (nice!!!)

Can i make a small adjustment, can we make the netlabels have the refdes of the chip? So "U1_OUT1" instead of "OUT1"? Should help prevent "collisions"

Nicely done!

AnasSarkiz:
@seveibar 

![Image](https://github.com/user-attachments/assets/a68b6fb8-56e6-4188-9cdd-7f3d65796cea)

seveibar:
Looks good although im surprised some of the long traces for those resistors arent converting (maybe we need to tweak conditions? Maybe 4 turns should convert?

AnasSarkiz:
How about a chip with four traces in the same face, and each trace has more than four turns? should we replace all of them with a netlabel? because I thought that we must have more than five traces on each face

seveibar:
@AnasSarkiz lets not consider faces yet, im looking a bit closer and i think even traces with more than 5 turns arent getting converted right now (am i wrong)

Im not opposed to the face thing, but if we can keep it simple thats better

seveibar:
Can i see it reduced to 4 turns? Maybe after we get that working we could do the face logic as an extension? Curious about the code complexity

AnasSarkiz:
@seveibar I modified the logic if one of the condition get true `complex trace's index is >= 5 for on the same face` **or** `the trace has _countComplexElements >=4`  

Output

![Image](https://github.com/user-attachments/assets/b0cc3a95-e48c-4173-9256-ac4e182c970f)

![Image](https://github.com/user-attachments/assets/43a2e578-dd4d-42c1-9a37-fcca767f1b4c)

[...]

PR #397, opened 2024-12-07
Automatic schematic net labeling for passive-chip for complex traces

AnasSarkiz:
/claim #387


seveibar (commented):
Can we remove code related to the sides, just to simplify the files

lib/components/primitive-components/Trace/Trace.ts
```diff
@@ -892,5 +939,60 @@ export class Trace
...
+        db.schematic_net_label.insert({
+          text: fullLabel,
+          source_net_id: passiveSidePosition.port.source_port_id!,
+          anchor_position: passiveSidePosition.position,
+          center: passiveSidePosition.position,
+          anchor_side:
+            getEnteringEdgeFromDirection(
+              passiveSidePosition.facingDirection!,
+            ) ?? "bottom",
+        })
```
seveibar, line 990: Use the existing function for creating the net labels, just modify it to derive the net label name if it doesnt have schDisplayLabel

Dont allow the schematic trace to be inserted if its complex

Moving the counting function outside this class, its not stateful, you give it the path and it returns the complexElementCount

seveibar (changes requested):

lib/components/primitive-components/Trace/Trace.ts
AnasSarkiz: modify the `_doInitialSchematicTraceRenderWithDisplayLabel` method and add the logic to derive the netlabel name if it doesn't have `schDisplayLabel`
Screenshot 2024-12-08 at 9 03 45 AM

lib/components/primitive-components/Trace/Trace.ts
```diff
@@ -658,18 +659,33 @@ export class Trace
...
+      firstPort.parent?.config.shouldRenderAsSchematicBox
+    const isSecondPortSchematicBox =
+      secondPort.parent?.config.shouldRenderAsSchematicBox
+    const firstPortName = isFirstPortSchematicBox
+      ? firstPort?.props.name
+      : `${secondPort?.parent?.props.name}_${secondPort?.props.name}`
+    const secondPortName = isSecondPortSchematicBox
+      ? secondPort?.props.name
+      : `${firstPort?.parent?.props.name}_${firstPort?.props.name}`
+
```
seveibar, line 686: The names should always be identical (its just a schematic convention for net labels), if both parents are schematic boxes we could do a joint name (“U1_SDA_U2_SCL”)

i know thats pretty gross/long but i cant think of a better solution for two boxes at the moment.

seveibar (commented):
AWESOME, but the matching labels is important, i was co fused pokking at the snapshot with PWR1 because the net labels dont match and that is essential

lib/components/primitive-components/Trace/Trace.ts
AnasSarkiz: @seveibar  The code is currently functioning only for passive-to-chip connections, as indicated by the issue. Should I modify it to handle all trace types, including passive-to-passive and chip-to-chip connections as well?

AnasSarkiz: @seveibar

AnasSarkiz:
> AWESOME, but the matching labels is important, i was confused poking at the snapshot with PWR1 because the net labels dont match and that is essential

@seveibar 
Screenshot 2024-12-09 at 6 17 01 PM

the name of the port is `pin2` with aliases `FB` maybe be we need to make some changes

Screenshot 2024-12-09 at 6 15 14 PM
Screenshot 2024-12-09 at 6 15 26 PM

seveibar:
@AnasSarkiz no the issue is one is `PWR_pin2` and the other is `pin2`, they should both be `PWR_pin2`

lib/components/primitive-components/Trace/Trace.ts
seveibar: no passive to chip is fine to start!

seveibar: but the labels should be the same on each side!

AnasSarkiz:
@seveibar 
now, both sides have the same name of the netlabel


Screenshot 2024-12-10 at 5 33 41 PM

PR #405, opened 2024-12-10
fix: rendered circuit json passed to the autorouter

imrishabh18:
Actually width and height now being passed to the autorouter and not the default ones's.

### DIFF
--- a/lib/components/primitive-components/Group/Group.ts
+++ b/lib/components/primitive-components/Group/Group.ts
@@ -152,7 +152,7 @@ export class Group<Props extends z.ZodType<any, any, any> = typeof groupProps>
           {
             method: "POST",
             body: JSON.stringify({
-              input_circuit_json: this.root!.db.toArray(),
+              input_circuit_json: this.root!.getCircuitJson(),
             }),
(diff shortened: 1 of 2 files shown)


lib/components/primitive-components/Group/Group.ts
```diff
@@ -152,7 +152,7 @@ export class Group<Props extends z.ZodType<any, any, any> = typeof groupProps>
           {
             method: "POST",
             body: JSON.stringify({
-              input_circuit_json: this.root!.db.toArray(),
+              input_circuit_json: this.root!.getCircuitJson(),
             }),
```
seveibar, line 156: GetCircuitJson should only be used by users, not internally (because it can cause rerenders, but db.toArray cant)

imrishabh18: But db.toArray has the default width and height, and not the processed width and height after the calculation

seveibar: That’s a render phase issue then, we still should not call getCircuitJson. We can eg approximate the board size in a render phase prior to autorouting

PR #422, opened 2024-12-15
feat:schematic_net_label.symbol_name for GND symbol.

### LINKED ISSUE #388: Implement `schematic_net_label.symbol_name`
Test that we're able to make a net label have the ground symbol

Requires PRs across props, core, circuit to svg, circuit json

If the net label is "GND" automatically use a GND symbol. Still should label the ground because there can be multiple GNDs, may requires adding label text to schematic symbols

/bounty $60

Abse2001:
needs a circuit-to-svg PR to work properly
/claim #388
/closes #388

image

pushed the symbols away from the sch box so it don't collide with netlabels


seveibar:
@Abse2001 is it possible for you to make the PR to circuit-to-svg first, we try to make core "the least broken" when there are dependent PRs

lib/components/primitive-components/Trace/Trace.ts
```diff
@@ -596,6 +596,157 @@ export class Trace
       }
     }
   }
+  __doInitialSchematicTraceRenderWithNetLabelSymbol(
```
seveibar, line 599:
```suggestion
  _doInitialSchematicTraceRenderWithDownwardSymbol(
```

seveibar: - We don't want to do this for all schematic symbols, just "downward symbols" like ground, we still want the old handling for other symbols
- convention is `_` not `__` (minor)

Abse2001:
> @Abse2001 is it possible for you to make the PR to circuit-to-svg first, we try to make core "the least broken" when there are dependent PRs

ofc

lib/components/primitive-components/Trace/Trace.ts
```diff
@@ -596,6 +596,157 @@ export class Trace
...
+      ],
+      junctions: [
+        {
+          x: toAnchorPos.x + toHorzPortDirectionOffset * toSchBoxHorzOffset,
+          y: toAnchorPos.y + toVertPortDirectionOffset * toSchBoxVertOffset,
+        },
+      ],
+      source_trace_id: this.source_trace_id!,
+    })
+  }
```
seveibar, line 749: this handling has a lot of math (the math looks pretty good) but I wonder if we could try to move it out of `Trace.ts`, possibly turning `Trace.ts` into a directory `Trace/Trace.ts` and `Trace/createDownwardNetLabelSymbol.ts`

You can use the "context passing pattern" to keep the function basically identical:

```tsx
/**
 * Write a good doc string, it's not obvious at a glance what fromPortAnchor or toPortAnchor is!
 **/
export const createDownwardNetLabelSymbol({ fromPort, toPort, fromPortAnchor, toPortAnchor }, {
 db
})
```

When you use this type of signature, you can do...

`createDownwardSymbolNetLabelSymbol({ toPort, fromPort, ... }, this)`

NOTE: We never use more than two parameters in a function, whenever there are more than two parameters we either convert to "named parameters" and take a single parameter as an object, or we switch to the context passing pattern.

When using the context-passing pattern, the most important thing is to make sure you're not inventing too many contexts. Having "{ db }" as the context is very standard/normal so it's an easy decision in this case!

seveibar (commented):
looking great! /tip $10

lib/components/primitive-components/Trace/Trace.ts
```diff
@@ -660,17 +662,33 @@ export class Trace
...
+      createDownwardNetLabelSymbol(
+        {
+          fromPort,
+          toPort,
+          fromAnchorPos,
+          toAnchorPos,
+          schDisplayLabel: this.props.schDisplayLabel!,
+          source_trace_id: this.source_trace_id!,
+        },
+        db,
```
seveibar, line 689: The context has to be a named object

```suggestion
        { db },
```

lib/components/primitive-components/Trace/Trace.ts
```diff
@@ -697,6 +705,14 @@ export class Trace
...
         getEnteringEdgeFromDirection(toPort.facingDirection!) ?? "bottom",
     })
+    db.schematic_net_label.insert({
+      text: this.props.schDisplayLabel! ?? pinFullName,
+      source_net_id: fromPort.source_port_id!,
+      anchor_position: fromAnchorPos,
+      center: fromAnchorPos,
+      anchor_side:
+        getEnteringEdgeFromDirection(fromPort.facingDirection!) ?? "bottom",
+    })
```
seveibar, line 715: try not to include changes that don't do anything 😉 , makes me have to 🤔

lib/components/primitive-components/Trace/Trace.ts
```diff
@@ -777,6 +793,29 @@ export class Trace
           connectedTo: [],
         })
       }
+      if (
+        elm.type === "schematic_net_label" &&
+        (elm.text.toLowerCase().includes("gnd") ||
+          elm.text.toLowerCase().includes("ground"))
```
seveibar, line 799: We should adapt this code to derive the symbol width dynamically, but we can do this in a future issue, I'm leaving this comment so that you can use "..." -> "Create Issue" after this is merged!

```suggestion
        elm.type === "schematic_net_label" &&
        elm.symbol_name
```

lib/components/primitive-components/Trace/create-downward-net-label-symbol.ts
```diff
@@ -0,0 +1,165 @@
...
+  {
+    fromPort,
+    toPort,
+    fromAnchorPos,
+    toAnchorPos,
+    schDisplayLabel,
+    source_trace_id,
+  }: CreateDownwardNetLabelSymbolParams,
+  db: SoupUtilObjects,
+): void => {
```
seveibar, line 23: ok so I'm able to read this a bit more specifically now,

it seems like you've made this code specific to "ground" so we should actually rename the function for "ground" i.e. `createDownwardGroundSymbol`

There are many many "magic numbers" here, these should _all_ be derived from the symbol data from `schematic symbols` so that this method can be generalized to any symbol and generally it's just best to avoid any magic numbers, every number should be "derived from" something

The ternaries in this function are very hard to read, the function also has duplicated logic for both the `from` and the `to`, each of these parts should be broken out to avoid repeating essentially the same logic.

In general, you shouldn't have nested ternaries. You can try the following pattern to simplify these and make them more readable:

```ts
const schBoxOffset = {
  "left": { x: -0.5, y: 0 },
  "top": { x: 0, y: 0.5 },
  "right": { /* ... */ },
  "bottom": {/* ... */}
}[port.facingDirection]
```

etc.

seveibar (changes requested):
nice, ok structurally we're in a good place but more cleanup to make maintainable /tip $10

lib/components/primitive-components/Trace/create-downward-net-label-symbol.ts
seveibar: @Abse2001 you can still use the magic numbers if you rename the function to be specific to ground! I think I typo'd a bit- only the "general purpose downward symbol function" shouldn't have magic numbers

Abse2001:
@seveibar I will test if we can make it a fn that takes any symbol without using magic numbers.

seveibar (approved):
awesome

seveibar:
@Abse2001 want to do that work as a future bountied PR? Create an issue and tag me or @imrishabh18 to bounty!

PR #430, opened 2024-12-17
Add type for webworker

imrishabh18:
Issue - https://github.com/tscircuit/snippets-cli/issues/7


seveibar (commented):
Why do we need this one? It's not an eval_error if we're not evaluating unsafe code, it would be called something else like a render error, but ideally we don't have an error that broad here (I think this case is covered by the eval-webworker anyway?)

imrishabh18:
Yeah correct, this thing is covered by web worker

PR #431, opened 2024-12-17
feat: move non-subcircuit group

### LINKED ISSUE #390: Non-subcircuit Groups don't properly offset schematic components, can cause traces not to render
https://github.com/user-attachments/assets/73f24fe9-dde2-45cf-a0c9-0ba99934cb29

/bounty $40

imrishabh18:
/close #390


seveibar (commented):
Not a bad implementation but i have a slightly different one in mind that i think coupd be better- each parent returns it’s “transform” to its children (something like _computeSchematicTransform), this is better than offsetting in a later render phase.

I can give it a try shortly, feel free to take a stab!

seveibar:
Basically _computeSchematicPositionBeforeLayout should return the correct schematic position for children (offset by all parent groups)

imrishabh18:
@seveibar This is what you mean't by having another function to compute the offset for the decedent's?

lib/components/primitive-components/Group/Group.ts
```diff
@@ -308,18 +308,48 @@ export class Group<Props extends z.ZodType<any, any, any> = typeof groupProps>
     }
   }
 
+  _computeSchematicPositionBeforeLayout(): void {
+    const { db } = this.root!
```
seveibar, line 312: This method shouldnt mutate, never mutate db unless you're definitely inside a render phase and the name of the method makes sense to mutate

seveibar (changes requested):

seveibar:
@imrishabh18 i'll record a video for you

PR #435, opened 2024-12-19
fix: overlap trace with pin length

### LINKED ISSUE #434: Stop routing through schematic box arms
This should be simple, just adding an obstacle to ensure we don't route between the arms

Image

imrishabh18:
/close #434


tests/components/normal-components/__snapshots__/chip-pins-not-overlapped-with-traces-schematic-schematic.snap.svg
imrishabh18, line 1: @seveibar Will need your help in this. Can you take a look at why the trace connecting the port 8 on the right side is not generating? I have created the obstacle on the extension of chip ports

Issue #389, opened 2024-12-05
Support four-pin crystal


seveibar:
Looks like this:

![Image](https://github.com/user-attachments/assets/c1105379-674b-4356-99bf-86ce7de4592e)

- [x] tscircuit/props#112
- [ ] Enable using the 4pin variant (we already have a symbol for it, it's called "crystal_4pin_horz"

/bounty $15

PR #440, opened 2024-12-20
add 4pin crystal

techmannih:
4pinCrystal
/claim #389

### DIFF
--- a/lib/components/normal-components/Crystal_4Pin.ts
+++ b/lib/components/normal-components/Crystal_4Pin.ts
@@ -26,10 +26,10 @@ export class Crystal4Pin extends NormalComponent<
   initPorts() {
     super.initPorts({
       additionalAliases: {
-        pin1: ["pos", "top_left"],
-        pin2: ["neg", "top_right"],
-        pin3: ["cap1", "bottom_left"],
-        pin4: ["cap2", "bottom_right"],
+        pin1: ["pos", "left"],
+        pin2: ["neg", "right"],
+        pin3: ["cap1", "top"],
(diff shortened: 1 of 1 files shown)


lib/components/normal-components/Crystal_4Pin.ts
```diff
@@ -0,0 +1,52 @@
...
+    }
+  }
+
+  initPorts() {
+    super.initPorts({
+      additionalAliases: {
+        pin1: ["pos", "top_left"],
+        pin2: ["neg", "top_right"],
+        pin3: ["cap1", "bottom_left"],
+        pin4: ["cap2", "bottom_right"],
```
imrishabh18, line 32: The alias for this probably should be `top`, `bottom`, `left` and `right`, according to the symbol image

lib/components/normal-components/Crystal_4Pin.ts
imrishabh18, line 1: I think probably we should use the `<Crystal />` component already present, and not introduce a new component just for when we are having 4pins?

cc: @seveibar

seveibar: yes

seveibar (changes requested):
see rishabh's comment

imrishabh18:
@techmannih Please address the comments before asking for the review's

lib/components/normal-components/Crystal_4Pin.ts
seveibar: No modify the existing Crystal to support the pinVariant prop

PR #469, opened 2024-12-30
testing elkjs bundle size

ShiboSoftwareDev:
this is just testing the bundle size

I also discovered some issue, If I remove the workerUrl to try to run it synchronously it breaks here's a related issue: https://github.com/oven-sh/bun/issues/15737


ShiboSoftwareDev:
@seveibar this library is huge

ShiboSoftwareDev:
@seveibar looks like it needs to be run with node not bun

seveibar:
oh wow yea that's big. I'm a bit hesitant to pull in something that big...

ShiboSoftwareDev:
@seveibar I'll do a bunch if tests and configure the algorithm properly and see how it does

ShiboSoftwareDev:
it's pretty good for placements but you'd have to manage your own traces
image

ShiboSoftwareDev:
what I did here is I made the doInitialSchematicTraceRender wait on the layout to be complete
image

ShiboSoftwareDev:
@seveibar i can go forward and configure the algorithm and make more circuits but I want to know what you say about the traces first

seveibar:
@ShiboSoftwareDev im not sure, i think we need a more representative circuit. Something like the esp32 that Anas put together

seveibar:
This one would also work, its tricky because it uses other snippets that cant be loaded into a test case, but this is the kind of layout that is pretty common and important for the autolayout engine to do properly https://tscircuit.com/seveibar/rp2040-module

seveibar:
This is the one i mentioned from Anas that's a pretty good representative layout, can't find that actual snippet though 😭 CC @AnasSarkiz 

image

AnasSarkiz:
@seveibar  https://tscircuit.com/anas-sarkez/esp32-modul

ShiboSoftwareDev:
@seveibar I'll yalc push core to snippets and test with the auto-load snippets feature

PR #481, opened 2025-01-05
implement mosfet

### LINKED ISSUE #462: implement P-Channel MOSFET
I think it's the last component needed for the wifi camera module, very important

Image

ShiboSoftwareDev:
/claim #462
/closes #462


seveibar (approved):
i think this is the best implementation, but i should note that mosfet perhaps technically should extend transistor- but tbh i can't see where that would be relevant in the near future and it's simpler to not have to worry about inheritance that much

I think the rule for inheritance should be that inheritance should only be used for things relating directly to the render process, not the "real world inheritance" of the objects. If we keep it simple there shouldn't be "mixin-style" issues later.

Make sure to fix that snapshot and ping me if it looks weird

nice work 👍

PR #485, opened 2025-01-06
Implement pcbOffsetX and pcbOffsetY in <board />

### LINKED ISSUE #484: Implement `<board />` `outlineOffsetX` and `outlineOffsetY`
as per props defined in https://github.com/tscircuit/props/pull/143

/bounty $4

techmannih:
fixes #484 
/claim #484


docs/CIRCUIT_JSON_PCB_OVERVIEW.md
```diff
@@ -344,4 +344,8 @@ export interface PcbBoard {
   center: Point
   outline?: Point[]
 }
+type PcbBoardWithOffsets = PcbBoard & {
```
seveibar, line 347:
```suggestion
  outline?: Point[]
```

btw this file is supposed to be replaced with tscircuit plop (npm install -g @tscircuit/plop then run plop)

lib/components/normal-components/Board.ts
```diff
@@ -4,6 +4,23 @@ import { NormalComponent } from "../base-components/NormalComponent"
...
+    pcb_board_id: string
+    thickness: number
+    num_layers: number
+    outline?: { x: number; y: number }[]
+  },
+  "type" | "pcb_board_id"
+> & {
+  pcbOffsetX: number
+  pcbOffsetY: number
+}
```
seveibar, line 22: nooooo no other files do this, why would you do this pattern?

seveibar (changes requested):
no test, and a bunch of irrelevant changes, 

@techmannih read over your work, make sure there aren't irrelevant changes, check the patterns and always start by writing a test that makes sure YOU understand what you're supposed to be implementing

with no test it's not worth requesting a review. I've seen a couple of these at this point from you and it takes up my time, this one is really really bad though. Why wouldn't you write a test? Write a snapshot PCB test, that will demonstrate to you that it doesn't work.

PR #486, opened 2025-01-06
added symbolName to chip component

Abse2001:
/cliam https://github.com/tscircuit/snippets/issues/314

@seveibar Im doing this PR while it is not 100% ready just to make sure that this is what you meant in the issue above


seveibar (commented):
This is an interesting idea but not what i meant, theoretically a user should be able to create use a `<pushbutton />` element for an actual pushbutton

i think it doesnt work because in practice push buttons are 4 pins with two being internally connected. So its unclear what the behavior should be (but a test could clarify it)

PR #487, opened 2025-01-06
Implemented <board /> outlineOffsetX and outlineOffsetY

Abse2001:
/claim #484
@seveibar is this ok or should I move the offset to pcbviewer?


seveibar (commented):
the board center should also adjust

tests/examples/example16-board-outline-offset-2.test.tsx
```diff
@@ -0,0 +1,15 @@
+import { expect, test } from "bun:test"
+import { getTestFixture } from "../fixtures/get-test-fixture"
+
+test("Board outline offset", async () => {
```
seveibar, line 4:
```suggestion
test("Board offsetOutline without output defined", async () => {
```

seveibar (approved):
fix snapshot/test name, then good to go!

PR #495, opened 2025-01-09
initial: cloud autorouter on group

### LINKED ISSUE #453: support changing the autorouter in subcircuits
I need this feature to create a complicated module
Image

ShiboSoftwareDev:
I don't know how to approach this, the problem is that zod doesn't register the autorouter prop since it is not defined in groupProps but in subcircuitGroupProps

I also need to update the types in props repo that's why I have as any in the test but that's not a difficult

/claim #453
/closes #453


lib/components/primitive-components/Group/Group.ts
```diff
@@ -38,7 +39,7 @@ export class Group<Props extends z.ZodType<any, any, any> = typeof groupProps>
 
   get config() {
     return {
-      zodProps: groupProps as unknown as Props,
+      zodProps: subcircuitGroupProps as unknown as Props,
```
seveibar, line 42: Could we use z.union here? (Maybe itll mess something up?)

tests/features/remote-autorouting-6-cloud-autorouting-on-group.test.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { getTestFixture } from "../fixtures/get-test-fixture"
+import { getTestAutoroutingServer } from "tests/fixtures/get-test-autorouting-server"
+
+test("remote-autorouter-6 cloud autorouting on group", async () => {
+  if (process.env.CI) return
+  const { circuit } = getTestFixture()
+
+  // Create a basic circuit that needs routing
+  circuit.add(
+    <group autorouter={"auto-cloud" as any}>
```
seveibar, line 11:
```suggestion
    <group autorouter={"auto-cloud"} subcircuit />
```

lib/components/primitive-components/Group/Group.ts
seveibar: Actually it should be conditional on the subscircuit prop

ShiboSoftwareDev:
@seveibar it worked lol, that was unclear and shouldn't it be a subcircuit by default?

seveibar:
Well subcircuit has really big effects and a group is more or less a container that is harmless by default. We are going to introduce a `<subcircuit />` type though

ShiboSoftwareDev:
@seveibar I guess you can close the issue now

PR #512, opened 2025-01-11
uprade ->react 19

bhuvan-crypto:
/claim 510

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,4737 @@
+{
+  "name": "@tscircuit/core",
+  "version": "0.0.261",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
+      "name": "@tscircuit/core",
+      "version": "0.0.261",
+      "dependencies": {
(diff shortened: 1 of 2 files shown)


Anshgrover23 (changes requested):
@bhuvan-crypto do you think changing the version of react in package.json solves the issue??

bhuvan-crypto:
hii @Anshgrover23 I am new to open source I think there is a need to upgrade the version and I am not sure if there is anything needed other than that

Anshgrover23:
@bhuvan-crypto welcome to open source, but this issues does not solve by only updating package.json, fist try to find some other issues which are good first issues, you can find some of them in `tscircuit/schematic-symbols` repository.

seveibar (changes requested):
tests not passing

seveibar:
@bhuvan-crypto make sure not to commit the package-lock as well.

This issue is kind of hard, you need to understand react-reconcilation changes for react 19

seveibar:
as @Anshgrover23 is saying, would recommend an easier issue

Issue #510, opened 2025-01-11
Upgrade React from 18 to 19


seveibar:
Currently if a user is using core they have to use React 18, lets make sure we work in both react 18 and 19

/bounty $20

bhuvan-crypto:
/attempt #510
react upgrade done

seveibar:
fixed in https://github.com/tscircuit/core/pull/532

PR #522, opened 2025-01-14
Upgrade React 18 -> 19

yashksaini-coder:
The PR includes a dependency update for the `react` library in the `package.json` file.

Dependency update:

* [`package.json`](diffhunk://#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L55-R55): Updated the `react` dependency from version `^18.3.1` to `^19.0.0`.

**Closes**:- #510

/claim #510

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
     "format-si-unit": "^0.0.2",
     "nanoid": "^5.0.7",
     "performance-now": "^2.1.0",
-    "react": "^18.3.1",
+    "react": "^19.0.0",
     "react-reconciler": "^0.29.2",
     "schematic-symbols": "^0.0.113",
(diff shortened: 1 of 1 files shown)


Abse2001:
@yashksaini-coder it is about updating react you need to refactor the repo to work in react 19 you can see that the whole repo is broken now most of the tests are failling

Anshgrover23 (changes requested):
@yashksaini-coder i think first you need to see the issue and then try to solve it do not rush to make PR, as you can see one pr is closed on this issue do the same reason.
- Changing the **Package.json** file does not solves this issue.

yashksaini-coder:
Okay understood

PR #527, opened 2025-01-15
Automatically add net labels when a chip-to-chip trace

### LINKED ISSUE #524: Automatically add net labels when a chip-to-chip trace fails to route, use the name of the larger chip for the net label
Image

https://tscircuit.com/seveibar/pico-w-3x5-led-matrix

/bounty $30

techmannih:
fixes #524 
claim #524


seveibar (changes requested):
can you simplify the test case (make it as simple as possible, just one trace!)

can you make sure your test has the correct behavior:

test with a bigger chip (8 pins) and a small chip (6 pins) and make sure that the net label uses the name of the bigger chip.

e.g. if U1 is the bigger chip, the net label should prefer to name itself after the big chip

thank you!!!

seveibar (approved):
Ok cool.

if you want a new issue, im curious what would happen if you removed the pinlabels from one chip. It could break- which would be a great new issue!!

PR #528, opened 2025-01-16
testing elkjs

ShiboSoftwareDev:


lib/components/base-components/PrimitiveComponent.ts
```diff
@@ -279,7 +279,8 @@ export abstract class PrimitiveComponent<
    * schematic components
    */
   computeSchematicPropsTransform(): Matrix {
-    return compose(translate(this.props.schX ?? 0, this.props.schY ?? 0))
+    // need to find a way to make schX and schY 0 before autolayout and apply them after autolayout
+    return compose(translate(0, 0))
```
seveibar, line 283: i talked about this on stream today at the 25m point, drew a diagram

https://x.com/seveibar/status/1879928186630443342

image


https://github.com/tscircuit/core/issues/529

lib/components/primitive-components/Group/Group.ts
```diff
@@ -309,44 +310,21 @@ export class Group<Props extends z.ZodType<any, any, any> = typeof groupProps>
...
-                p.schematic_component_id === component.schematic_component_id,
-            )
-
-          components.push(component)
-          ports.push(...schPorts)
-        }
+    this._queueAsyncEffect("schematic-layout", async () => {
+      try {
+        await layoutSchematicWithElk(db)
+        this._markDirty("SchematicLayout")
```
seveibar, line 323: nice

lib/utils/schematic/elkAutoLayout.ts
```diff
@@ -0,0 +1,161 @@
...
+  }>
+}
+
+export async function layoutSchematicWithElk(db: SoupUtilObjects) {
+  const components = db.schematic_component.list()
+  const sourceTraces = db.source_trace.list()
+  const ports = db.schematic_port.list()
+
+  const elk = new Elk({
+    workerUrl: "./node_modules/elkjs/lib/elk-worker.min.js",
```
seveibar, line 25: we'll need to load this via blobUrl probably (unfortunately)

will require bundling step, similar to @tscircuit/eval-webworker

I'm thinking maybe we create a separate package for this just to simplify where this code lives and to make as many test cases as possible, but not sure

seveibar (commented):
this is great, so my thought on this is, we might not want to remove the old stuff which I know how to improve versus LGS, which I don't know the limitations of we might wanna create an MPM repo that specifically test the elk JS can layout different. Things are almost have a repo that's just filled with tests that we can use to test different ways to use help ass so this is my this is my thought.

PR #536, opened 2025-01-20
feat: group waits until the subcircuit is routed

### LINKED ISSUE #535: `Group.ts` should wait until subcircuits are fully routed before attempt to route traces
https://tscircuit.com/seveibar/subcircuit-routing-test

imrishabh18:
/close #535


lib/components/primitive-components/Group/Group.ts
```diff
@@ -112,7 +104,7 @@ export class Group<Props extends z.ZodType<any, any, any> = typeof groupProps>
     }
   }
 
-  doInitialPcbTraceRender() {
+  async doInitialPcbTraceRender() {
```
seveibar, line 107: render functions can't be async

```suggestion
  doInitialPcbTraceRender() {
```

lib/components/primitive-components/Group/Group.ts
```diff
@@ -121,14 +113,20 @@ export class Group<Props extends z.ZodType<any, any, any> = typeof groupProps>
     const serverMode = this.props.autorouter?.serverMode ?? "job"
 
     const debug = Debug("tscircuit:core:autorouting")
+    debug("props", this.props.autorouter)
```
seveibar, line 116:
```suggestion
    debug("props.autorouter", this.props.autorouter)
```

lib/components/primitive-components/Group/Group.ts
```diff
@@ -121,14 +113,20 @@ export class Group<Props extends z.ZodType<any, any, any> = typeof groupProps>
     const serverMode = this.props.autorouter?.serverMode ?? "job"
 
     const debug = Debug("tscircuit:core:autorouting")
```
seveibar, line 115: initialize this at the top, don't reininitialize inside render functions

lib/components/primitive-components/Group/Group.ts
```diff
@@ -121,14 +113,20 @@ export class Group<Props extends z.ZodType<any, any, any> = typeof groupProps>
...
       debug("fetching", url)
       return fetch(url, options)
     }
 
-    // Queue the autorouting request
-    this._queueAsyncEffect("make-http-autorouting-request", async () => {
+    // Wait for any existing autorouting to complete
+    if (Group.currentAutoroutingPromise) {
+      await Group.currentAutoroutingPromise
+    }
```
seveibar, line 126: i get the idea with this method, but unrelated subcircuits should be able to go in parallel

lib/components/primitive-components/Group/Group.ts
```diff
@@ -218,6 +207,8 @@ export class Group<Props extends z.ZodType<any, any, any> = typeof groupProps>
           this._asyncAutoroutingResult = {
             output_pcb_traces: autorouting_job_output.output_pcb_traces,
           }
+          // Apply the routing results immediately before the next autorouting job starts
+          this._updatePcbTraceRenderFromPcbTraces()
```
seveibar, line 211: we're not suppose to touch circuit json while inside of an effect (this is a rule of react, but unfortunately we don't enforce it currently)

seveibar (commented):
i think this is a pretty good attempt, but i'm curious if we can try an approach a bit more like `group.areAnyOfMyChildrenAutorouting()` because that would allow for more parallelism

We can't really use statics, it would need to be on the RootCircuit because different tscircuit/core instances would interact if we use statics

lib/components/primitive-components/Group/Group.ts
imrishabh18: Yeah, I knew we can't async the renderPhases. But this was the way to synchronise the two calls to the autorouter

PR #537, opened 2025-01-20
feat: pcb_group and source_group getting added

imrishabh18:


lib/components/primitive-components/Group/Group.ts
```diff
@@ -65,6 +55,56 @@ export class Group<Props extends z.ZodType<any, any, any> = typeof groupProps>
...
+      })
+      if (currentComponent) {
+        db.pcb_smtpad.update(currentComponent.pcb_smtpad_id, {
+          pcb_group_id: pcb_group.pcb_group_id,
+          subcircuit_id: this.isSubcircuit
+            ? `subcircuit_${this._renderId}`
+            : undefined,
+        })
+      }
+    }
```
imrishabh18, line 105: @seveibar As the child primitive components are getting rendered first and then the `Group`, so updating the primitive components like `smtpad` in this case after the `Group` is getting rendered is the correct approach?

seveibar: the smtpad when it's created should just add it's subcircuit_id

seveibar: Each render function runs for children first, but render phases run for ALL components before moving on to the next render phase

PR #541, opened 2025-01-21
feat: route child subcircuit and then the parent

imrishabh18:
- Route the child subcircuits and wait until that is rendered and then route the Parent


lib/components/primitive-components/Group/Group.ts
```diff
@@ -147,15 +149,32 @@ export class Group<Props extends z.ZodType<any, any, any> = typeof groupProps>
...
 
-    const debug = Debug("tscircuit:core:autorouting")
+    // Find all child groups that are subcircuits
+    const subcircuits = this.children.filter(
+      (child) => child instanceof Group && child.isSubcircuit,
+    ) as Group[]
+
+    const subcircuitPromises = subcircuits.map((subcircuit) =>
+      subcircuit._queueAutorouting(serverUrl, serverMode, debug),
+    )
```
seveibar, line 159: this isn't quite right either, a subcircuit should only be responsible for itself

PR #559, opened 2025-01-25
test: verify footprinter soic4 using autocloud

imrishabh18:
New core outputs the correct circuit json for soic4 footprint, but the current snippets webworker or some other service using old version of core most probably

https://registry-api.tscircuit.com/admin/autorouting_jobs/get?autorouting_job_id=6948acc4-db57-4064-8415-9b793a24f10b (circuit json output which is given by snippets)

cc: @seveibar


seveibar:
@imrishabh18 not sure i understand

PR #565, opened 2025-01-25
support outline without width and height

### LINKED ISSUE #564: When a board has an outline, width and height should not be required
Image

/bounty $10

techmannih:
fixes #564 
/claim #564

### DIFF
--- a/lib/components/normal-components/Board.ts
+++ b/lib/components/normal-components/Board.ts
@@ -35,11 +35,12 @@ export class Board extends Group<typeof boardProps> {
     if (this.root?.pcbDisabled) return
     const { db } = this.root!
     const { _parsedProps: props } = this
-
-    if (!props.width || !props.height) {
+  
+    // If outline is not provided, width and height must be specified
+    if (!props.outline && (!props.width || !props.height)) {
       throw new Error("Board width and height or an outline are required")
     }
-
+  
     const pcb_board = db.pcb_board.insert({
       center: {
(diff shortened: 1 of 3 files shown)


lib/components/normal-components/Board.ts
```diff
@@ -49,8 +50,8 @@ export class Board extends Group<typeof boardProps> {
       thickness: this.boardThickness,
       num_layers: this.allLayers.length,
 
-      width: props.width,
-      height: props.height,
+      width: props.width ?? 0,
+      height: props.height ?? 0,
```
seveibar, line 54: Dont default to 0, compute the width and height from the outline if its not provided

seveibar (changes requested):
See comment

seveibar (approved):
nice

tests/components/normal-components/board-outline.test.tsx
```diff
@@ -0,0 +1,24 @@
...
+      ]}
+    >
+      <resistor name="R1" resistance="10k" footprint="0402" />
+      <capacitor name="C1" capacitance="10uF" footprint="0603" />
+    </board>,
+  )
+
+  circuit.render()
+
+  expect(circuit).toMatchPcbSnapshot(import.meta.path)
```
seveibar, line 23: this would be slightly better if you did `expect(circuit.db.pcb_board.list()[0]).toMatchInlineSnapshot()

PR #566, opened 2025-01-26
add check to only comment if changed size > 10kb

### LINKED ISSUE #542: Modify the `bundle-size.yml` action to only comment if the PR has changed the size of the package by 10KB+
https://github.com/tscircuit/core/blob/main/.github/workflows/bundle-size.yml

/bounty $10

Anshgrover23:
/claim #542
Fixes #542

@seveibar 
- check this my forked repo pr for example.
https://github.com/Anshgrover23/core/pull/2

### DIFF
--- a/.github/workflows/bundle-size.yml
+++ b/.github/workflows/bundle-size.yml
@@ -89,7 +89,31 @@ jobs:
             echo "diff=$diff" >> $GITHUB_OUTPUT
           fi
 
+      - name: Check size threshold
+        id: size-threshold
+        run: |
+          # Get the differences from previous steps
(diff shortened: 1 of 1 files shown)


imrishabh18 (commented):
@Anshgrover23 Seems something is off, your PR on fork is not having any changes in the dependency only the workflow is added but still the comment showed up.

Anshgrover23:
@imrishabh18 sorry added wrong pr in description now check plss?

imrishabh18:
I can still see two size diff comments on that PR

imrishabh18:
@Anshgrover23  Maybe try to create a new clean PR, with one commit with no dependency change, then another commit with change with less that 10kb and last more that 10kb. That would verify it

PR #599, opened 2025-02-07
feat: Auto adjust board size to fit children components if width/height isn't provided

### LINKED ISSUE #562: Automatically adjust board size to fit children components if width/height isn't provided
/bounty $15

This may require introducing a new render phase

AbhinavTheDev:
/claim #562 
Fixes #562

### DIFF
--- a/lib/components/normal-components/Board.ts
+++ b/lib/components/normal-components/Board.ts
@@ -3,6 +3,7 @@ import type { z } from "zod"
 import { NormalComponent } from "../base-components/NormalComponent/NormalComponent"
 import { identity, type Matrix } from "transformation-matrix"
 import { Group } from "../primitive-components/Group/Group"
+import { getBoundsOfPcbComponents } from "lib/utils/get-bounds-of-pcb-components"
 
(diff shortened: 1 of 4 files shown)


lib/components/normal-components/Board.ts
```diff
@@ -31,6 +51,25 @@ export class Board extends Group<typeof boardProps> {
...
+    ) {
+      // console.log("Skipping auto-size - dimensions specified", this._parsedProps)
+      console.log("Skipping auto-size because dimensions are specified")
+      return
+    }
+
+    const newProps = getBoardSize(this._parsedProps, this.children)
+    // console.log("newProps",newProps);
+    if (newProps) {
+      this._parsedProps = newProps
```
seveibar, line 70: ? you're still setting _parsedProps, which is immutable

seveibar (changes requested):
you're not allowed to change props, you cannot change props (in any way)

you can create a function like `board.getSize()` and refactor the code to not use props where it would normally be getting the width/height or whatever

you can't change props, props are the user input, you can create new variables etc.

PR #604, opened 2025-02-08
add a subcircuit_id to pcb_Trace

### LINKED ISSUE #602: `pcb_trace` should be assigned a `subcircuit_id` (for the subcircuit it was created for)

Anshgrover23:
/claim #602
Fixes #602


seveibar (changes requested):
Need to handle when traces are added using the cloud autorouter (also best practice is to edit a test and add a quick check for subcircuit id!)

Issue #556, opened 2025-01-25
improve error message that occurs when you put something that is not a react component inside the jsx


ShiboSoftwareDev:
![Image](https://github.com/user-attachments/assets/e6adf3cf-1c6a-43ab-9c1b-331be9268a34)

seveibar:
the fix for this is likely in the react fiber layer

PR #609, opened 2025-02-09
patch: improve error message

ArnavK-09:
/claim #556

### Before 

image

### After
image

### DIFF
--- a/lib/components/base-components/PrimitiveComponent.ts
+++ b/lib/components/base-components/PrimitiveComponent.ts
@@ -466,6 +466,11 @@ export abstract class PrimitiveComponent<
   }
 
   add(component: PrimitiveComponent) {
+    if (!component.onAddToParent) {
+      throw new Error(
+        "Invalid JSX Element: Expected a React component but received a non-component value. Ensure you are using a valid component (function or class) and that it's properly imported.",
+      )
+    }
     component.onAddToParent(this)
     component.parent = this
     this.children.push(component)
--- a/lib/components/base-components/PrimitiveComponent.ts
+++ b/lib/components/base-components/PrimitiveComponent.ts
@@ -468,7 +468,7 @@ export abstract class PrimitiveComponent<
   add(component: PrimitiveComponent) {
     if (!component.onAddToParent) {
       throw new Error(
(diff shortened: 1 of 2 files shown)


Anshgrover23 (changes requested):
@ArnavK-09 can u show if this works using yalc package??

ArnavK-09:
> @ArnavK-09 can u show if this works using yalc package??

@Anshgrover23 

# BEFORE

image


# AFTER

image

lib/components/base-components/PrimitiveComponent.ts
```diff
@@ -466,6 +466,11 @@ export abstract class PrimitiveComponent<
   }
 
   add(component: PrimitiveComponent) {
+    if (!component.onAddToParent) {
+      throw new Error(
+        "Invalid JSX Element: Expected a React component but received a non-component value. Ensure you are using a valid component (function or class) and that it's properly imported.",
```
seveibar, line 471: can you attempt to stringify it to make the error message even better.

Also no need for the latter half

```suggestion
        "Invalid JSX Element: Expected a React component but received "...".
```

seveibar (commented):
nice work, can we improve the error message?

tests/components/primitive-components/invalid-jsx.test.tsx
```diff
@@ -0,0 +1,14 @@
...
+  const circuit = new Circuit()
+
+  expect(() => {
+    circuit.add(
+      <board width="10mm" height="10mm">
+        {1}
+      </board>,
+    )
+  }).toThrowError("Invalid JSX Element")
+})
```
seveibar, line 14: make sure to test with `new ClassName()` because i think JSON.stringify could error

seveibar (changes requested):
JSON.stringify fails sometimes no? we don't want to accidentally throw an unrelated error (I think)

tests/components/primitive-components/invalid-jsx.test.tsx
seveibar: or you could put in a function or something

seveibar (approved):
eh it's pretty hard to break JSON.stringify

PR #610, opened 2025-02-09
fix: autorouter is inhertited by the subcircuit child

imrishabh18:
issue - https://github.com/tscircuit/tscircuit.com/issues/649


lib/components/primitive-components/Group/Group.ts
```diff
@@ -469,6 +469,15 @@ export class Group<Props extends z.ZodType<any, any, any> = typeof groupProps>
...
    */
   _shouldUseTraceByTraceRouting(): boolean {
+    // Skip if the parent subcircuit is performing async autorouting
+    if (
+      this.parent &&
+      this.parent instanceof Group &&
+      this.parent._shouldRouteAsync()
+    ) {
+      return false
+    }
```
seveibar, line 479: not the right behavior, this is the correct behavior:

- The props define what the subcircuit should do FIRST
- If the props don't specify what you should do, THEN you look at your subcircuit, _not your parent_

seveibar: you can't do trace-by-trace autorouting until your children have finished autorouting, but that's outside the scope of this function and may already be implemented

lib/components/primitive-components/Group/Group.ts
```diff
@@ -469,10 +469,12 @@ export class Group<Props extends z.ZodType<any, any, any> = typeof groupProps>
    * or if using a "fullview" or "rip and replace" autorouting mode
    */
   _shouldUseTraceByTraceRouting(): boolean {
-    const props = this._parsedProps as SubcircuitGroupProps
-    if (props.autorouter === "auto-local") return true
-    if (props.autorouter === "sequential-trace") return true
-    if (props.autorouter) return false
+    // Inherit from parent if not set by props
+    const autorouter =
+      this._parsedProps.autorouter ?? this.getInheritedProperty("autorouter")
```
seveibar, line 474: ahhh very nice

seveibar (approved):

PR #625, opened 2025-02-13
add additional aliases to transistor ports

### LINKED ISSUE #581: connect to transistor terminals
pins should correspond to the transistor terminals: base, emitter and collector
same for the other similar components

Image

similar to the diode:
Image

Anshgrover23:
Fixes #581 
/claim #581


ShiboSoftwareDev (changes requested):
create a snapshot test connecting to a transistor port using the new port aliases for example: "T1 > emitter", "T1 > c", etc.

tests/components/normal-components/transistor.test.tsx
```diff
@@ -88,3 +89,57 @@ it("should render a PNP transistor with 270° rotation", async () => {
...
+  )
+  circuit.render()
+
+  const components = circuit.db.pcb_component.list()
+  expect(components).toHaveLength(1)
+  const component = components[0]
+
+  const ports = circuit.db.pcb_port.list()
+
+  const base = ports.find((port) => port.source_port_id === "b")
```
imrishabh18, line 122: Why the `source_port_id` just letter? It shouldn't be the case it should follow the standard format!

imrishabh18 (changes requested):

seveibar (approved):
Can you move your test to a separate file to avoid overriding snapshots? Generally we want one test per file, you can always add “2” to the filename and “1” to the original if you want to reuse the name

ShiboSoftwareDev (approved):

imrishabh18 (approved):

PR #628, opened 2025-02-15
implement switch component

### LINKED ISSUE #585: implement switch component
@seveibar AI thought there should be a switch component, what do you think?
Image

techmannih:
fixes #585 
/claim #585


lib/components/normal-components/Switch.ts
```diff
@@ -0,0 +1,49 @@
...
+    if (spst) return "spst"
+    if (spdt) return "spdt"
+    if (dpst) return "dpst"
+    return type ?? "spst"
+  }
+
+  get config() {
+    const switchType = this._getSwitchType()
+
+    const baseSymbolName: Record<string, string> = {
```
seveibar, line 19: bad var name, also unnecessary type def

```suggestion
    const baseSymbolNameMap = {
```

seveibar (approved):
see comment, otherwise gtg

seveibar (approved):

PR #629, opened 2025-02-15
add maxVoltageRating to capacitor

### LINKED ISSUE #622: Add optional voltage prop for the capacitor component
we need to support voltage for capacitors.

Image

Anshgrover23:
Fixes #622


Abse2001 (approved):
great job!!!

seveibar (commented):
It’s a bit odd to display the max voltage on the schematic imo

sometimes its done but its actually quite confusing when mixed with simulation.

i think its sometimes important, we could introduce schShowRatings to enable it to be displayed but it should default to off i think

imrishabh18 (commented):
I think what @seveibar meant with the `schShowRatings` prop was, that it should be a parent level prop. Like in `board` or `group`

Anshgrover23:
@imrishabh18 no, i dont know about that because he merged my this prop pr https://github.com/tscircuit/props/pull/182 so is this also wrong. or only in core we need to change.

lib/components/normal-components/Capacitor.ts
```diff
@@ -73,9 +80,12 @@ export class Capacitor extends NormalComponent<
       manufacturer_part_number: props.manufacturerPartNumber ?? props.mfn,
       supplier_part_numbers: props.supplierPartNumbers,
       capacitance: props.capacitance,
+      max_voltage_rating: props.maxVoltageRating,
+      sch_show_ratings: props.schShowRatings,
       max_decoupling_trace_length: props.maxDecouplingTraceLength,
       display_capacitance: this._getSchematicSymbolDisplayValue(),
-    } as SourceSimpleCapacitorInput)
+    } as any)
```
seveibar, line 87: bad type

seveibar (changes requested):
see comment

seveibar:
> I think what @seveibar meant with the `schShowRatings` prop was, that it should be a parent level prop. Like in `board` or `group`

@imrishabh18 it's not a bad idea for a top level prop but it's fine here too. We will probably have a "showRatings" mode at a subcircuit level at some point because it also makes sense there

seveibar (approved):
nice work /tip $3

PR #630, opened 2025-02-15
footprinter update

imrishabh18:


seveibar (approved):
Lol ok the padding is a bit weird i guess. Also while we’re tweaking this stuff we should get the silkscreen text working! These are all supposed to have labels like R1

PR #634, opened 2025-02-17
feat: board auto-size when dimensions are not provided

imrishabh18:
`PcbBoardAutoSize` phase is run after the calculation of the `PcbComponentSizeCalculation` so that we can get the actual dimensions of the child


seveibar (approved):
this is a very exciting feature

PR #642, opened 2025-02-22
feat: internallyConnectedPorts prop

### LINKED ISSUE #518: `internallyConnectedPorts` prop
Indicates that ports are internally connected inside of a chip

Solves the "push button problem" because push buttons have internally connected ports, but schematic representation only has two ports

/bounty $25

Anshgrover23:
/claim #518
Fixes #518 

Screenshot from 2025-02-23 01-18-12


seveibar (changes requested):
there is no trace if something is internally connected

seveibar (changes requested):
?

PR #649, opened 2025-02-24
insert a pcb_trace_error when trace out of board

### LINKED ISSUE #540: When the component is placed outside the board and connected, It should throw error
https://snippets.tscircuit.com/editor?&snippet_type=board#data:application/gzip;base64,H4sIAPuPj2cAA42QvQ6CMBRGd57iphMMQlFH6ELiAzC5llJso/1JqdHE+O422AbD5Hi+nJuTXP60xnkY+UTvNw95AS2BPANoBkPdCA85etGiWikEgsuL8AFwIBKcYDk+y9kbtxDAF6lmPGhXFNfJGG+d1OEWH/E+zZqqoPV14pmJc/s6vCNaNqxYxR6jljK5BiPHIsbYnv6qdpvqbpNNnLreUcZhcka1qOxrIFBaqWsE3oSh+xmWi6Za3key4gPg67TLYQEAAA==

Image

When the components are placed outside the board and are connected, then the snippet should throw error and not draw the traces.

Abse2001:
/claim #540
/closes #540


lib/components/primitive-components/Trace/outside-board.ts
imrishabh18, line 1: This could have been in `utils` folder

imrishabh18 (changes requested):

seveibar:
> I think we should not create the trace when the components are outside the board, cause it doesn't make sense.
> 
> cc: @seveibar

Hehe this is a good point, we also want to be able to autoresize the board for the traces if the traces are available

i think in the short term its ok to create the traces with an error but yea long term we should probably not create them. I just think it might help users debug if theyre drawn. There is a flag iirc called is_speculative or somrthing on traces, we could have a flag like is_outside_board that could indicate it cant be physically placed

lib/utils/is-route-outside-board.ts
```diff
@@ -0,0 +1,25 @@
...
+
+  const outsideBoard = mergedRoute.some((point) => {
+    return (
+      point.x < boardCenterX - boardWidth / 2 ||
+      point.y < boardCenterY - boardHeight / 2 ||
+      point.x > boardCenterX + boardWidth / 2 ||
+      point.y > boardCenterY + boardHeight / 2
+    )
+  })
+  return outsideBoard
```
seveibar, line 24: this doesn't work for board outlines

seveibar: you can handle board outlines in a future issue but you need to at least handle the case by returning true if you detect an outline

seveibar (approved):
see comments

imrishabh18 (approved):

PR #655, opened 2025-02-27
Improve message Error: Cannot read properties of undefined reading getPortSelector

### LINKED ISSUE #345: Improve message Error:  Cannot read properties of undefined reading getPortSelector
Reproduction:

```tsx
import { useUsbC } from "@tsci/seveibar.smd-usb-c"
import { useRP2040 } from "@tsci/seveibar.rp2040"

export const RP2040Module = (props: { name: string }) => {
  const RP2040 = useRP2040("U1")
  const USBC = useUsbC("J1")
  
  return (
    <group>
      <RP2040 pcbY={8} />
      <USBC schX={-6} pcbY={-8} />

      <trace from={USBC.GND} to="net.GND" />
      <trace from={RP2040.GND} to="net.GND" />
      
    </group>
  )
}
```

### Error
```
Render Error: Cannot read properties of undefined (reading 'getPortSelector')
```

Make sure to create a reproduction, but don't import snippets for the reproduction, just use a resistor or something.

techmannih:
fixes #345 
/claim #345


seveibar (changes requested):
needs test

PR #664, opened 2025-03-01
feat: add circuit change event to RootCircuit

### LINKED ISSUE #660: Add support for `change` event on RootCircuit that updates whenever circuit json changes
We should emit this on any detected changes after a render cycle (we may need a heuristic for a change like `db.toArray().length`)

kom-senapati:
/claim #660
Closes #660


seveibar (changes requested):
extremely computationally expensive

PR #665, opened 2025-03-01
feat(sel): implement function pattern for component port selection

### LINKED ISSUE #663: Implement `sel` "function pattern"
Instead of having component hooks, which are misused a lot, we should enable the `sel` object to act as a function and allow selecting ports on a component via inspection

Consider the following component:

```tsx
import { createUseComponent, type CommonLayoutProps } from "tscircuit"

interface Props extends CommonLayoutProps {
  name: string
}


const pinLabels = {
  pin1: "VCC",
  pin2: "DISCH",
  pin3: "THRES",
  pin4: "CTRL",
  pin5: "GND",
  pin6: "TRIG",
  pin7: "OUT",
  pin8: "RESET"
} as const

export const A555Timer = (props: Props) => (
  <chip
    footprint="soic8"
    pinLabels={pinLabels}
    {...props}
  />
)
```

You should be able to do this with `sel` with good types:

```tsx
import { sel } from "@tscircuit/core"
import { A555Timer } from "@tsci/seveibar.A555Timer"

export default () => (
  <board>
    <A555Timer name="U1" />
    <trace from="net.GND" to={sel(A555Timer).U1.VCC} />
  </board>
)
```

/bounty $35

kom-senapati:
/claim #663
Closes #663

### DIFF
--- a/lib/sel/sel-utility-types.ts
+++ b/lib/sel/sel-utility-types.ts
@@ -193,19 +193,17 @@ export type CommonPinNames =
 export type TransistorPinNames = "base" | "collector" | "emitter"
 
 // New types for function pattern
+export type ComponentWithPinLabels = { pinLabels: Record<string, string> }
+
+export type ComponentInstance<T extends ComponentWithPinLabels> = {
+  [K in `U${Nums40}`]: Record<keyof T["pinLabels"], string>
+}
+
 export type ComponentType<T> = {
   new (...args: any[]): T
   pinLabels: Record<string, string>
 }
 
-export type ExtractPinLabels<T> = T extends { pinLabels: infer P }
-  ? keyof P
-  : never
-
-export type ComponentInstance<T extends { pinLabels: Record<string, string> }> =
-  {
-    [K in `U${Nums40}`]: Record<keyof T["pinLabels"], string>
-  }
-
-export type ComponentSelector<T extends { pinLabels: Record<string, string> }> =
-  (component: ComponentType<T>) => ComponentInstance<T>
+export type ComponentSelector<T extends ComponentWithPinLabels> = (
+  component: ComponentType<T>,
+) => ComponentInstance<T>
--- a/lib/sel/sel-utility-types.ts
+++ b/lib/sel/sel-utility-types.ts
@@ -177,12 +177,6 @@ export type CommonPinNames =
   | "VIN"
   | "VOUT"
   | "VREF"
-  | "VIN"
-  | "VOUT"
-  | "VREF"
-  | "VIN"
-  | "VOUT"
-  | "VREF"
   | "OUT"
   | "DISCH"
   | "EN"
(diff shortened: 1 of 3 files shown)


lib/sel/sel-utility-types.ts
```diff
@@ -177,12 +177,6 @@ export type CommonPinNames =
   | "VIN"
   | "VOUT"
   | "VREF"
-  | "VIN"
-  | "VOUT"
-  | "VREF"
-  | "VIN"
-  | "VOUT"
-  | "VREF"
```
seveibar, line 185: ? why remove?

lib/sel/sel.ts
```diff
@@ -76,10 +77,38 @@ type SelWithoutSubcircuit = NonPolarizedSel &
...
+export const sel: Sel & ComponentSelector<any> = new Proxy(
+  function (component?: any) {
+    if (component) {
+      // Function pattern
+      return new Proxy(
+        {},
+        {
+          get: (_, prop1: string) => {
+            // Only allow U1-U40
+            if (!prop1.match(/^U([1-9]|[1-3][0-9]|40)$/)) return undefined
```
seveibar, line 89: no need for this check

lib/sel/sel-utility-types.ts
kom-senapati: duplicate

tests/sel/sel2-function-pattern.test.ts
```diff
@@ -0,0 +1,45 @@
...
+
+test("sel net pattern still works", () => {
+  const result = sel.net.GND
+  expect(result).toBe("net.GND")
+})
+
+test("sel subcircuit pattern still works", () => {
+  const result = sel.subcircuit.S1.U1.pin1
+  expect(result).toBe("subcircuit.S1 > .U1 > .pin1")
+})
```
seveibar, line 36: remove any tests from this file that don't test the function pattern

tests/sel/sel2-function-pattern.test.ts
```diff
@@ -0,0 +1,45 @@
...
+    CTRL: "CTRL",
+    GND: "GND",
+    TRIG: "TRIG",
+    OUT: "OUT",
+    RESET: "RESET",
+  } as const
+}
+
+test("sel function pattern works with component types", () => {
+  const result = sel(A555Timer).U1.VCC
```
seveibar, line 19: can you add a test that uses // @ts-expect-error that makes sure you can't access a pin like `sel(A555Timer).U1.DOES_NOT_EXIST`

seveibar (commented):
really good work, see comments

lib/sel/sel-utility-types.ts
seveibar: 👍

seveibar (commented):
one other thing: test when the `pinLabels` can't be properly inferred, in these cases the code should still work but the type autocompletion may not work

important: `sel(A555Timer).U1.*` should have proper completion with any of the component's pins when the pinLabels are properly defined. If this doesn't work the feature isn't useful! Good work 👍

tests/sel/sel2-function-pattern.test.ts
kom-senapati: added

lib/sel/sel.ts
```diff
@@ -76,10 +78,50 @@ type SelWithoutSubcircuit = NonPolarizedSel &
 
 export type Sel = SubcircuitSel & SelWithoutSubcircuit
 
-export const sel: Sel = new Proxy(
-  {},
+type ComponentProxy = <T extends ComponentWithPinLabels>(
```
seveibar, line 81:
```suggestion
type FunctionSelComponentProxy = <T extends ComponentWithPinLabels>(
```

tests/sel/sel2-function-pattern.test.ts
```diff
@@ -0,0 +1,26 @@
+import { sel } from "../../lib/sel"
+import { expect, test } from "bun:test"
+
+// Mock component class
+class A555Timer {
```
seveibar, line 5: this.... isn't a component

seveibar (changes requested):
you can't change the way A555Timer is defined

PR #687, opened 2025-03-04
Dispatch workflow to @tscircuit/eval to update core after a release

seveibar:
- **trigger dispatch on successful release**
- **reduce paths that trigger release**


seveibar:
CC @imrishabh18 this will make it a bit easier to keep @tscircuit/eval up to date. @tscircuit/eval has to build the webworker (although we could have @tscircuit/eval dynamically import core which is even easier in some ways)

PR #690, opened 2025-03-05
implement function pattern for component port selection

### LINKED ISSUE #663: Implement `sel` "function pattern"

techmannih:
fixes #663 
/claim #663

### DIFF
--- a/lib/sel/sel-utility-types.ts
+++ b/lib/sel/sel-utility-types.ts
@@ -177,12 +177,6 @@ export type CommonPinNames =
   | "VIN"
   | "VOUT"
   | "VREF"
-  | "VIN"
-  | "VOUT"
-  | "VREF"
-  | "VIN"
-  | "VOUT"
-  | "VREF"
   | "OUT"
   | "DISCH"
   | "EN"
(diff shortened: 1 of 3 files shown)


seveibar (changes requested):
/sabotage, the implementation has “U1” to get it to pass in the test

PR #691, opened 2025-03-06
fix: hover bug for traces

imrishabh18:


seveibar (approved):
yep that's the fix for sure! nice!!!!

PR #701, opened 2025-03-13
add import path in tests

Anshgrover23:


seveibar (approved):
fwiw iirc you can specify the name as the second param to toMatchSchematicSnapshot

PR #703, opened 2025-03-17
fix: auto-size board when it has group as child

imrishabh18:
After the ComponentSizeRender phase, calculating the board dimensions in the BoardAutoSize phase.


tests/components/normal-components/board-auto-size.test.tsx
```diff
@@ -79,3 +79,46 @@ test("board centers around components", () => {
...
+        />
+      </group>
+    </board>,
+  )
+  circuit.render()
+  const pcb_board = circuit.db.pcb_board.list()[0]
+  expect(pcb_board.center.x).toBe(0)
+  expect(pcb_board.center.y).toBe(0)
+  expect(pcb_board.width).toBeGreaterThan(10)
+  expect(pcb_board.height).toBeGreaterThan(10)
```
seveibar, line 123: good one for a visual test as well!

seveibar (approved):

PR #704, opened 2025-03-18
Don't add vias as pcb traces if returned from server in output_pcb_traces

seveibar:


lib/components/primitive-components/Group/Group.ts
```diff
@@ -582,18 +584,28 @@ export class Group<Props extends z.ZodType<any, any, any> = typeof groupProps>
...
-            y: point.y,
-            hole_diameter: 0.3,
-            outer_diameter: 0.6,
-            layers: [point.from_layer as LayerRef, point.to_layer as LayerRef],
-            from_layer: point.from_layer as LayerRef,
-            to_layer: point.to_layer as LayerRef,
-          })
+      if (pcb_trace.type === "pcb_via") {
+        // TODO handling here- may need to handle if redundant with pcb_trace
+        // below (i.e. don't insert via if one already exists at that location)
```
seveibar, line 589: CC @imrishabh18 , the output from the server can sometimes have vias that we need to handle, as indicated in the comment we may need some special handling to avoid duplicates. Ideally, we avoid vias defined like this- it's a better better if they're defined as part of the pcb_trace route (that way you could e.g. remove the route with all vias included)

image

imrishabh18: Will look more onto this issue, cause now I see the via's are getting skipped.

image

PR #712, opened 2025-03-22
feat: introduce stampboard

### LINKED ISSUE #213: Introduce `<stampboard />`

Anshgrover23:
/claim #213
Fixes #213


lib/components/normal-components/Stampboard.ts
```diff
@@ -0,0 +1,24 @@
...
+      shouldRenderAsSchematicBox: true,
+    }
+  }
+
+  doInitialSourceRender() {
+    if (!this.root) return
+    const { db } = this.root
+
+    const source_component = db.source_component.insert({
+      ftype: "simple_chip",
```
imrishabh18, line 18: We don't have a `simple_stampboard` for this in circuit-json?

imrishabh18 (commented):
Looks okay to me, see the comments. Also try to add a schematic svg for this

lib/components/normal-components/Stampboard.ts
Anshgrover23: no, do we need that. @imrishabh18

imrishabh18: Yes, we need to follow the convention!

PR #715, opened 2025-03-23
fix: improve port selector error message to show available ports and component name

### LINKED ISSUE #345: Improve message Error:  Cannot read properties of undefined reading getPortSelector

Anshgrover23:
/claim #345
Fixes #345
Screenshot from 2025-03-23 19-33-53

### DIFF
--- a/lib/components/primitive-components/Trace/Trace.ts
+++ b/lib/components/primitive-components/Trace/Trace.ts
@@ -147,15 +147,24 @@ export class Trace
         const parentSelector = selector.replace(/\>.*$/, "")
         const targetComponent = this.getSubcircuit().selectOne(parentSelector)
         if (!targetComponent) {
-          this.renderError(`Could not find port for selector "${selector}"`)
-        } else {
           this.renderError(
-            `Could not find port for selector "${selector}" (did you forget to include the pin name?)\nsearched component ${targetComponent.getString()}, which has ports: ${targetComponent.children
-              .filter((c) => c.componentName === "Port")
-              .map(
-                (c) => `${c.getString()}(${c.getNameAndAliases().join(",")})`,
-              )
-              .join(" & ")}`,
+            `Could not find component for selector "${parentSelector}"`,
+          )
+        } else {
+          const availablePorts = targetComponent.children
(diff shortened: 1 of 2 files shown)


lib/components/primitive-components/Trace/Trace.ts
```diff
@@ -147,15 +147,24 @@ export class Trace
...
+          const portName = selector.split(">").pop()?.trim()
           this.renderError(
-            `Could not find port for selector "${selector}" (did you forget to include the pin name?)\nsearched component ${targetComponent.getString()}, which has ports: ${targetComponent.children
-              .filter((c) => c.componentName === "Port")
-              .map(
-                (c) => `${c.getString()}(${c.getNameAndAliases().join(",")})`,
-              )
-              .join(" & ")}`,
+            `Port "${portName}" not found on component "${targetComponent.props.name}". ` +
+              `Available ports are: ${availablePorts.join("; ")}`,
```
imrishabh18, line 167: I don't think it makes sense to show the availablePorts. Think what if the component has 30 or more ports. The error message would be unreadable in the view.

imrishabh18 (changes requested):

lib/components/primitive-components/Trace/Trace.ts
Anshgrover23: @imrishabh18 done changes.

Anshgrover23:
resolved.

Issue #717, opened 2025-03-27
Get DRC Checks to work with async autorouters


seveibar:
The DRC errors currently aren't added if the pcb traces don't compute in the first render

to do this, you'll need to handle traces being changed, the functionality in `Group.ts` will need to be modified. `doInitialPcbDesignRuleChecks` has a lot of the functionality that will need to be moved to `updatePcbDesignRuleChecks`

/bounty $15

PR #720, opened 2025-03-27
fix: Add DRC checks support for async autorouters

onyedikachi-david:
Fixes: #717 
/claim #717


tests/features/async-drc-error-detection.test.tsx
```diff
@@ -0,0 +1,135 @@
...
+      <resistor name="R4" footprint="0402" resistance="10k" pcbX={0} pcbY={3} />
+
+      {/* Connect resistors with traces that will cross in the center */}
+      <trace from=".R1 > .pin1" to=".R2 > .pin2" />
+      <trace from=".R3 > .pin1" to=".R4 > .pin2" />
+    </board>,
+  )
+
+  // Ensure the circuit is fully rendered and async autorouting is complete
+  await circuit.renderUntilSettled()
```
Anshgrover23, line 115: do not use renderuntilsettled method use only `render`

Anshgrover23 (changes requested):

onyedikachi-david:
I do not understand why the snapshot is failing on the CI but passes locally @Anshgrover23

Anshgrover23:
@onyedikachi-david have u ran `bun run snapshot` to update the new snapshots?

onyedikachi-david:
> bun run snapshot

I ran this `➜ bun test --update-snapshots tests/features/async-drc-error-detection.test.tsx` @Anshgrover23

Anshgrover23:
@onyedikachi-david try to delete the original snapshot and then generate again see if this resolves.

onyedikachi-david:
> @onyedikachi-david try to delete the original snapshot and then generate again see if this resolves.

Yeah, I did that also during the last commit.

MustafaMulla29 (changes requested):
image

you can here that its cause of diff snapshots. Check if there are diff png's in your snapshots folder

seveibar:
I don't think this PR does anything meaningfully related to the issue

lib/components/normal-components/Board.ts
```diff
@@ -146,11 +146,31 @@ export class Board extends Group<typeof boardProps> {
   }
 
   doInitialPcbDesignRuleChecks() {
+    if (this.root?.pcbDisabled) return
+    if (this.getInheritedProperty("routingDisabled")) return
+    this.updatePcbDesignRuleChecks()
```
seveibar, line 151: don't call lifecycle methods directly, you may need a hidden function e.g. `_doDrcChecks()`

lib/components/normal-components/Board.ts
```diff
@@ -146,11 +146,31 @@ export class Board extends Group<typeof boardProps> {
...
-    for (const error of errors) {
+
+    // Store existing non-DRC errors
+    const existingErrors = db.pcb_trace_error
+      .list()
+      .filter((error) => !error.message.includes("overlaps with"))
+
+    // Clear all trace errors
+    db.pcb_trace_error.list().forEach((error) => {
+      db.pcb_trace_error.delete(error.pcb_trace_error_id)
```
seveibar, line 166: don't delete all pcb trace errors, only delete ones that are created from this board's DRC. You may need to store a property to indicate it was created by this Board

PR #721, opened 2025-03-29
add connection prop to diode

### LINKED ISSUE #719: `<diode />` doesn't accept `connections` prop
/bounty $7

Anshgrover23:
/claim #719
Fixes #719


seveibar (changes requested):
The default values for pcb component width/hright are weird, im not sure you need to be redefining those functions- at the very least you should make sure that footprints work with a snapshot test if making changes to that fn

Anshgrover23:
@seveibar i think you are right no need of that pcb render function for this issue. i have removed that.

lib/components/normal-components/Diode.ts
```diff
@@ -34,4 +34,26 @@ export class Diode extends NormalComponent<
   anode = this.portMap.pin1
   neg = this.portMap.pin2
   cathode = this.portMap.pin2
+
+  doInitialSchematicComponentRender(): void {
+    const { _parsedProps: props } = this
+
+    super.doInitialSchematicComponentRender()
+  }
```
seveibar, line 42: why extend this method?

lib/components/normal-components/Diode.ts
```diff
@@ -34,4 +34,26 @@ export class Diode extends NormalComponent<
...
+      ftype: "simple_diode",
+      name: props.name,
+    })
+
+    this.source_component_id = source_component.source_component_id!
+  }
+
+  doInitialCreateTracesFromProps(): void {
+    this._createTracesFromConnectionsProp()
+  }
```
seveibar, line 58: why extend this method?

seveibar (changes requested):
@Anshgrover23 you've got to read over your code before requesting review, the `super` calls are clearly not needed!

Anshgrover23:
@seveibar i think i misinterpret the functions there is only needed change in props after doing that this all are working fine.

seveibar (approved):
yep, was just the prop definition that was missing

PR #724, opened 2025-04-07
fix: same name of the components should add an error to the circuit json

### LINKED ISSUE #723: This circuit doesn't throw any error, why?
When both the components are having the same name and a trace is being added. This should error out, with something like components having same name, right?

Image

https://tscircuit.com/editor?&snippet_type=board#data:application/gzip;base64,H4sIANpG8WcAA7WQvQ6CMBSFd57iphMMAlVH2odgci2l2Eb7k3KNJsZ3t0EaE3bG7+Q7OclRr+Ajwqgm8bgjlBUwDmUB0A1exBGeZkTNCG2tJaCVuWpciScpaVHNZkYfFwL4oXBSJe1G1nTyHkM0LnXbc3vMsRM2aT3NPEt9Ye/TZ8Ughz82u+0dNoOZ8yJGIRVM0VtG6p4ChzoYRwmg3wRLo2uW53hRfQHYgQfKXAEAAA==

@seveibar

imrishabh18:
/close #723

### DIFF
--- a/lib/components/normal-components/Board.ts
+++ b/lib/components/normal-components/Board.ts
@@ -144,4 +144,17 @@ export class Board extends Group<typeof boardProps> {
   _computePcbGlobalTransformBeforeLayout(): Matrix {
     return identity()
   }
+
+  doInitialPcbDesignRuleChecks() {
+    if (this.root?.pcbDisabled) return
+    if (this.getInheritedProperty("routingDisabled")) return
+    const { db } = this.root!
+
+    super.doInitialPcbDesignRuleChecks()
+
+    const errors = checkEachPcbTraceNonOverlapping(db.toArray())
+    for (const error of errors) {
+      db.pcb_trace_error.insert(error)
+    }
+  }
 }
--- a/lib/components/normal-components/Board.ts
+++ b/lib/components/normal-components/Board.ts
@@ -144,14 +144,4 @@ export class Board extends Group<typeof boardProps> {
   _computePcbGlobalTransformBeforeLayout(): Matrix {
     return identity()
   }
-
-  doInitialPcbDesignRuleChecks() {
-    if (this.root?.pcbDisabled) return
-    if (this.getInheritedProperty("routingDisabled")) return
-    const { db } = this.root!
-    const errors = checkEachPcbTraceNonOverlapping(db.toArray())
-    for (const error of errors) {
-      db.pcb_trace_error.insert(error)
-    }
-  }
 }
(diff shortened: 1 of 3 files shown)


lib/components/normal-components/Board.ts
```diff
@@ -144,14 +144,4 @@ export class Board extends Group<typeof boardProps> {
...
   }
-
-  doInitialPcbDesignRuleChecks() {
-    if (this.root?.pcbDisabled) return
-    if (this.getInheritedProperty("routingDisabled")) return
-    const { db } = this.root!
-    const errors = checkEachPcbTraceNonOverlapping(db.toArray())
-    for (const error of errors) {
-      db.pcb_trace_error.insert(error)
-    }
```
seveibar, line 155: I think we want to only run this on the board, because DRC requirements are board-wide, it is a bit of an unusual case but i do think ots correct to have it on board

lib/components/primitive-components/Group/Group.ts
```diff
@@ -714,4 +718,47 @@ export class Group<Props extends z.ZodType<any, any, any> = typeof groupProps>
...
     return autorouter.groupMode === "sequential-trace"
   }
+
+  doInitialPcbDesignRuleChecks() {
+    if (this.root?.pcbDisabled) return
+    if (this.getInheritedProperty("routingDisabled")) return
+    const { db } = this.root!
+
+    if (this.isSubcircuit) {
+      const subcircuitComponentsByName = new Map<string, PrimitiveComponent[]>()
```
seveibar, line 728: Subcircuit component name checking is good 👍 this one doesnt need to go on the board

tests/repros/repro12-same-name-of-components.test.tsx
```diff
@@ -0,0 +1,96 @@
...
+        />
+        <resistor
+          resistance="1k"
+          footprint="0402"
+          name="R2"
+          schX={-3}
+          pcbX={-3}
+        />
+      </subcircuit>
+    </group>,
```
seveibar, line 88: This is a valid circuit that shouldnt error. Names must be unique within a subcircuit

lib/components/primitive-components/Group/Group.ts
imrishabh18: So, we don't need to check it here in Board. But we will have to import from the Group class for this behaviour to be accepted, right?

seveibar: You can call the super method inside the board so that it runs the DRC checks that run for any subcircuit. You can alternatively create an internal underscore function that does what the subcircuit does. For example _checkForConflictingNames()

tests/repros/repro12-same-name-of-components.test.tsx
seveibar: Woops sorry ty!!

seveibar (approved):

PR #726, opened 2025-04-10
applySchematicEditEventsToManualEditsFile method added

imrishabh18:


seveibar (approved):
don't you need to export the method as well?

PR #727, opened 2025-04-10
Bump props to 0.0.167

andrii-balitskyi:


seveibar (approved):
i think a lot of packages on tscircuit.com may break, i'll do an announcement on the discord (the SmdUsbC is widely used for example)

Issue #725, opened 2025-04-10
Bug: `pnp` transistor `collector` pin is numbered incorrectly


imrishabh18:
<img width="681" alt="Image" src="https://github.com/user-attachments/assets/0d3ca489-a148-42c8-a950-fcab8da61a49" />

/bounty $2

PR #729, opened 2025-04-10
Make pnp check

Slaviiiii:
/claim #725


imrishabh18 (commented):
Can you please add the test case with the circuit which was in the issue to make sure we have resolved it?

Slaviiiii:
@imrishabh18 just wrote the test

tests/components/normal-components/transistor-mapping.test.tsx
imrishabh18, line 1: This test doesn't make sense! Use the example in the issue, where you draw the trace using the alias in the `to/from` props. Also, add a schematic snapshot for verification

imrishabh18 (changes requested):

tests/components/normal-components/transistor-mapping.test.tsx
Slaviiiii: Not sure how to make a more specific test for it, but I am pretty sure it covers the need, no?

PR #734, opened 2025-04-13
fix: transistor collector pin numbering

### LINKED ISSUE #725: Bug: `pnp` transistor `collector` pin is numbered incorrectly

Anshgrover23:
/claim #725
Fixes #725


lib/components/normal-components/Transistor.ts
```diff
@@ -30,15 +30,15 @@ export class Transistor extends NormalComponent<
       pinCount: 3,
       additionalAliases: {
         pin1: ["emitter", "e"],
-        pin2: ["collector", "c"],
-        pin3: ["base", "b"],
+        pin2: ["base", "b"],
+        pin3: ["collector", "c"],
```
imrishabh18, line 34: The pin numbers differ for `npn` and `pnp` so you need to make a check for that

imrishabh18 (changes requested):

lib/components/normal-components/Transistor.ts
Anshgrover23: @imrishabh18 done.

tests/components/normal-components/__snapshots__/transistor-aliases-schematic.snap.svg
imrishabh18, line 1: image

I don't think so it's right

imrishabh18 (changes requested):

PR #736, opened 2025-04-14
porting the `applyEditEvents` method from pcb-viewer to core

imrishabh18:


lib/utils/edit-events/apply-trace-hint-edit-event.ts
```diff
@@ -0,0 +1,46 @@
+import type { AnyCircuitElement } from "circuit-json"
+import { su } from "@tscircuit/soup-util"
+import type { EditTraceHintEvent } from "@tscircuit/props"
+
+export const applyTraceHintEditEvent = (
+  soup: AnyCircuitElement[],
```
seveibar, line 6:
```suggestion
  circuitJson: AnyCircuitElement[],
```

seveibar (approved):
we ideally should have some tests for this, can create an issue from this comment

Issue #740, opened 2025-04-16
Schematic Trace crossing "hop" generated incorrectly in some cases


seveibar:
You MUST reproduce prior to fixing in a snapshot test to claim bounty

![Image](https://github.com/user-attachments/assets/2b4d01a0-f583-474a-a5eb-2cb7680297fe)

/bounty $20

PR #753, opened 2025-04-20
repro Schematic Trace crossing "hop" generated incorrectly

Abse2001:
related to https://github.com/tscircuit/core/issues/740


Abse2001:
@seveibar is there a way to use the snippet without copying the whole code.

seveibar:
@Abse2001 yes but you have to skip the test because we never run tests with external network calls on ci

Its in the eval module, something called runTscircuitModule or something

ShiboSoftwareDev:
@seveibar can't you use the importSnippet function?

seveibar:
@ShiboSoftwareDev @Abse2001 apologies, the function i was referring to doesn't exist, i lost it in a branch i forgot to PR, I am PR'ing in shortly

Abse2001:
I get it now!

seveibar (approved):

PR #758, opened 2025-04-24
reproduction for manualEdits not causing warning

seveibar:
@MustafaMulla29 i reproduced the bug by copying code from tscircuit.com


ArnavK-09:
instead #760

ArnavK-09:
This test is added in #760

PR #762, opened 2025-04-26
fixed trace going into symbol

### LINKED ISSUE #578: Fix schematic traces routing through symbols
Image

/bounty $20

ShiboSoftwareDev:
/claim #578
/closes #578


tests/repros/__snapshots__/repro13-schematic-trace-hop-schematic.snap.svg
imrishabh18, line 1: Something is wrong, multiple traces between the same ports and even the break point on turn of the is wrong

imrishabh18 (changes requested):

seveibar:
This one looks a lot worse, not sure why: tests/examples/__snapshots__/bug-high-port-number-schematic.snap.svg

seveibar:
I would also try to isolate the math out of Trace.ts if possible, but this looks like a really good attempt, I'm hopeful!

PR #764, opened 2025-04-28
add the jlcPartsEngine as default

imrishabh18:


seveibar (changes requested):
We need to be careful here, every circuit thats rendered will now make network calls with the parts engine. I also think we may want to avoid network calls by default to avoid inconsistent behavior for downstream testing.

As a compromise, maybe we have eval inject this parts engine rather than core.

seveibar:
i think #

PR #767, opened 2025-05-01
fixed trace going into symbol 

ShiboSoftwareDev:
now refactored the logic out of Trace.ts

the double trace fix helped a lot

/claim https://github.com/tscircuit/core/issues/578
/closes https://github.com/tscircuit/core/issues/578

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@tscircuit/core",
   "type": "module",
-  "version": "0.0.384",
+  "version": "0.0.385",
   "types": "dist/index.d.ts",
   "main": "dist/index.js",
   "module": "dist/index.js",


lib/components/primitive-components/Trace/Trace.ts
```diff
@@ -862,7 +861,7 @@ export class Trace
 
     const portIds = connectedPorts.map((p) => p.port.schematic_port_id).sort()
     const portPairKey = portIds.join(",")
-    const board = this.parent as Board
+    const board = this.root?._getBoard()
```
seveibar, line 864: this should be operating on the subcircuit, not the board, otherwise subcircuits aren't isolated (subcircuits have a major optimization strategy where they render independently from the board)

seveibar (commented):
generally this is much much better with only one small regression that i can see!

PR #769, opened 2025-05-01
introduced a smoke test to verify dist after build

ShiboSoftwareDev:


.github/workflows/bun-typecheck.yml
```diff
@@ -22,5 +22,8 @@ jobs:
       - name: Install dependencies
         run: bun i
 
+      - name: Build Project
+        run: bun run build
```
seveibar, line 26: ahh i merged too soon, you should add `// @ts-ignore` to the script file, we never want to have a build required to typecheck projects, it causes other issues

PR #774, opened 2025-05-02
fixed trace going into symbol

ShiboSoftwareDev:


seveibar (commented):
I think something is wrong because of the issue in repro13, i generally understand the concept (i think) but why can one resistor’s trace travel out at a 90 degree angle but not the other one? I may look at this myself later, im not totally comfortable maintaining this logic atm

PR #776, opened 2025-05-03
fixed trace going into symbol

ShiboSoftwareDev:
I was unsatisfied with my previous fix after you said it was hard to maintain, it's true my fix wasn't good since I was patching on top of the issue without addressing it, so I experimented more and found out that the issue was that the autorouter doesn't know what to do when the trace has to go to the opposite side of the target with nothing to block it, so I created an obstacle which is a border around the schematic.


lib/components/primitive-components/Trace/get-obstacles-for-trace.ts
```diff
@@ -23,7 +23,49 @@ export const getSchematicObstaclesForTrace = (trace: Trace): Obstacle[] => {
...
+      typeof cx === "number" &&
+      typeof cy === "number" &&
+      typeof w === "number" &&
+      typeof h === "number"
+    ) {
+      minX = Math.min(minX, cx - w / 2)
+      maxX = Math.max(maxX, cx + w / 2)
+      minY = Math.min(minY, cy - h / 2)
+      maxY = Math.max(maxY, cy + h / 2)
+    }
```
seveibar, line 67: You should isolate this into a separate function eg getBoundsForSchematicElement, where possible and ergonomic we should avoid extending these already large class files

lib/components/primitive-components/Trace/get-obstacles-for-trace.ts
```diff
@@ -110,5 +152,51 @@ export const getSchematicObstaclesForTrace = (trace: Trace): Obstacle[] => {
...
+    // Right border (vertical)
+    obstacles.push({
+      type: "rect",
+      layers: ["top"],
+      center: { x: right, y: (top + bottom) / 2 },
+      width: thickness,
+      height: top - bottom,
+      connectedTo: [],
+    })
+  }
```
seveibar, line 199: Great comment, obstacle construction could also be isolated to a separate fn here, eg getObstaclesFromBounds(bounds,opts:{ padding })

seveibar (approved):
Minor refactor. This is a good contribution but im not sure its the actual fix (since it seems its still technically possible that the autorouter could draw through a pin)

that said, we can accept this and reopen the issue on prodbif we see an issue, i like this contribution 👍

seveibar (approved):

PR #779, opened 2025-05-06
Add `source_failed_to_create_component_error` to skip complete circuit render failure and add only the failed component to db

imrishabh18:
/close https://github.com/tscircuit/tscircuit.com/issues/990


lib/fiber/create-instance-from-react-element.ts
```diff
@@ -38,6 +40,22 @@ export function prepare<T extends Renderable>(
...
+        super({})
+        this._parsedProps = {}
+      }
+    })()
+
+  return errorComponent
+}
+
+const CircuitContext = {
+  current: null as RootCircuit | null,
```
seveibar, line 56:
```suggestion
  rootCircuit: null as RootCircuit | null,
```

lib/fiber/create-instance-from-react-element.ts
```diff
@@ -69,9 +87,23 @@ const hostConfig: HostConfig<
...
+      const rootCircuit = CircuitContext.current
 
-    return instance
+      if (rootCircuit?.db) {
+        // Add the error to the circuit-json database
+        rootCircuit.db.source_failed_to_create_component_error.insert({
+          type: "source_failed_to_create_component_error",
+          component_type: type,
+          message: error instanceof Error ? error.message : String(error),
+          error_type: "source_failed_to_create_component_error",
```
seveibar, line 102: nice, I think we could detect a zod/props error here and maybe make the error more specific e.g.

```suggestion
          error_type: "source_failed_to_create_component_error", // I can't remember why we have this, but it might be valid to use `invalid_props_for_component` as the error_type
          sub_error_type: "invalid_props_for_component",
          
          // 
          pcb_center: hasPosition ? { x: props.pcbX ?? parent.pcbX }: undefined, // Is it possible to get this? This could make it so we could display an error on the PCB /schematic so people don't wonder where their component is. It would be optional
          schematic_center: //...
```

lib/fiber/create-instance-from-react-element.ts
```diff
@@ -166,59 +198,70 @@ if (React.version.startsWith("19.")) {
...
-    reconciler.updateContainer(reactElm, container, null, () => {})
-  }
+  const previousCircuit = CircuitContext.current
+  CircuitContext.current = rootCircuit
 
-  // Don't throw here if you want to avoid synchronous errors
-  if (containerErrors.length > 0) {
-    throw containerErrors[0]
-  }
+  try {
```
seveibar, line 206: i don't think you need this giant try catch here

lib/fiber/create-instance-from-react-element.ts
```diff
@@ -69,9 +87,23 @@ const hostConfig: HostConfig<
...
+      if (rootCircuit?.db) {
+        // Add the error to the circuit-json database
+        rootCircuit.db.source_failed_to_create_component_error.insert({
+          type: "source_failed_to_create_component_error",
+          component_type: type,
+          message: error instanceof Error ? error.message : String(error),
+          error_type: "source_failed_to_create_component_error",
+        })
+      }
+      return createErrorPlaceholderComponent(type, props, error)
```
seveibar, line 105: this is an interesting idea, why not have the placeholder component render the error? We could have it be a Renderable with props, that would allow inspection. It would render the error to circuit json

seveibar (commented):
nice

imrishabh18:
Most of the tests are failing because of position of the silkscreen text, 🤔

image

lib/components/primitive-components/ErrorPlaceholder.ts
```diff
@@ -0,0 +1,48 @@
...
+    }
+  }
+
+  doInitialSourceRender() {
+    if (this.root?.db) {
+      this.root.db.source_failed_to_create_component_error.insert({
+        component_name: this._parsedProps.component_name,
+        message: this._parsedProps.message,
+        pcb_center: this._parsedProps.pcb_center,
+        schematic_center: this._parsedProps.schematic_center,
```
seveibar, line 40: it's ok to do this in the next PR, but you need to translate this by the parent group's position, you can look at schematic and pcb primitives to see how they do it

seveibar (approved):
@imrishabh18 this is amazing, INCREDIBLE

PR #781, opened 2025-05-06
fixing pnp transistor collector pin is numbered incorrectly

### LINKED ISSUE #725: Bug: `pnp` transistor `collector` pin is numbered incorrectly

Abse2001:
/claim #725
/closes #725


tests/components/normal-components/transistor-pin-mapping.test.tsx
imrishabh18, line 1: Create multiple tests to verify individually that the pin aliases are correct.

image

imrishabh18 (changes requested):

seveibar (commented):
this might not be the best test because i think it's hard to know from the snapshots if it's correct.

You could add `<schematictext />` to the snapshots to indicate it, or possible use `toMatchInlineSnapshot` with a nice curated string

PR #789, opened 2025-05-10
updated footprinter

ShiboSoftwareDev:


bun.lock
```diff
@@ -51,7 +51,7 @@
         "tsup": "^8.2.4",
       },
       "peerDependencies": {
-        "@tscircuit/footprinter": "*",
+        "@tscircuit/footprinter": "^0.0.155",
```
seveibar, line 54: Why the change here?

seveibar (approved):
The jumper boxes may want a sillscreen path or additional text (pin labels) in future

bun.lock
ShiboSoftwareDev: happened automatically, I didn't notice

PR #790, opened 2025-05-11
Jumper internallyConnectedPins are now actually connected

ShiboSoftwareDev:


lib/components/normal-components/Jumper.ts
```diff
@@ -69,4 +69,53 @@ export class Jumper<PinLabels extends string = never> extends NormalComponent<
...
+              segment.end_pcb_port_id &&
+              typeof segment.end_pcb_port_id === "string" &&
+              segment.end_pcb_port_id.startsWith("{PIN")
+            ) {
+              const pin = segment.end_pcb_port_id
+                .replace("{PIN", "")
+                .replace("}", "")
+              if (pinLabelToPortId[pin]) {
+                segment.end_pcb_port_id = pinLabelToPortId[pin]
+              }
```
Anshgrover23, line 114: @ShiboSoftwareDev can't we use switch case instead of this if-else loop hell?

Anshgrover23 (changes requested):

lib/components/normal-components/Jumper.ts
```diff
@@ -69,4 +69,53 @@ export class Jumper<PinLabels extends string = never> extends NormalComponent<
...
+
+  doInitialPcbTraceRender() {
+    const { db } = this.root!
+    const pcb_ports = db.pcb_port.list({
+      pcb_component_id: this.pcb_component_id,
+    })
+    const pinLabelToPortId: Record<string, string> = {}
+    // Map pin labels ("1", "2", etc.) to pcb_port_id
+    for (const port of pcb_ports) {
+      const match = port.source_port_id && port.source_port_id.match(/(\d+)$/)
```
seveibar, line 81:
```suggestion
      const match = port.source_port_id && port.source_port_id.match(/\{PIN(\d+)\}/)
```

seveibar: otherwise you're matching on stuff like `pcb_port_3` which i believe is not desired

lib/components/normal-components/Jumper.ts
```diff
@@ -69,4 +69,53 @@ export class Jumper<PinLabels extends string = never> extends NormalComponent<
...
+      if (match) {
+        const label = (parseInt(match[1], 10) + 1).toString()
+        pinLabelToPortId[label] = port.pcb_port_id
+      }
+    }
+    const traces = db.pcb_trace.list()
+    for (const trace of traces) {
+      if (trace.route) {
```
seveibar, line 89: switch to early bail
```suggestion
      if (!trace.route) continue
```

lib/components/normal-components/Jumper.ts
```diff
@@ -69,4 +69,53 @@ export class Jumper<PinLabels extends string = never> extends NormalComponent<
...
+        for (const segment of trace.route) {
+          if (segment.route_type === "wire") {
```
seveibar, line 91: switch to early bail

```suggestion
          if (segment.route_type !== "wire") continue
```

seveibar (approved):
you did it! minor changes requested but great work /tip $10

lib/components/normal-components/Jumper.ts
ShiboSoftwareDev: this is the source_port where are trying to place into the trace

PR #798, opened 2025-05-12
fixed solder jumper pins not mapping correctly

ShiboSoftwareDev:
+ modified the test to repro

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@tscircuit/core",
   "type": "module",
-  "version": "0.0.403",
+  "version": "0.0.404",
   "types": "dist/index.d.ts",
   "main": "dist/index.js",
   "module": "dist/index.js",


lib/components/normal-components/Jumper.ts
```diff
@@ -77,12 +77,10 @@ export class Jumper<PinLabels extends string = never> extends NormalComponent<
     })
     const pinLabelToPortId: Record<string, string> = {}
     // Map pin labels ("1", "2", etc.) to pcb_port_id
-    for (const port of pcb_ports) {
-      const match = port.source_port_id && port.source_port_id.match(/(\d+)$/)
-      if (match) {
-        const label = (parseInt(match[1], 10) + 1).toString()
-        pinLabelToPortId[label] = port.pcb_port_id
-      }
+    for (let i = 0; i < pcb_ports.length; i++) {
```
seveibar, line 80: this ordering is totally arbitrary now

seveibar (changes requested):
I must be missing something- this would break all jumper pin ordering- the order from the db isn't guaranteed

PR #799, opened 2025-05-12
Add support for parent group's transformation

imrishabh18:
Ref issue - https://github.com/tscircuit/core/pull/779#discussion_r2085031875


lib/components/primitive-components/ErrorPlaceholder.ts
```diff
@@ -13,14 +12,10 @@ class ErrorPlaceholderComponent extends PrimitiveComponent {
...
-        y: props.pcbY || 0,
-      },
-      schematic_center: {
-        x: props.schX || 0,
-        y: props.schY || 0,
-      },
+      pcbX: props.pcbX || 0,
+      pcbY: props.pcbY || 0,
+      schX: props.schX || 0,
+      schY: props.schY || 0,
```
seveibar, line 18: I would not set defaults, just let them be falsy/undefined

seveibar (approved):

lib/components/primitive-components/ErrorPlaceholder.ts
seveibar: (Dont default to 0)

PR #800, opened 2025-05-13
fixed solder jumper pins not mapping correctly & disable cadModel with empty string

ShiboSoftwareDev:
we are getting the ports for that specific component and they're always rendered in the same order
image


lib/components/normal-components/Jumper.ts
```diff
@@ -77,12 +77,10 @@ export class Jumper<PinLabels extends string = never> extends NormalComponent<
...
     const pinLabelToPortId: Record<string, string> = {}
     // Map pin labels ("1", "2", etc.) to pcb_port_id
-    for (const port of pcb_ports) {
-      const match = port.source_port_id && port.source_port_id.match(/(\d+)$/)
-      if (match) {
-        const label = (parseInt(match[1], 10) + 1).toString()
-        pinLabelToPortId[label] = port.pcb_port_id
-      }
+    for (let i = 0; i < pcb_ports.length; i++) {
+      const port = pcb_ports[i]
```
seveibar, line 81: Use the port hints or the literal label value, the order isnt guaranteed _by the api contract_, it doesnt matter if it works now, it _will_ break if the api contract doesn't specifically guarantee the order

seveibar (changes requested):

lib/components/normal-components/Jumper.ts
```diff
@@ -77,12 +77,21 @@ export class Jumper<PinLabels extends string = never> extends NormalComponent<
...
+      let label: string
+      if (typeof sourcePort?.pin_number === "number") {
+        label = sourcePort.pin_number.toString()
+      } else if (Array.isArray(sourcePort?.port_hints)) {
+        const numericHint = sourcePort.port_hints.find((h: string) =>
+          /^\d+$/.test(h),
+        )
+        label = numericHint || (i + 1).toString()
+      } else {
+        label = (i + 1).toString()
```
seveibar, line 92:
```suggestion
```

lib/components/normal-components/Jumper.ts
```diff
@@ -77,12 +77,21 @@ export class Jumper<PinLabels extends string = never> extends NormalComponent<
...
-        pinLabelToPortId[label] = port.pcb_port_id
+    for (let i = 0; i < pcb_ports.length; i++) {
+      const port = pcb_ports[i]
+      const sourcePort = db.source_port.get(port.source_port_id)
+      let label: string
+      if (typeof sourcePort?.pin_number === "number") {
+        label = sourcePort.pin_number.toString()
+      } else if (Array.isArray(sourcePort?.port_hints)) {
+        const numericHint = sourcePort.port_hints.find((h: string) =>
+          /^\d+$/.test(h),
```
seveibar, line 88:
```suggestion
          /^(pin)?\d+$/.test(h),
```

ShiboSoftwareDev: we just want the number to add it to the map
```
{
  type: "source_port",
  source_port_id: "source_port_8",
  name: "pin3",
  pin_number: 3,
  port_hints: [ "pin3", "3" ],
  source_component_id: "source_component_3",
  subcircuit_id: "subcircuit_source_group_0",
}
```

seveibar: You should handle the case where port hints only contains “pin3”, in this case you would still test for the presence of “pin” but would remove it eg with a replace

lib/components/base-components/NormalComponent/NormalComponent.ts
```diff
@@ -924,6 +924,7 @@ export class NormalComponent<
 
     const pcb_component = db.pcb_component.get(this.pcb_component_id!)
 
+    if (cadModel === "") return
```
seveibar, line 927: Dont use empty strings

```

Avoid using empty strings ("") to represent absence, nullability, or special states. Use null, undefined, or omit the field instead.
```

```suggestion
    if (cadModel === null) return
```

seveibar (approved):
Good except for cad model empty string

btw it is common for a jumper to have repeat labels, eg multiple GND or VCC, this code currently wont work in those cases

ShiboSoftwareDev:
@seveibar I reverted as camodel doesn't accept null in props, let's do in another pr

lib/components/normal-components/Jumper.ts
```diff
@@ -77,12 +77,25 @@ export class Jumper<PinLabels extends string = never> extends NormalComponent<
...
+      const sourcePort = db.source_port.get(port.source_port_id)
+      let pinLabel = ""
+      if (typeof sourcePort?.pin_number === "number") {
+        pinLabel = sourcePort.pin_number.toString()
+      } else if (Array.isArray(sourcePort?.port_hints)) {
+        let matchedHint = sourcePort.port_hints.find((h: string) =>
+          /^(pin)?\d+$/.test(h),
+        )
+        if (matchedHint) {
+          if (/^pin\d+$/.test(matchedHint)) {
```
seveibar, line 91: No need to test, you can just always run the replacement

seveibar (approved):

PR #808, opened 2025-05-16
add support circular hole with rect pad platedhole

techmannih:
fix https://github.com/tscircuit/tscircuit/issues/624


lib/components/primitive-components/PlatedHole.ts
```diff
@@ -28,6 +32,9 @@ export class PlatedHole extends PrimitiveComponent<typeof platedHoleProps> {
     if (props.shape === "oval" || props.shape === "pill") {
       return { width: props.outerWidth, height: props.outerHeight }
     }
+    if (props.shape === "circularHoleWithRectPad") {
```
seveibar, line 35: Enum convention is dashes in props, props has an error

image

seveibar: I think we should maybe switch to snake case enums in props, @techmannih lets refactor to snake case enums. Dash enums are css convention but i think we can start to move to snake case because its just better in ides and matches circuit json better

tests/components/primitive-components/pcb-circular-hole-rect-plated.test.tsx
```diff
@@ -0,0 +1,23 @@
+import { test, expect } from "bun:test"
+import { getTestFixture } from "tests/fixtures/get-test-fixture"
+
+test("pcb circular hole rect plated", () => {
+  const { project } = getTestFixture()
```
seveibar, line 5:
```suggestion
  const { circuit } = getTestFixture()
```

seveibar (commented):
Very nice work @techmannih , see comments

lib/components/primitive-components/PlatedHole.ts
seveibar: Make sure to update to snake case in the props repo

seveibar (changes requested):
Waiting on snake case

bun.lock
```diff
@@ -57,7 +57,7 @@
         "@tscircuit/footprinter": "*",
         "@tscircuit/infgrid-ijump-astar": "*",
         "@tscircuit/math-utils": "*",
-        "@tscircuit/props": "*",
+        "@tscircuit/props": "^0.0.183",
```
seveibar, line 60: this should have failed the check dependency test 🤔 

```suggestion
        "@tscircuit/props": "*",
```

seveibar: https://github.com/tscircuit/dependency-check/issues/2

seveibar (approved):

PR #811, opened 2025-05-17
GND symbol goes down slightly on right/left facing pins

ShiboSoftwareDev:


seveibar (changes requested):
Bad junctions (junction indicates that two traces are crossing)

image

tests/examples/example11-net-symbol.test.tsx
```diff
@@ -90,7 +90,7 @@ test("schematic net symbol", () => {
       <trace schDisplayLabel="GND2" from=".U1 > .pin13" to=".R1 > .pin2" />
       <trace schDisplayLabel="GND2" from=".U1 > .pin20" to=".R1 > .pin2" />
       <trace schDisplayLabel="GND2" from=".U1 > .pin28" to=".R1 > .pin2" />
-      <trace from=".U1 > .pin4" to=".R1 > .pin2" />
+      <trace schDisplayLabel="GND2" from=".U1 > .pin4" to=".R1 > .pin2" />
```
seveibar, line 93: lets revert this, now that i have the context it doesnt make sense to change the test this way

```suggestion
      <trace from=".U1 > .pin4" to=".R1 > .pin2" />
```

ShiboSoftwareDev: @seveibar you said it's a redundant trace, should we make it automatically create a ground symbol if it's in a net with another ground symbol?

ShiboSoftwareDev: maybe someone does want to draw the trace alongside the symbol,
and there is this case where two traces go into the same symbol, but this case only looks good if the ports are close to each other
image

ShiboSoftwareDev: what do you think?

lib/components/primitive-components/Trace/create-downward-net-label-ground-symbol.ts
```diff
@@ -48,52 +48,57 @@ export const createDownwardNetLabelGroundSymbol = (
...
+  ]
+
+  if (isLeftOrRight) {
+    edges.push({
+      from: horizontalEndPos,
+      to: groundSymbolPos,
+    })
+  }
+
+  const junctions = [groundSymbolPos]
```
seveibar, line 97: sorry, I'm looking at it and I don't think you're supposed to have a junction here either. A junction is where two traces cross and connect, but in this case there is no other trace

```suggestion
  const junctions = []
```

seveibar (approved):

ShiboSoftwareDev:
@seveibar the junction was acting like a port to the ground symbol, the symbol must be changed, I'll do that later

ShiboSoftwareDev:
should I add a red port like the other symbols or should I not?

seveibar:
@ShiboSoftwareDev I'm not sure we should have those little dots if it's not in the reference design (kicad/eagle)

seveibar (approved):

PR #826, opened 2025-05-25
feat: add group relationship in pcb_component, sch_component and source_component

imrishabh18:
Ref - https://github.com/tscircuit/pcb-viewer/issues/227

### DIFF
--- a/lib/components/primitive-components/Group/Group.ts
+++ b/lib/components/primitive-components/Group/Group.ts
@@ -33,6 +33,7 @@ export class Group<Props extends z.ZodType<any, any, any> = typeof groupProps>
   implements ISubcircuit
 {
   pcb_group_id: string | null = null
+  schematic_group_id: string | null = null
   subcircuit_id: string | null = null
 
   _hasStartedAsyncAutorouting = false
@@ -60,6 +61,12 @@ export class Group<Props extends z.ZodType<any, any, any> = typeof groupProps>
     db.source_group.update(source_group.source_group_id, {
       subcircuit_id: this.subcircuit_id!,
     })
+
+    for (const child of this.children) {
+      db.source_component.update(child.source_component_id!, {
+        source_group_id: this.source_group_id!,
+      })
+    }
   }
 
   doInitialSourceParentAttachment() {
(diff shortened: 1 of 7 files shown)


lib/components/primitive-components/Group/Group.ts
```diff
@@ -613,6 +626,29 @@ export class Group<Props extends z.ZodType<any, any, any> = typeof groupProps>
...
+    if (this.root?.schematicDisabled) return
+    const { db } = this.root!
+    const { _parsedProps: props } = this
+    const schematic_group = db.schematic_group.insert({
+      is_subcircuit: this.isSubcircuit,
+      subcircuit_id: this.subcircuit_id!,
+      name: this._parsedProps.name,
+      center: this._getGlobalSchematicPositionBeforeLayout(),
+      width: 0,
+      height: 0,
```
seveibar, line 639: you're not updating these to match the content anywhere- likely to be buggy

tests/components/primitive-components/source-group-id-in-source-component.test.tsx
```diff
@@ -0,0 +1,243 @@
...
+        <resistor name="R2" footprint={"0402"} resistance={100} />
+      </group>
+      <capacitor name="C1" footprint={"0402"} capacitance={100} />
+    </board>,
+  )
+
+  circuit.renderUntilSettled()
+
+  const schComponents = circuit.db.schematic_component.list()
+  expect(schComponents).toMatchInlineSnapshot(`
```
seveibar, line 20: this snapshot is going to change too frequently, the main point is to test the schematic groups. We can reduce the test to just test for schematic group snapshots

seveibar: you can also test that the schematic components have a schematic group id

tests/components/primitive-components/source-group-id-in-source-component.test.tsx
```diff
@@ -0,0 +1,243 @@
...
+        "name": undefined,
+        "source_group_id": "source_group_1",
+        "subcircuit_id": "subcircuit_source_group_1",
+        "type": "source_group",
+      },
+    ]
+  `)
+
+  const pcbComponents = circuit.db.pcb_component.list()
+  expect(pcbComponents).toMatchInlineSnapshot(`
```
seveibar, line 174: same here, too likely to change frequently, just test for pcb group id

seveibar (approved):

PR #827, opened 2025-05-27
calculate the pcb_group dimension

imrishabh18:
Dimension calculation of the pcb_group according to the child components


seveibar (approved):

seveibar:
We may want to do this after layout in the future (nbd)

PR #835, opened 2025-05-28
Add support for pin order in PinHeader

imrishabh18:


imrishabh18:
Yeah, I get what you are talking about. But the bug I was still trying to resolve is that the direction is not working of schPinArrangement like how it does in the Chip component

seveibar (approved):
See previous comments, but good to merge when ready

PR #848, opened 2025-05-30
feat: add Fuse component support

techmannih:


lib/components/normal-components/Fuse.ts
```diff
@@ -0,0 +1,68 @@
...
+      zodProps: fuseProps,
+      sourceFtype: FTYPE.simple_fuse,
+    }
+  }
+
+  initPorts() {
+    super.initPorts({
+      additionalAliases: {
+        pin1: ["pin1"],
+        pin2: ["pin2"],
```
seveibar, line 25: you can remove this, pin1 is already an alias of pin1

PR #853, opened 2025-05-30
New component schematicbox

Abse2001:


lib/components/base-components/Renderable.ts
```diff
@@ -21,8 +21,8 @@ export const orderedRenderPhases = [
   "SourceTraceRender",
   "SourceAddConnectivityMapKey",
   "SchematicComponentRender",
-  "SchematicPrimitiveRender",
   "SchematicPortRender",
+  "SchematicPrimitiveRender",
```
seveibar, line 25: i'm a bit worried about switching this order, but it doesn't seem to break anything

seveibar (approved):

lib/components/base-components/Renderable.ts
seveibar: are ports ever dependent on primitives? If so, we need to do separate render phases that break up dependencies appropriately, e.g. `SchematicBoxSetOverlaySize` could be a render phase.

Like i said, if this isn't causing issues it's not a big deal, but if ports are somehow dependent on primitives then it is important (footprint ports are dependent on primitives for example)

PR #854, opened 2025-05-30
feat: pinrow pinlabel customization

ShiboSoftwareDev:


seveibar (approved):
Nice fixes: just a note: pinrow pinsorthogonal/pinsparallel are often wrong. Parallel means the text grows in parallel to the direction the the pins grow. This is a footprinter issue iiuc

PR #857, opened 2025-06-01
feat: pinrow pinlabel text alignment options

ShiboSoftwareDev:


seveibar (approved):
nice, good fit for repro7 as well

PR #858, opened 2025-06-01
update switch schematic symbol

techmannih:


seveibar (approved):
introduces bug with transitor symbols?

image

PR #862, opened 2025-06-02
feat: vcc symbol for vcc nets

ShiboSoftwareDev:


lib/components/primitive-components/Trace/Trace.ts
```diff
@@ -783,7 +785,53 @@ export class Trace
...
+            schDisplayLabel: netLabelText!,
+            source_trace_id: this.source_trace_id!,
+          },
+          { db },
+        )
+      }
+      return
+    } else if (
+      lowercaseNetLabelText?.includes("vcc") ||
+      lowercaseNetLabelText?.includes("vdd")
```
seveibar, line 797: Making a note: This is fine for now, in the future we'll have net characteristics (technically we already do but it's quite old code) that will determine whether something is VCC/VDD or GND

seveibar (commented):
there is a huge offset on the text to the right, also the arrow is quite big (probably about 2x as big as it should be).

PR #866, opened 2025-06-04
Validate net selectors

seveibar:
## Summary
- prevent periods in net selectors
- create nets from props with invalid net names throw an error
- add test for invalid net selectors

## Testing
- `npm run format`
- `bun test tests/sel/net-name-period.test.ts`
- `bun test`

------
https://chatgpt.com/codex/tasks/task_b_68406cb56bc8832e9313eef08afa72cf


andrii-balitskyi:
@seveibar what about those starting with a number?

seveibar:
> @seveibar what about those starting with a number?

Yea we will need to add support for testing for that as well!!!

PR #873, opened 2025-06-06
Fixed X and Y type issue in smtpad

techmannih:


seveibar (changes requested):
lots of unrelated changes, including the removal of solder paste. Misleading PR title

PR #880, opened 2025-06-07
pinrow silkscreen labels in the bottom layer

ShiboSoftwareDev:


seveibar (changes requested):
Youre creating it on both layers iiuc

seveibar:
@ShiboSoftwareDev thr snapshots show yellow and blue text overlaid

PR #884, opened 2025-06-08
add support <testpoint /> component

techmannih:


seveibar (changes requested):
why are you inserting smtpads in the source render?

PR #888, opened 2025-06-08
fixed jumper pin issues

ShiboSoftwareDev:


seveibar (approved):
we should know without the user specifying it if the pins are internally connected if they provide the bridged parameter, also the bridged parameter is invalid (it cant support pins with two numbers)

these are preexisting issues, issues for future

PR #890, opened 2025-06-08
Implemented title for fixed size + dynamic sized schematic box

Abse2001:


seveibar (commented):
also we're going to deprecate `titleAnchorPosition`, I don't think it makes sense, use `titleAnchorAlignment` instead

PR #891, opened 2025-06-08
implemented nine point title alignment for schematic box

Abse2001:


lib/components/primitive-components/SchematicBox.ts
```diff
@@ -112,5 +110,126 @@ export class SchematicBox extends PrimitiveComponent<typeof schematicBoxProps> {
...
+      }
+
+      const anchor = props.titleAlignment ?? "bottom_center"
+      const anchorPos = getAnchorPosition(anchor)
+      let titleOffsetY: number
+      let titleOffsetX: number
+      const textAnchor = anchorPos.textAnchor
+      if (isInside) {
+        titleOffsetY = anchor.includes("top")
+          ? -0.05 * height
```
seveibar, line 197: Don't use percentage offsets, it looks terrible as things scale up.

Use fixed-margins e.g. 0.1mm etc.

seveibar (commented):
very close, two remaining concerns

lib/components/primitive-components/SchematicBox.ts
```diff
@@ -112,5 +111,60 @@ export class SchematicBox extends PrimitiveComponent<typeof schematicBoxProps> {
...
+        width,
+        height,
+        isInside,
+      })
+      let titleOffsetY: number
+      let titleOffsetX: number
+      const textAnchor = anchorPos.textAnchor
+      if (isInside) {
+        titleOffsetY = anchor.includes("top")
+          ? -0.1
```
seveibar, line 132: "magic numbers" are bad, just put `const TITLE_PADDING = 0.1` at the top

seveibar (approved):

PR #920, opened 2025-06-14
Fix solderjumper internallyConnectedPins symbol bug

ShiboSoftwareDev:
## Summary
- fix SolderJumper schematic symbol name derivation when pinCount is omitted
- add repro test for using `internallyConnectedPins` on Jumper/SolderJumper

## Testing
- `bun test`

this was the error which now is fixed
image

------
https://chatgpt.com/codex/tasks/task_e_684d4b032b4883328c953439aabb8ce2


seveibar (changes requested):
confused, do you mean to be modifying the SolderJumper.ts file? The Jumper should not use schematic symbols

lib/components/normal-components/Jumper.ts
```diff
@@ -16,10 +16,25 @@ export class Jumper<PinLabels extends string = never> extends NormalComponent<
...
+          if (typeof p === "number") return p
+          if (p.startsWith("pin")) return Number(p.slice(3))
+          return Number(p)
+        })
+        .filter((n) => !Number.isNaN(n))
+      const maxPin = nums.length > 0 ? Math.max(...nums) : 0
+      if (maxPin === 2 || maxPin === 3) {
+        resolvedPinCount = maxPin as 2 | 3
+      }
+    }
```
seveibar, line 33: why are you resolving the pin count if you don't use it?

PR #924, opened 2025-06-14
Implement schematic border to <group /> component

Abse2001:
## Summary
- allow groups to draw schematic borders with padding
- update @tscircuit/props to 0.0.227
- add snapshot test for border padding

## Testing
- `bun test tests/groups/group-schematic-border.test.tsx`
- `bun test tests/groups/group-schematic-border-padding.test.tsx`


lib/components/primitive-components/Group/Group.ts
```diff
@@ -700,6 +702,70 @@ export class Group<Props extends z.ZodType<any, any, any> = typeof groupProps>
...
+    const finalWidth = width + paddingLeft + paddingRight
+    const finalHeight = height + paddingTop + paddingBottom
+
+    db.schematic_box.insert({
+      width: finalWidth,
+      height: finalHeight,
+      x: left,
+      y: bottom,
+      is_dashed: props.border?.dashed ?? false,
+      schematic_component_id: "",
```
seveibar, line 765: we should make this optional in circuit-json (empty strings are always a bad hack and will cause many problems over time!)

```suggestion
```

seveibar (approved):
this is fking awesome

lib/components/primitive-components/Group/Group.ts
seveibar: https://github.com/tscircuit/circuit-json/pull/226

PR #949, opened 2025-06-18
Fix unnamed component render

imrishabh18:
## Summary
- add `AssignNameToUnnamedComponents` render phase
- default unnamed component names before source render
- provide placeholder name for JSX elements missing one
- expose `nextUnnamedComponentId` on subcircuits
- add test for unnamed chip

## Testing
- `bun test tests/repros/repro22-unnamed-component.test.tsx`
- `bun test`

------
https://chatgpt.com/codex/tasks/task_b_68531572b59c8327af87b1e139fceaec


lib/components/base-components/NormalComponent/NormalComponent.ts
```diff
@@ -329,11 +329,7 @@ export class NormalComponent<
   }
 
   _addChildrenFromStringFootprint() {
-    const {
-      name: componentName,
-      pcbRotation: componentRotation,
-      pinLabels,
-    } = this.props
+    const { pcbRotation: componentRotation, pinLabels } = this.props
```
seveibar, line 332:
```suggestion
    const { pcbRotation, pinLabels } = this.props
```

lib/components/base-components/NormalComponent/NormalComponent.ts
```diff
@@ -342,7 +338,12 @@ export class NormalComponent<
       if (this._isFootprintUrl(footprint)) return
       const fpSoup = fp.string(footprint).soup()
       const fpComponents = createComponentsFromCircuitJson(
-        { componentName, componentRotation, footprint, pinLabels },
+        {
+          componentName: this.name,
+          componentRotation,
```
seveibar, line 343:
```suggestion
          componentRotation: pcbRotation,
```

lib/components/base-components/NormalComponent/NormalComponent.ts
seveibar: not the right place for this rename/mapping because the context isn't good

seveibar (approved):
looks correct!

PR #956, opened 2025-06-19
feat(sel): support J refs in selector (sel.J1, sel.J2, etc)

Abse2001:


lib/sel/sel.ts
```diff
@@ -209,8 +209,8 @@ export const sel: Sel = new Proxy(
             {
               get: (_, pinOrSubComponentName: string) => {
                 const pinResult = `.${prop1} > .${pinOrSubComponentName}`
-
-                if (prop1.startsWith("U")) {
+                const chipPrefixes = ["U", "J"]
```
seveibar, line 212: the JumperSel specifies CN is also valid

```suggestion
                const chipPrefixes = ["U", "J", "CN"]
```

seveibar (approved):

PR #958, opened 2025-06-20
fix: enforce schWidth behavior below 1 when pinLabels are present

Anshgrover23:
now, we have full control over schWidth like schHeight i know it can cause overflow issue, but this is the only solution i found out.

before:
 
image

after:

image


seveibar (changes requested):
@Anshgrover23 you need to put more effort into your PRs.

1. Create a new test file. Don't mess up existing tests
2. Delete the code rather than commenting it out so that it's actually mergable if it's a fix
3. Update other snapshots so we can see the impact
4. Write a real PR title that explains what problem your fixing with proper grammar

PR #971, opened 2025-06-21
Display load capacitor value in crystal component

techmannih:


Anshgrover23 (changes requested):
@techmannih can u create a seperate function for this one?

PR #978, opened 2025-06-23
Now transistor/mosfet can access 'gate', 'source', 'drain' types

MustafaMulla29:


MustafaMulla29:
@seveibar is this the correct fix for the type error: `Property 'gate' does not exist on type 'Record<TransistorPinNames, string>'.`?

tests/sel/sel-mosfet-gate-source-drain-access.test.tsx
```diff
@@ -0,0 +1,28 @@
...
+    <board>
+      <mosfet
+        name="Q1"
+        footprint="sot23"
+        channelType="n"
+        mosfetMode="enhancement"
+      />
+      <trace from={sel.Q1.gate} to={sel.net.GND} />
+      <trace from={sel.Q1.source} to={sel.net.V3_3} />
+      <trace from={sel.Q1.drain} to={sel.net.V3_3} />
```
seveibar, line 18: GND would be a better choice

seveibar (approved):

PR #981, opened 2025-06-23
Expanded CommonPinNames in sel-utility-types.ts to include serial and USB pin identifiers

techmannih:


seveibar (changes requested):
these aren't all common, use sel.U1(ChipClass).... for these

PR #998, opened 2025-06-26
add board information to the board silkscreen with `printBoardInformationToSilkscreen`

imrishabh18:


lib/RootCircuit.ts
```diff
@@ -48,8 +51,12 @@ export class RootCircuit {
     // TODO rename to rootCircuit
     this.root = this
     this.platform = platform
-    this.projectUrl = projectUrl
+    this.projectUrl = projectUrl ?? platform?.url
     this.pcbDisabled = platform?.pcbDisabled ?? false
+    this.printBoardInformationToSilkscreen =
+      platform?.printBoardInformationToSilkscreen ?? false
+    this.version = platform?.version ?? undefined
+    this.name = platform?.projectName ?? undefined
```
seveibar, line 59: don't need to do this,

Bound.doInitialPcbComponentRender (or something like that- you can also add it earlier, there are steps for creating children etc.) can access this.root.platform

seveibar (approved):

PR #1000, opened 2025-06-26
implement obstacles for cutout components

ShiboSoftwareDev:


ShiboSoftwareDev:
@seveibar what do you expect from the circle? we don't support circle obstacles

PR #1002, opened 2025-06-26
fix: subcircuit having multiple traces between same pins

imrishabh18:
`getSimpleRouteJsonFromCircuitJson` collects all the `source_trace` elements whose `subcircuit_id` is this group OR any descendant sub-circuit (even when that subcircuit is having traces which is already routed). And then that simpleRouteJson is passed to the autorouter where the already resolved `source_trace` gets dragged.


lib/components/primitive-components/Group/Group.ts
```diff
@@ -397,13 +397,21 @@ export class Group<Props extends z.ZodType<any, any, any> = typeof groupProps>
     debug(`[${this.getString()}] starting local autorouting`)
     const autorouterConfig = this._getAutorouterConfig()
 
-    // Get the routing problem in SimpleRouteJson format
-    const { simpleRouteJson, connMap } = getSimpleRouteJsonFromCircuitJson({
+    const tracesNeedsRouting = (this.selectAll("trace") as Trace[])
```
seveibar, line 400:
```suggestion
    const sourceTraceIdsToRoute = (this.selectAll("trace") as Trace[])
```

lib/components/primitive-components/Group/Group.ts
```diff
@@ -397,13 +397,21 @@ export class Group<Props extends z.ZodType<any, any, any> = typeof groupProps>
...
+
+    const { simpleRouteJson } = getSimpleRouteJsonFromCircuitJson({
       db,
       minTraceWidth: this.props.autorouter?.minTraceWidth ?? 0.15,
       subcircuit_id: this.subcircuit_id,
     })
 
+    // throw away any connection that is not in our "needs routing" set
+    simpleRouteJson.connections = simpleRouteJson.connections.filter((conn) =>
+      tracesNeedsRouting.includes(conn.name),
```
seveibar, line 412:
```suggestion
      directChildTraces.includes(conn.name),
```

PR #1006, opened 2025-06-27
fix netlabel const netLabel source net id error

Anshgrover23:
image


seveibar (changes requested):
the repro must be put in the repros/ folder or simplified (it is more complex that it needs to be, complex reproductions must be labeled `repro*` and put in the repros folder)

the fix is incorrect, all nets should have a source_net_id, the underlying issue is something else

seveibar:
the fix most likely involves introducing CreateNetsFromProps for netlabels

PR #1010, opened 2025-06-28
update schematic symbol and test point snapshot

techmannih:


Anshgrover23 (changes requested):
@techmannih create one example test like a chip of 3 pins connected to 3 test point each so that we get to know this size problem solved. I hope you understand.
CC; @seveibar

PR #1011, opened 2025-06-28
update schematic symbol and test point snapshot

techmannih:


Anshgrover23 (changes requested):
@techmannih create one example test like a chip of 3 pins connected to 3 test point each so that we get to know this size problem solved. I hope you understand.
CC; @seveibar

Anshgrover23:
@techmannih actually, that i was trying to say, how can we check if ti right small, by creating a test suggested above, what's wrong in creating a new unit test.

ShiboSoftwareDev (approved):

ShiboSoftwareDev:
@techmannih like ansh said let's put this symbol and other netlabels in a sch connected to a chip, let's see how it looks

Anshgrover23 (approved):
test failing just see that.

PR #1012, opened 2025-06-28
Add types for the package `css-select` failing the `build`

imrishabh18:
The source file is present in the installed package but not the compiled `.d.ts` declaration file, or the package `@types/css-select`. So, had to add the decalartion types in our project.

### DIFF
--- a/lib/components/base-components/PrimitiveComponent/PrimitiveComponent.ts
+++ b/lib/components/base-components/PrimitiveComponent/PrimitiveComponent.ts
@@ -26,7 +26,6 @@ import type { IGroup } from "lib/components/primitive-components/Group/IGroup"
 import type { Ftype } from "lib/utils/constants"
 import { selectOne, selectAll, type Options } from "css-select"
 import {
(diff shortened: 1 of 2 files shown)


tests/snippet-imports/snippet-import1-contribution-board.test.tsx
```diff
@@ -1,9 +1,11 @@
 import { test, expect } from "bun:test"
-import { importSnippet } from "@tscircuit/import-snippet"
 import { Circuit } from "lib"
 import * as tscircuitCore from "lib"
 
 test.skip("snippet-import1-contribution-board", async () => {
+  // Import moved inside test to avoid module resolution issues when @tscircuit/core is not installed
+  const { importSnippet } = await import("@tscircuit/import-snippet")
```
seveibar, line 7: ??

seveibar (commented):
Multiple changes, not sure i understand why, in general we should never have dynamic imports? Also typecheck is passing on main

PR #1033, opened 2025-07-05
add support for the autorouter config `preset` prop

imrishabh18:


seveibar (changes requested):
you can use the fake  autorouting server for this, i don't think we should have any tests with process.env.CI because it messes up local development

seveibar (approved):

PR #1037, opened 2025-07-07
Fix the missing handling of `preset` property of autorouter config

imrishabh18:


lib/components/primitive-components/Group/Group.ts
```diff
@@ -874,6 +874,46 @@ export class Group<Props extends z.ZodType<any, any, any> = typeof groupProps>
     }
   }
 
+  _getPresetConfig(
```
seveibar, line 877: (this should not be an instance method, can be moved to separate file `getPresetAutoroutingConfig`

```suggestion
  _getPresetAutoroutingConfig(
```

lib/components/primitive-components/Group/Group.ts
```diff
@@ -898,29 +959,11 @@ export class Group<Props extends z.ZodType<any, any, any> = typeof groupProps>
       }
     }
```
seveibar, line 961: all of this section can just be `return getPresetAutoroutingConfig(autorouter)` (even if autorouter is not defined, getPresetAutoroutingConfig can handle it)

seveibar (approved):
(my video should clarify my comments here, sent via slack)

PR #1064, opened 2025-07-13
Repro for pinheaders `pin1` and `pin4` not connecting

MustafaMulla29:
image


tests/repros/repro28-pinheader-connections.test.tsx
```diff
@@ -0,0 +1,226 @@
+import { test, expect } from "bun:test"
+import { getTestFixture } from "../fixtures/get-test-fixture"
+import { sel } from "lib/sel"
+
+test(
+  "pinheader connections with netlabels",
```
seveibar, line 6:
```suggestion
  "pinheader connections with netlabels",
```

describe what you're reproducing here

MustafaMulla29: Ok. The pin1 and pin4 are not getting connected of the JP1

PR #1069, opened 2025-07-14
chore: update schematic-symbols to v0.0173

techmannih:


seveibar (approved):
Can you follow up and modify the transistor tests to have net labels that help understand which pin is which, eg “PIN1” and “SOURCE” and “DRAIN” etc

PR #1079, opened 2025-07-18
Added repro for externally connected pins bug

Abse2001:


tests/repros/repro29-externally-connected-pins.test.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+  let caughtError: any = null
+  try {
+    circuit.render()
+  } catch (e) {
+    caughtError = e
+    console.error("Render error:", e)
+  }
+
+  expect(caughtError).toBeTruthy() // ensures render threw an error
```
seveibar, line 58:
```suggestion
  circuit.render()
```

seveibar (approved):

PR #1084, opened 2025-07-22
Fix Missing Cutout from Footprint

abimaelmartell:


seveibar (commented):
can you add a quick snapshot test, you can even just create a <chip footprint="m2host" name="U1" />

seveibar (approved):
yes excellent muhahah

PR #1102, opened 2025-07-26
simulation_voltage_source from PowerSource component

ShiboSoftwareDev:


lib/components/normal-components/PowerSource.ts
```diff
@@ -43,6 +43,14 @@ export class PowerSource extends NormalComponent<
       are_pins_interchangeable: false,
     } as any)
     this.source_component_id = source_component.source_component_id
+
+    if (props.voltage) {
+      db.simulation_voltage_source.insert({
+        voltage: props.voltage,
```
seveibar, line 49: we should also put a source_component_id on the simulation_voltage_source (I think it can be required)

seveibar (approved):

seveibar:
@ShiboSoftwareDev note that generally this isn't how people will create voltage sources, instead we'll create voltage sources using any chip's `pinAttributes`

PR #1132, opened 2025-08-07
Hide pinlabels by default and control the visibility by supporting showSilkscreenPinlabels prop

MustafaMulla29:


lib/components/normal-components/PinHeader.ts
```diff
@@ -19,16 +19,31 @@ export class PinHeader extends NormalComponent<typeof pinHeaderProps> {
...
+          footprintString = `pinrow${pinCount}`
+        } else {
+          return null
         }
-        return `pinrow${pinCount}_p${pitch}_id${holeDiameter}_od${platedDiameter}`
       }
-      if (!holeDiameter && !platedDiameter) {
-        return `pinrow${pinCount}`
+
+      // Add _hidesilkscreenpinlabels if showSilkscreenPinLabels is false or undefined
```
seveibar, line 41: why does the comment not match the implementation? `nopinlabels` vs `_hidesilkscreenpinlabels`

tests/components/normal-components/pinheader-show-silkscreen-pin-labels.test.tsx
```diff
@@ -0,0 +1,82 @@
...
+      />
+    </board>,
+  )
+
+  circuit.render()
+
+  expect(circuit).toMatchPcbSnapshot(
+    import.meta.path + "-explicit-footprint-always-shows",
+  )
+})
```
seveibar, line 82: put these all in the same pcb snapshot, in a single test, and give a name to each pinheader that indicates what scenario is being tested. This way we have one easily-read snapshot

`name="P4_showSilkscreenPinLabels_false"`

lib/components/base-components/NormalComponent/NormalComponent.ts
```diff
@@ -344,6 +346,9 @@ export class NormalComponent<
           footprint,
           pinLabels,
           pcbPinLabels,
+          showSilkscreenPinLabels: isExplicitFootprint
```
seveibar, line 349: remove this

seveibar (commented):
lots of snapshot updates due to previous code, maybe just copy over the relevant changes to a new PR to keep it clean, i think the approach is coming together but it seems like the approach isn't consistent.

PR #1142, opened 2025-08-12
fix(resonator): display frequency and load capacitance units

techmannih:


seveibar (approved):

seveibar:
@techmannih can you refactor to just have one test that has arranges each different symbol, we don't need 8 tests just to test that symbols display properly in core

PR #1155, opened 2025-08-13
throw when source trace lacks pcb trace

pxlpal:
## Summary
- ensure `getObstaclesFromCircuitJson` throws if a source_trace has no matching pcb_trace
- exclude source traces when generating obstacles during trace rendering
- add regression test for missing pcb_trace on source_trace

## Testing
- `bunx biome check lib/utils/obstacles/getObstaclesFromCircuitJson.ts lib/components/primitive-components/Trace/Trace_doInitialPcbTraceRender.ts tests/utils/obstacles/source-trace-missing-pcb-trace.test.ts`
- `bun test tests/utils/obstacles/source-trace-missing-pcb-trace.test.ts tests/components/primitive-components/cutout-obstacles.test.tsx`

------
https://chatgpt.com/codex/tasks/task_b_689caacc0e58832c9d36f71453871320


seveibar (changes requested):
I think @imrishabh18  is working on issue

This should be a DRC check (I think it already exists in tscircuit/checks) and be at the appropriate render stage, it doesn't make sense to do it when we're just trying to pull out obstacles

PR #1169, opened 2025-08-21
fix: invalid pinLabel fails to render the whole component

imrishabh18:


lib/components/normal-components/Chip.ts
```diff
@@ -1,20 +1,32 @@
...
 
 export class Chip<PinLabels extends string = never> extends NormalComponent<
   typeof chipProps,
   PinLabels
 > {
   schematicBoxDimensions: SchematicBoxDimensions | null = null
+  private _invalidPinLabelMessages: string[] = []
+
+  constructor(props: z.input<typeof chipProps>) {
+    const filteredProps = { ...props }
```
seveibar, line 17: Call super after this, dont have two places where super is called

seveibar (commented):
Yes maybe the error can just be "source_property_ignored_warning"

imrishabh18:
> Yes maybe the error can just be "source_property_ignored_warning"

Why the `schematic_error` type should start with `source_`? You mean this.

image

PR #1170, opened 2025-08-22
feat: handle worker origin in getClientOrigin

techmannih:
## Summary
- handle worker `self.location.origin` in `getClientOrigin`
- add tests verifying origin selection in browser and worker contexts


seveibar (approved):
where did you see this issue?

PR #1184, opened 2025-08-26
Fix/transistor pins

thesloppyguy:
Fix [742](https://github.com/tscircuit/tscircuit/issues/742):

Pin Mapping for Transistor was causing conflicts for npn spec.

Updated with robust custom mapping for transistor.

### DIFF
--- a/lib/components/normal-components/Transistor.ts
+++ b/lib/components/normal-components/Transistor.ts
@@ -5,6 +5,8 @@ import {
   type TransistorPorts,
 } from "lib/utils/constants"
 import { NormalComponent } from "../base-components/NormalComponent/NormalComponent"
+import { Port } from "../primitive-components/Port"
+import { symbols } from "schematic-symbols"
 
 export class Transistor extends NormalComponent<
   typeof transistorProps,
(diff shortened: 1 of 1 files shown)


thesloppyguy:
Hi @seveibar for the workflow tests, I an unable to find the 1 failed test case. logs dont show anything and running from local also gives me no errors. 

I changed one of the testcase snapshots do you think it is related to that ?

lib/components/base-components/NormalComponent/NormalComponent.ts
```diff
@@ -239,6 +239,17 @@ export class NormalComponent<
...
+          if (
+            symPort.labels.includes("emitter") &&
+            opts.additionalAliases?.[`pin${pinNumber}`].includes("collector")
+          )
+            symPort.labels.splice(symPort.labels.indexOf("emitter"), 1)
+
+          if (
+            symPort.labels.includes("collector") &&
+            opts.additionalAliases?.[`pin${pinNumber}`].includes("emitter")
+          )
```
seveibar, line 251: this custom logic shouldn't be here in `NormalComponent` which is basically the base class for everything

thesloppyguy:
@seveibar  Tested the same with `connections` as suggested. 

BEFORE  
image
AFTER
Screenshot 2025-08-27 at 3 46 25 PM

Let me know if you want me to add/modify test cases for the same.

thesloppyguy:
Closing this as the bounty was sniped.

PR #1185, opened 2025-08-26
fix: add minimum gap of `1mm` to pcbPack

imrishabh18:


lib/components/primitive-components/Group/Group_doInitialPcbLayoutPack.ts
```diff
@@ -15,6 +15,7 @@ import {
 import { translate, rotate, compose } from "transformation-matrix"
 import Debug from "debug"
 
+const MIN_GAP = "1mm"
```
seveibar, line 18:
```suggestion
const DEFAULT_MIN_GAP = "1mm"
```

seveibar (approved):

Issue #1204, opened 2025-08-27
Regression: Basic Schematic Trace Routing Failures


seveibar:
<img width="2448" height="706" alt="Image" src="https://github.com/user-attachments/assets/3cec63ae-ba6f-4a3c-baf8-93a2d98a4f67" />

/bounty $50

Issue #1203, opened 2025-08-27
Regression: Duplicate overlapping net labels


seveibar:
<img width="2686" height="1712" alt="Image" src="https://github.com/user-attachments/assets/d4da78e5-5293-4217-b99d-d1682979efa3" />

/bounty $50

Issue #1202, opened 2025-08-27
Regression: Netlabel facing wrong direction


seveibar:
<img width="2182" height="746" alt="Image" src="https://github.com/user-attachments/assets/1c0926d1-67d7-49ce-b45d-8d17d6e62165" />

/bounty $50

Issue #1201, opened 2025-08-27
Regression: Missing Ground Label in normal-components-schematic.snap.svg


seveibar:
Related to Group_doInitialSchematicTraceRender.ts

<img width="2690" height="718" alt="Image" src="https://github.com/user-attachments/assets/58dc15ca-97d8-4139-b726-6d001b3d5be2" />

PR #1206, opened 2025-08-28
Fix schematic trace routing failures

t431w:
These changes fix all the issues — disabled the MSP-based schematic trace router by default, so basic schematic traces now render using the stable routing logic

Ref #1201
Ref #1202 
Ref #1203 
Ref #1204 

/claim #1201
/claim #1202 
/claim #1203 
/claim #1204


seveibar:
We're migrating to MSP schematic tracing

PR #1213, opened 2025-08-29
render net label instead of traces in Trace_doInitialSchematicTraceRender when MSP routing is enabled

AnasSarkiz:
Fixes: #1202 
Fixes: #1204 

/claim #1202

Can I get a tip for the second, as I can't claim both


AnasSarkiz:
@seveibar @imrishabh18, can I get A review

seveibar (changes requested):
@AnasSarkiz good to see you

we are deprecating this function, so its not valid to enabled it. The feature flag will remain on indefinitely, the code is only there as a reference to understand regressions

PR #1224, opened 2025-08-30
Fix: Correctly resolve nets across subcircuit boundaries for schematic routing

### LINKED ISSUE #1201: Regression: Missing Ground Label in normal-components-schematic.snap.svg

ShiboSoftwareDev:
the net label was not being applied to a port on the main board when that port was connected via a trace to a net defined within a child subcircuit.

/claim #1201
/closes #1201


seveibar (commented):
🤔 breaks subcircuit isolation. I think this will cause adjacent subcircuits to reroute traces. Maybe you could quickly just test what happens if there are two subcircuitswith both internal and external nets between them just to sanity check

seveibar (approved):

PR #1225, opened 2025-08-30
Added repro for automatic schematic

MustafaMulla29:


seveibar (approved):
just a note: this seems different from the output in the matchpack repo- you might have discovered a bug

seveibar:
@MustafaMulla29 the bug: I think SJ1 isn't properly being rotated according to the matchpack algorithm

image

PR #1232, opened 2025-08-30
fix: floating net label over traces

imrishabh18:


seveibar (changes requested):
it doesn't fix the referenced issue

seveibar:
@imrishabh18 ah ok a bit of confusion. That net should be labeled, we're not trying to exclude it, we just dont want it to overlap something else. It should not be excluded because it has a source net/name (this is best convention)

PR #1234, opened 2025-08-31
Fix schematic trace routing failures

t431w:
Ref #1204

/claim #1204


seveibar (commented):
this fixes a valid issue (CC @imrishabh18) but it doesn't fix the issue you linked to (which is for fixing schematic trace routing). Just because the test shows a change doesn't mean the underlying issue is fixed

seveibar:
the underlying issue has to do with schematic trace routing, you would need to debug by inspecting the `InputProblem` for the SchematicTracePipelineSolver, probably pasting the input into the schematic trace debugger here: https://schematic-trace-solver.vercel.app/?fixture=%7B%22path%22%3A%22site%2FPasteInput.page.tsx%22%7D

Issue #1238, opened 2025-08-31
Automatic Silkscreen Reference Designator (i.e. R1/C1 text) Adjustment for passives (resistors, capacitors and inductors)


seveibar:
<img width="1370" height="1310" alt="Image" src="https://github.com/user-attachments/assets/1d4eafeb-4580-4be7-a831-102489221d8c" />

seveibar:
@baeoc it is not fixed. The fix is in this repo

seveibar:
I'm giving to @MustafaMulla29 for 2-3 days because he has a valid implementation but just needs to improve the code

PR #1239, opened 2025-08-31
fix: Adjust reference designators for passives

t431w:
Ref #1238 
/claim #1238


seveibar (changes requested):
should not be in the board file. Maybe you need to introduce a new render phase. It should be inside NormalComponent and broken out into the relevant file e.g. NormalComponent_doInitialSilkscreenOverlapAdjustment

seveibar:
the snapshots show a buggy implementation with reference designators being placed asymmetrically

PR #1246, opened 2025-09-01
New render phase for automatic silkcreen adjustment

### LINKED ISSUE #1238: Automatic Silkscreen Reference Designator (i.e. R1/C1 text) Adjustment for passives (resistors, capacitors and inductors)

MustafaMulla29:
/closes #1238
/claim #1238


lib/components/base-components/NormalComponent/NormalComponent_doInitialSilkscreenOverlapAdjustment.ts
```diff
@@ -0,0 +1,396 @@
...
+  const { db } = component.root!
+
+  // Only process if this is a subcircuit or board
+  if (!component.isSubcircuit && component.componentName !== "Board") return
+
+  // Get all silkscreen texts in this component and its children that are reference designators for passives
+  const allSilkscreenTexts = db.pcb_silkscreen_text.list()
+  const passiveReferenceTexts = allSilkscreenTexts.filter((text) => {
+    // Check if this is a passive component reference designator
+    return /^(R\d+|C\d+|L\d+)$/i.test(text.text)
```
seveibar, line 20: avoid regex! Remove!!

you can have logic like this:

```
if (!component._adjustSilkscreenTextAutomatically) return
```

then make sure to set _adjustSilkscreenTextAutomatically = true inside Resistor.ts, Capacitor.ts and Inductor.ts

lib/components/base-components/NormalComponent/NormalComponent_doInitialSilkscreenOverlapAdjustment.ts
```diff
@@ -0,0 +1,396 @@
...
+  })
+
+  if (passiveReferenceTexts.length === 0) return
+
+  // Get all other silkscreen elements that could cause overlaps
+  const allSilkscreenPaths = db.pcb_silkscreen_path.list()
+  const allSilkscreenRects = db.pcb_silkscreen_rect.list()
+  const allSilkscreenCircles = db.pcb_silkscreen_circle.list()
```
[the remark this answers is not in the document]
seveibar: we want to avoid querying the db if possible because you're not doing enough filtering otherwise and it gets quite complex.

MustafaMulla29:
Note:The code only considers opposite sides.

lib/components/base-components/NormalComponent/NormalComponent_doInitialSilkscreenOverlapAdjustment.ts
```diff
@@ -0,0 +1,353 @@
+import type { NormalComponent } from "./NormalComponent"
+
+interface SilkscreenElement {
+  bounds: Bounds
+  layer: string
+  type: "text" | "path" | "rect" | "circle"
+  pcb_component_id?: string
+  id: string
```
seveibar, line 8: `id` is a banned word. Also this type name is too similar to other type names, i think you can get a better name e.g. `SilkscreenElementWithBounds`

lib/components/base-components/NormalComponent/NormalComponent_doInitialSilkscreenOverlapAdjustment.ts
```diff
@@ -0,0 +1,353 @@
...
+          pcbComp.pcb_component_id,
+          db,
+          elements,
+        )
+      }
+    }
+  }
+
+  return elements
+}
```
seveibar, line 170: this is hacky, let's get `component.getSubcircuit().selectAll("normalcomponent")` working

MustafaMulla29: yea working on it

lib/components/base-components/NormalComponent/NormalComponent_doInitialSilkscreenOverlapAdjustment.ts
```diff
@@ -0,0 +1,353 @@
...
+    return {
+      x: component._parsedProps.pcbX,
+      y: component._parsedProps.pcbY,
+    }
+  }
+
+  // Fallback to DB lookup (single query per component)
+  const pcbComponent = db.pcb_component.get(component.pcb_component_id)
+  return pcbComponent?.center || null
+}
```
seveibar, line 260: remove. This is a totally invalid way to get a component center. Component's have a function for getting the bounds DO NOT USE DB!

lib/components/base-components/NormalComponent/NormalComponent_doInitialSilkscreenOverlapAdjustment.ts
```diff
@@ -0,0 +1,353 @@
...
+      }
+    }
+  }
+}
+
+/**
+ * Collect silkscreen elements from nearby components using spatial awareness
+ * This reduces the need to query the entire database
+ */
+function collectNearbySilkscreenElements(
```
seveibar, line 120: this shouldn't be something you have to do

seveibar (commented):
ok so the good news is you have a mostly valid implementation that is doing the right thing (this is great), but the code is hacky and doesn't use the methods i mentioned that would be far far more stable

Just as an example pcbX and pcbY is never a reliable way to get the center of the component!!!!! It could be the XY relative to the parent!

Use `@tscircuit/math-utils` for computing the distance between bounds etc.

You may want to do this across a few PRs just to make sure you're moving in the right direction, e.g. test this one: https://github.com/tscircuit/core/issues/1250

I will assign you the issue so people don't steal it so you can work more slowly towards it.

MustafaMulla29:
@seveibar Component props don't have calculated positions `anchor_position`, `font_size`- only raw input data. Positioning is calculated during render and stored in database only. So uses .selectAll() + minimal db for accurate overlap detection.

MustafaMulla29:
> @MustafaMulla29 in the some snapshots Text is overlapping with pads
This pr is WIP and also this pr is not about overlapping pads, it's about overlapping silkscreen.

MustafaMulla29:
> @MustafaMulla29 in the some snapshots Text is overlapping with pads

This pr is WIP and also this pr is not about overlapping pads, it's about overlapping silkscreen.

lib/components/base-components/NormalComponent/NormalComponent_doInitialSilkscreenOverlapAdjustment.ts
```diff
@@ -0,0 +1,342 @@
...
+    }
+  }
+
+  return elements
+}
+
+/**
+ * Get component center using component's built-in bounds system
+ */
+function getComponentCenterFromCache(
```
seveibar, line 238: THERES NO CACHE! Don't lie in your function names! Also don't use a cache!!! Also there's no need for this function!!!! Remove it, it also shouldn't take the unused db!!!! `component._getPcbCircuitJsonBounds().center` is extremely clear

lib/components/base-components/NormalComponent/NormalComponent_doInitialSilkscreenOverlapAdjustment.ts
```diff
@@ -0,0 +1,342 @@
...
+  } catch (error) {
+    // If bounds calculation fails, return null
+    return null
+  }
+}
+
+/**
+ * Calculate bounds for silkscreen text using @tscircuit/math-utils
+ */
+function getSilkscreenTextBounds(text: any): Bounds {
```
seveibar, line 257: remove the any cast

PR #1248, opened 2025-09-01
Prevent erroneous junctions at trace crossings

### LINKED ISSUE #1245: Invalid Junction (small green circle indicating two traces cross) in repro48
These nets are not connected, therefore there should be no junction

Relevant code: https://github.com/tscircuit/core/blob/main/lib/components/primitive-components/Group/Group_doInitialSchematicTraceRender/compute-junctions.ts

Image

/bounty $75

ShiboSoftwareDev:
/claim #1245
/closes #1245

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@tscircuit/core",
   "type": "module",
-  "version": "0.0.682",
+  "version": "0.0.683",
   "types": "dist/index.d.ts",
   "main": "dist/index.js",
   "module": "dist/index.js",


seveibar (commented):
there's something concerning that i don't understand, why is the junction removed in example7?

ShiboSoftwareDev:
pin7 only connects to pin3 
image

seveibar:
sorry for the delay @ShiboSoftwareDev , i'm still confused because your logic doesn't relate to connectivity maps or anything like that, it relates to corners. I just can't figure out the junction was previously there, is there a larger issue related to establishing connectivity? Like the input to computeJunctions?

PR #1256, opened 2025-09-02
fix(layout): Correctly apply schRotation in matchpack schematic layout

### LINKED ISSUE #1226: Repro46: Rotations not being inherited from matchpack for chips/solderjumpers
I think SJ1 isn't properly being rotated according to the matchpack algorithm

image

_Originally posted by @seveibar in https://github.com/tscircuit/core/issues/1225#issuecomment-3239367490_

ShiboSoftwareDev:
this is the `outputLayout`:
```{
  chipPlacements: {
    U1: {
      x: 0,
      y: 0,
      ccwRotationDegrees: 0,
    },
    SJ1: {
      x: 0,
      y: -1.2,
      ccwRotationDegrees: 0,
    },
    R1: {
      x: 1.7000000000000002,
      y: -1.1047276750000001,
      ccwRotationDegrees: 0,
    },
    R2: {
      x: 0.34999999999999964,
      y: -2.3047276749999996,
      ccwRotationDegrees: 0,
    },
    C2: {
      x: -0.5999999999999988,
      y: 1.165,
      ccwRotationDegrees: 0,
    },
  },
  groupPlacements: {},
}
```
/claim #1226
/fixes #1226


ShiboSoftwareDev:
@seveibar the algo returns 0 rotation for all components, but that doesn't matter since the user's preference will override it right?

seveibar:
@ShiboSoftwareDev it's confusing but the algorithm is accepting allowed _offsets_ from the user's given rotation. So this fix doesn't look correct, the original code was correct b/c when a user specifies schRotation we basically say "do not offset it, the only allowed offset rotation is 0"

PR #1258, opened 2025-09-02
fix: orientation of non power/gnd net label

### LINKED ISSUE #1257: Nets that are not power nets (VCC/GND etc.) should prioritize horizontal net label orientation over vertical orientation
i think one of the most recent repros has a snapshot svg that shows this issue

Issue involves the InputProblem passed to the schematic-trace-solver

Image

/bounty $75

imrishabh18:
/close #1257

@seveibar Is this the correct approach? Or what is some better way to find out the net which are power or ground?


seveibar:
Its a property of the source_net, ideally we could do Net.isGroundNet and net.isPowerNet etc

seveibar:
This approach isnt correct, you should be changing the InputProblem

PR #1260, opened 2025-09-02
Added repro of same net not combine circuit

AnasSarkiz:
this for https://github.com/tscircuit/schematic-trace-solver/issues/29


AnasSarkiz:
@seveibar, I need to write out the InputProblem. I forgot how

seveibar:
@AnasSarkiz you have to set DEBUG=Group_doInitialSchematicTraceRender (or something like that)

Issue #1263, opened 2025-09-03
Don't create components if there's a same-name naming conflict, create circuit json error failed_to_create


seveibar:
/bounty $50

Make sure to reproduce in test

PR #1269, opened 2025-09-03
Fix: Prevent component creation on same-name conflicts

t431w:
Ref #1263 
claim #1263


seveibar:
e.g. rename from "obstacle" to "obstacle1" and "obstacle2"

PR #1270, opened 2025-09-03
fix: Support group rotation to  manual pcb trace Paths

Abse2001:
/claim https://github.com/tscircuit/tscircuit/issues/752
/closes https://github.com/tscircuit/tscircuit/issues/752

Before 
image

After 
image


lib/components/primitive-components/Trace/Trace_doInitialPcbManualTraceRender.ts
```diff
@@ -70,11 +70,21 @@ export function Trace_doInitialPcbManualTraceRender(trace: Trace) {
     layer: layer as LayerRef,
     start_pcb_port_id: anchorPort.pcb_port_id!,
   })
+  // Get group rotation in radians
+  const group = trace.getGroup()
+  const rotation = (group?._parsedProps.pcbRotation || 0) * (Math.PI / 180)
```
seveibar, line 75: what happens if a group is rotated inside another group that has rotation? This will fail.

You can use the `group._getGlobalTransform()` to get a transformation-matrix that can be applied to all points (I may have gotten the name slightly wrong)

Use the `transformation-matrix` library to applied the rotation to each point. It should have a function that handles this.

This will also handle translations etc.

tests/components/primitive-components/trace-with-pcb-path.test.tsx
```diff
@@ -32,7 +32,7 @@ test("trace with manual pcbPath", async () => {
   const routeCoords = pcbTraces[0].route.map((p) => ({ x: p.x, y: p.y }))
   expect(routeCoords).toEqual([
     { x: -1.5, y: 0 },
-    { x: -0.5, y: 0 },
+    { x: -1, y: 0 },
```
seveibar, line 35: i sort of doubt this is supposed to change (double check)

seveibar (approved):

PR #1275, opened 2025-09-04
feat: Don't create components if there's a same-name naming conflict,create circuit json error failed_to_create

### LINKED ISSUE #1263: Don't create components if there's a same-name naming conflict, create circuit json error failed_to_create

MustafaMulla29:
/closes #1263 
/claim #1263


lib/components/base-components/PrimitiveComponent/PrimitiveComponent.ts
```diff
@@ -583,6 +613,65 @@ export abstract class PrimitiveComponent<
     }
   }
 
+  /**
+   * Automatically check for naming conflicts after source component creation.
+   * This method is called automatically after doInitialSourceRender.
+   */
+  _checkNamingConflictAfterSourceCreation() {
```
seveibar, line 620: name doesn't imply that it has an effect

seveibar:
This is so strange and non-obvious in how it works i don't know how to recommend changing it, the approach doesn't make sense at all (why would setting a source component id do a name check)

PR #1284, opened 2025-09-05
feat(schematic): Add power direction arrows for pinAttributes

ShiboSoftwareDev:
This PR introduces the ability to display power direction arrows on schematic ports based on the pinAttributes property 
of a component.                                                                                                                  

 • If a port's pinAttributes include requiresPower: true, an input arrow is rendered.                                            
 • If providesPower: true is set, an output arrow is rendered.                                                                   
 • If both are set, the port is treated as bidirectional and both arrows are shown.                                              

This feature improves schematic clarity by visually indicating the intended direction of power flow for pins. A new snapshot test
has been added for a chip component to verify that input, output, and bidirectional arrows render correctly on all sides.


lib/components/primitive-components/Port/Port.ts
```diff
@@ -404,7 +404,8 @@ export class Port extends PrimitiveComponent<typeof portProps> {
       bestDisplayPinLabel = labelHints[0]
     }
 
-    const schematic_port = db.schematic_port.insert({
+    const pinAttributes = (this.parent as any)?.props?.pinAttributes
```
seveibar, line 407:
```suggestion
    const pinAttributes = (this.parent as any)?._parsedProps?.pinAttributes
```

lib/components/primitive-components/Port/Port.ts
```diff
@@ -404,7 +404,8 @@ export class Port extends PrimitiveComponent<typeof portProps> {
...
+    const schematicPortInsertProps: any = {
```
seveibar, line 408:
```suggestion
    const schematicPortInsertProps: SchematicPort = {
```

idk why people are using any so much, it degrades the codebase

seveibar (commented):
some notes on circuit-to-svg and the snapshot

image

PR #1285, opened 2025-09-05
Reproduce + Fix backwards net label issue in test

seveibar:


seveibar:
i think the regressions are too serious to merge this atm

PR #1287, opened 2025-09-06
feat: Don't create components if there's a same-name naming conflict

### LINKED ISSUE #1263: Don't create components if there's a same-name naming conflict, create circuit json error failed_to_create

MustafaMulla29:
/closes #1263
/claim #1263


lib/components/base-components/Renderable.ts
```diff
@@ -18,6 +18,7 @@ export const orderedRenderPhases = [
   "CreateTraceHintsFromProps",
   "SourceGroupRender",
   "AssignNameToUnnamedComponents",
+  "SourceNameValidation",
```
imrishabh18, line 21: If this renderPhase is just for explicitly named components, and not auto assigned names. I think it's better you do it before the `AssignNameToUnnamedComponents` phase

lib/components/base-components/Renderable.ts
```diff
@@ -281,6 +282,9 @@ export abstract class Renderable implements IRenderable {
 
   runRenderPhaseForChildren(phase: RenderPhase): void {
     for (const child of this.children) {
+      // Skip children if this component is marked for removal
+      if (this.shouldBeRemoved) continue
+
```
imrishabh18, line 287: Not sure if this is a correct pattern, why would you remove the children?

cc @seveibar

MustafaMulla29: @imrishabh18 the children of that component (like ports) still try to render normally, and many of them expect their parent to have a `source_component_id`. Then Reach `SourceParentAttachment` phase and it will crash if we dont mark the children as well.

Is this correct?

lib/components/base-components/Renderable.ts
seveibar: this should be immediately after ReactSubtreesRender i think, and it's not `SourceNameValidation` i would call it `SourceNameDuplicateComponentRemoval`

seveibar (commented):
I don't mind this approach that much but it should be renamed and executed earlier

lib/components/base-components/NormalComponent/NormalComponent.ts
```diff
@@ -150,6 +150,62 @@ export class NormalComponent<
...
+      if (component.children) {
+        for (const child of component.children) {
+          matches.push(...findEarlierComponentsWithSameName(child, targetName))
+        }
+      }
+
+      return matches
+    }
+
+    const conflictingComponents = findEarlierComponentsWithSameName(
```
seveibar, line 186: this feels a bit hacky to me, you should just be able to go `this.getSubcircuit(`.${name}`)` and see if multiple components are returned. If multiple components are returned, then there's an issue. You will still need to check if the component was initialized first, as you're already doing

tests/components/normal-components/name-conflict-error.test.tsx
```diff
@@ -0,0 +1,106 @@
...
+  circuit.add(
+    <board width="10mm" height="10mm">
+      <group name="Group1">
+        <resistor name="R3" resistance="1k" footprint="0402" />
+      </group>
+      <group name="Group2">
+        <resistor name="R4" resistance="2k" footprint="0402" />
+      </group>
+    </board>,
+  )
```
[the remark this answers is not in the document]
seveibar: there are no duplicate names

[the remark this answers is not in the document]
MustafaMulla29: @seveibar that test was checking there is a error in the db. Should i remove the entire test file?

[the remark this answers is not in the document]
MustafaMulla29: Yea sorry, I miss read it😅

tests/components/normal-components/name-conflict-error.test.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+  expect(sourceComponents).toHaveLength(1)
+  expect(sourceComponents[0]).toMatchObject({
+    name: "R1",
+    type: "source_component",
+  })
+})
+
+test("should allow components with different names", async () => {
```
seveibar, line 43: remove these useless tests!!!

also one test per file

seveibar (approved):

PR #1289, opened 2025-09-06
support `footprintLibraryMap`

imrishabh18:


lib/components/base-components/NormalComponent/NormalComponent.ts
```diff
@@ -375,6 +376,22 @@ export class NormalComponent<
...
+    s: string,
+  ): { lib: string; footprintName: string } | null {
+    if (this._isFootprintUrl(s)) return null
+    const idx = s.indexOf(":")
+    if (idx <= 0) return null
+    const lib = s.slice(0, idx)
+    const footprintName = s.slice(idx + 1)
+    if (!lib || !footprintName) return null
+    return { lib, footprintName }
+  }
```
seveibar, line 393: this doesn't need to be a class method, similar to `isFootprintUrl`, we should generally avoid class methods which are hard to refactor. You can just move this to a file, even in the NormalComponent directory is fine. `parseLibraryFootprintRef` and `isFootprintUrl`. It will help reduce the size of this file

lib/components/base-components/NormalComponent/NormalComponent.ts
```diff
@@ -375,6 +376,22 @@ export class NormalComponent<
...
+  _parseLibraryFootprintRef(
+    s: string,
+  ): { lib: string; footprintName: string } | null {
+    if (this._isFootprintUrl(s)) return null
+    const idx = s.indexOf(":")
+    if (idx <= 0) return null
+    const lib = s.slice(0, idx)
+    const footprintName = s.slice(idx + 1)
+    if (!lib || !footprintName) return null
+    return { lib, footprintName }
```
seveibar, line 392: i think this breaks url handling ("http" or "https" would be the lib), i would also qualify lib because it's a better variable name and also done for footprint
 
```suggestion
    return { footprintLib, footprintName }
```

lib/components/base-components/NormalComponent/NormalComponent.ts
```diff
@@ -755,6 +773,72 @@ export class NormalComponent<
...
+          )
+          this.addAll(fpComponents)
+          this._markDirty("InitializePortsFromChildren")
+        }
+
+        // Find a default resolver in the library map
+        let resolverFn: ((path: string) => Promise<any>) | undefined
+        if (libMap && typeof libMap === "object") {
+          // Prefer common keys for wildcard/default resolvers
+          const preferredKeys = ["*", "resolve", "default", ""]
```
seveibar, line 808: i don't know what this is at all

tests/footprint/footprint-library-map1.test.tsx
```diff
@@ -0,0 +1,40 @@
...
+test("footprint library map 1", async () => {
+  const { url: footprintServerUrl } = getTestFootprintServer(
+    external0402Footprint,
+  )
+  const { circuit } = getTestFixture({
+    platform: {
+      footprintLibraryMap: {
+        kicad: {
+          // Default resolvewr for any KiCad path
+          resolve: async (path: string) => {
```
seveibar, line 15: what is this? this isn't right

seveibar: why resolve? it should just be a function

seveibar: I think you re-implemented the interface for an async function or something

lib/components/base-components/NormalComponent/footprintUtils.ts
```diff
@@ -0,0 +1,20 @@
+/**
```
seveibar, line 1: just put into their own files, i wish i could have a linting rule but these files that put together multiple utils cause tons of problems with searching (you can actually see AI take longer whenever they're used because it does a file search then a grep)

This should be two files: `isFootprintUrl.ts` and `parseLibraryFootprintRef.ts`

PR #1292, opened 2025-09-07
Fix trace rendered (twice) when only manual routing is expected

techmannih:
fix https://github.com/tscircuit/tscircuit/issues/759
/claim https://github.com/tscircuit/tscircuit/issues/759


seveibar (commented):
this is definitely the right idea, but it probably fails in some cases

please create a circuit with two subcircuits and a board that share a net:

```tsx
<board>
<subcircuit name="S1">
<resistor connections={{ pin1: "net.GND" }}  pcbX={2} name="R1" footprint="0402" />
</subcircuit>
<subcircuit name="S2">
<resistor connections={{ pin1: "net.GND" }}  pcbX={2} name="R2" footprint="0402" />
</subcircuit>
</board>


```

seveibar (changes requested):
This is a hack, if you add a second resistor connected to GND inside one of the subcircuits your hack will fail

seveibar (approved):

PR #1330, opened 2025-09-11
feat: Add <copperpour> Component for PCB Layout

ShiboSoftwareDev:
This pull request implements the new <copperpour> primitive component.                                                                                      

The component automatically generates a large copper area on a specified layer, connecting to a designated net. It intelligently 
creates clearances around all other PCB elements (pads, holes, traces) that are not part of the same electrical net.             

 • Net Connectivity: Connects to a net using the connectsTo prop (e.g., <copperpour connectsTo="net.GND" />).                    
 • Layer Specificity: Pours are created on a single PCB layer specified by the layer prop.                                       
 • Automatic Obstacle Avoidance: The pour flows around existing pads, holes, and traces, maintaining a safe distance.            
 • Configurable Clearances:                                                                                                      
    • padMargin: Controls the clearance around plated hole pads.                                                                 
    • traceMargin: Controls the clearance around traces and SMT pads.                                                            
 • Board Outline Confinement: The pour is automatically constrained to the shape of the board's outline.                         

The component uses @flatten-js/core to perform geometric calculations. It identifies all obstacles on the target layer, performs 
boolean subtraction for rectangular obstacles, and then programmatically adds circular inner-rings for holes. The final geometry 
is stored as a brep shape in the circuit JSON.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@tscircuit/core",
   "type": "module",
-  "version": "0.0.732",
+  "version": "0.0.733",
   "types": "dist/index.d.ts",
   "main": "dist/index.js",
   "module": "dist/index.js",


lib/components/base-components/Renderable.ts
```diff
@@ -297,9 +299,26 @@ export abstract class Renderable implements IRenderable {
...
+        }
+      }
+    }
     const deps = asyncPhaseDependencies[phase] || []
     for (const depPhase of deps) {
-      if (this._hasIncompleteAsyncEffectsInSubtreeForPhase(depPhase)) return
+      if (
+        asyncCheckContext._hasIncompleteAsyncEffectsInSubtreeForPhase(depPhase)
+      )
+        return
```
seveibar, line 321: not sure if this is still in active development, but i think we should revert the changes to this import-to-keep-simple runRenderPhase function

ShiboSoftwareDev:
@seveibar I've simplified the logic, but I couldn't find another way to make copper pour wait for the autorouter
image

seveibar (commented):
Gotcha ill take a look to see if i can help, asynCheckContext is super weird naming, i dont understand that code or how it generalizes to the render process

PR #1361, opened 2025-09-15
Support `symbols` as an alias for `symbolName`

imrishabh18:
Ref - https://github.com/tscircuit/tscircuit/issues/785


seveibar:
@imrishabh18 this isnt what you were showing me earlier right? This mixes string and react components iiuc

imrishabh18:
> @imrishabh18 this isnt what you were showing me earlier right? This mixes string and react components iiuc

No, this is just that the `symbol` when passed the symbolName also works

PR #1367, opened 2025-09-16
feat: Add support for includeInBoardPinout

ShiboSoftwareDev:
This pull request adds support for the includeInBoardPinout property within pinAttributes.                                                                         

When a port on a component has the includeInBoardPinout: true attribute, the corresponding pcb_port in the circuit output will have its is_board_pinout flag set to true.                                                                                                                                                              

A new test file has been added to verify this functionality.

RE https://github.com/tscircuit/tscircuit/issues/788


ShiboSoftwareDev:
@seveibar please review the changes to hole and platedhole

lib/components/primitive-components/PlatedHole.ts
```diff
@@ -165,6 +165,9 @@ export class PlatedHole extends PrimitiveComponent<typeof platedHoleProps> {
         rect_ccw_rotation: props.pcbRotation ?? 0,
         rect_pad_width: props.outerWidth,
         rect_pad_height: props.outerHeight,
+        PcbHoleRotatedPillWithRectPad: false,
```
seveibar, line 168:
```suggestion
```

seveibar (approved):
see comment

lib/components/primitive-components/PlatedHole.ts
ShiboSoftwareDev: nervermind, just had to update deps

PR #1372, opened 2025-09-17
fix: standardize board anchor alignment behavior

### LINKED ISSUE #1371: Implement `boardAnchorPosition` and `boardAnchorAlignment` (which specifies what XY coordinates a board will appear and grow into for auto-sizing)
/bounty $100

AnasSarkiz:
- Fixed inconsistencies in board anchor alignment calculations
- Updated tests to verify all anchor positions work as expected

/fixes #1371 
/lcaim #1371

### DIFF
--- a/lib/components/normal-components/Board.ts
+++ b/lib/components/normal-components/Board.ts
@@ -108,8 +108,8 @@ export class Board extends Group<typeof boardProps> {
     const { db } = this.root!
     const { _parsedProps: props } = this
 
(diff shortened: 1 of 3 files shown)


seveibar (changes requested):
one test per file, snapshot test (maybe with some kind of origin indication with silkscreentext or smtpads, your test cases are invalid because they use outlineOffsetX which is incompatible with boardAnchorPosition (in fact it should error)

PR #1374, opened 2025-09-17

nailoo:
/claim #1371


seveibar:
btw looks really good! I think this is the right approach

seveibar (commented):
the snapshots are not quite obvious, maybe say `"top_left(X,Y)"` with the specific coordinate, then always place some fabrication note text at (0,0) with the text (0,0). I'm having trouble understanding your snapshots.

You can also rename the resistors like `R_X_Y` if they are placed explicitly at a position.

Just by looking at a snapshot i should be able to tell if the behavior is correct, it will make the project easier to maintain when snapshots change

Make sure to also have tests for bottom_right etc. (it's hard to tell from the snapshots if we've tested anything except top_left)

PR #1385, opened 2025-09-17

nailoo:


seveibar (changes requested):
The updated snapshots demonstrate incorrect behavior eg top left is not on the top left and bottom right is not on the bottom right

lib/components/normal-components/Board.ts
```diff
@@ -67,6 +69,38 @@ export class Board extends Group<typeof boardProps> {
   _drcChecksComplete = false
   _connectedSchematicPortPairs = new Set<string>()
 
+  private _applyBoardAnchorAlignment(
```
seveibar, line 72: two parameter rule:  https://github.com/tscircuit/handbook/blob/main/guides/code.md

also the name of this function is wrong, it doesn't have any effects (it doesn't mutate anything) so it shouldn't use the word `apply`, i also don't understand what it's returning but it's not a board alignment. I think this function could be called `computeBoardAlignmentOffset` and maybe moved out of this file (it's important to keep these class files very small), you can turn `Board.ts` into `Board/Board.ts` and have `Board/computeBoardCenterWithAlignment.ts`

lib/components/normal-components/Board.ts
```diff
@@ -67,6 +69,38 @@ export class Board extends Group<typeof boardProps> {
...
+    width: number,
+    height: number,
+  ) {
+    const {
+      boardAnchorAlignment,
+      boardAnchorPosition,
+    }: {
+      boardAnchorAlignment?: NinePointAnchor
+      boardAnchorPosition?: { x: number; y: number }
+    } = this._parsedProps as any
```
seveibar, line 83: just update props, try to avoid `as any`

seveibar (commented):
Ok you need some tests that have `pcbPack`, you're moving the board but I don't think automatically-placed components will be in the correct place (and it may change your approach)

The tests are readable now thank you!!!

PR #1394, opened 2025-09-20

### LINKED ISSUE #1371: Implement `boardAnchorPosition` and `boardAnchorAlignment` (which specifies what XY coordinates a board will appear and grow into for auto-sizing)

nailoo:
fix #1371 
/claim #1371


seveibar (commented):
Should R1 be in the middle in the pcbPack test? Also like i mentioned, we should make a Board directory and move the functions into files with the same name to reduce the size of Board.ts

im worried that this implementation is wrong and we need to make sure the board position before layout is adjusted

Issue #1348, opened 2025-09-14
Support a `<subcircuit />`/`<group />` having a single schematic box representation


seveibar:
We want to enable subcircuits to have a single schematic box representation, this way you can have e.g. a group of pinheaders representing a single schematic box for an arduino shield.

- [x] Propose any necessary props to tscircuit/props
- [x] Introduce any changes required to circuit-json, perhaps something like `schematic_component.is_schematic_group`   
- [ ] Implement in core

As discussed here: https://github.com/tscircuit/common/pull/9#pullrequestreview-3221726590

Theoretical example:

```tsx
<group showAsSchematicBox
connections={{
  D1: "J1.pin1",
  D2: "J2.pin1",
  // ...
}}
schPinArrangement={{
  leftSide: { direction: "top-to-bottom", pins: ["D1", "D2" ] }
}}>
  <pinheader name="J1" pinCount={3}  />
  <pinheader name="J2" pinCount={3}  />
</group>
```

/bounty $400

seveibar:
CC @Abse2001

rishi-jat:
@seveibar is this issue is still open can I work on this issue ?

PR #1407, opened 2025-09-25
Add Group Schematic Box Rendering Support

Abse2001:
## Summary
- Allow `<group showAsSchematicBox>` to render descendants as a single schematic box with generated interface pins.
- Add `GroupInterfacePort`, owner lookups, and `is_schematic_group` support in circuit JSON typing.
- Tighten board/port config handling

/claim #1348 

image

image

### DIFF
--- a/lib/components/primitive-components/Trace/Trace__findConnectedPorts.ts
+++ b/lib/components/primitive-components/Trace/Trace__findConnectedPorts.ts
@@ -67,21 +67,6 @@ export function Trace__findConnectedPorts(trace: Trace):
         })
       }
 
-      // Try to use the component's selectOne method for port resolution first
-      // This handles cases like Groups that implement custom port resolution
-      const resolvedPort = targetComponent.selectOne(`.${portToken}`, {
-        port: true,
-      })
-      if (resolvedPort && resolvedPort.componentName === "Port") {
-        portsWithSelectors[
-          portsWithSelectors.findIndex((p) => p.selector === selector)
-        ] = {
-          selector,
-          port: resolvedPort as Port,
-        }
-        continue
-      }
-
       const ports = targetComponent.children.filter(
         (c) => c.componentName === "Port",
       ) as Port[]
--- a/lib/utils/resolvePortFromSelector.ts
+++ b/lib/utils/resolvePortFromSelector.ts
@@ -32,6 +32,16 @@ export const resolvePortFromSelector = (
   }
   if (!parentComponent) return null
 
+  // Try the component's custom selectOne method first
+  // This handles cases like Groups that implement custom port resolution
+  const customResolved = parentComponent.selectOne(`.${portToken}`, {
+    port: true,
+  })
+  if (customResolved && customResolved.componentName === "Port") {
+    return customResolved as Port
+  }
+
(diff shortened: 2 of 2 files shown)


lib/components/primitive-components/Group/Group.ts
```diff
@@ -744,6 +847,111 @@ export class Group<Props extends z.ZodType<any, any, any> = typeof groupProps>
         })
       }
     }
+
+    if (this._isShowAsSchematicBoxEnabled()) {
+      this._renderShowAsSchematicBoxComponent()
+    }
+  }
+
+  private _renderShowAsSchematicBoxComponent() {
```
seveibar, line 856:
```suggestion
  private _renderAsSchematicBoxComponent() {
```

lib/components/primitive-components/Group/Group.ts
```diff
@@ -56,6 +68,91 @@ export class Group<Props extends z.ZodType<any, any, any> = typeof groupProps>
...
+    }
+    if (!parentComponent) return null
+
+    const ports = parentComponent.children.filter(
+      (c) => c.componentName === "Port",
+    ) as Port[]
+
+    return (
+      ports.find((port) => port.getNameAndAliases().includes(portToken)) ?? null
+    )
```
seveibar, line 108: it feels like this code is copied from somewhere and it's dangerous to repeat it since if there's a bug it won't be fixed in both places

You can either call the method of the implementing class i.e. `Port._resolvePortFromSelector(this, ...)` or you can extend a class (usually better)

lib/components/primitive-components/Group/GroupInterfacePort.ts
```diff
@@ -0,0 +1,42 @@
...
+    const primaryTarget = this._targetPorts[0]
+    if (!primaryTarget?.source_port_id) return
+    this.source_port_id = primaryTarget.source_port_id
+    this.source_component_id = primaryTarget.source_component_id
+    this.subcircuit_connectivity_map_key =
+      primaryTarget.subcircuit_connectivity_map_key ?? null
+  }
+
+  doInitialSourceParentAttachment(): void {
+    // The underlying target port already has the correct parent attachment.
```
seveibar, line 36: ...then why are you overriding this method? by overriding this method you are removing the implementation for source parent attachment

lib/components/primitive-components/Group/GroupInterfacePort.ts
```diff
@@ -0,0 +1,42 @@
...
+    return this._targetSelectors
+  }
+
+  doInitialSourceRender(): void {
+    const primaryTarget = this._targetPorts[0]
+    if (!primaryTarget?.source_port_id) return
+    this.source_port_id = primaryTarget.source_port_id
+    this.source_component_id = primaryTarget.source_component_id
+    this.subcircuit_connectivity_map_key =
+      primaryTarget.subcircuit_connectivity_map_key ?? null
```
seveibar, line 32: this isn't right, you're not creating source ports

seveibar (commented):
some comments, this is the right track

seveibar:
I THINK the reason why you keep needing to modify `selectOne` is because you're not adding ports as children to the group

PR #1418, opened 2025-09-29
Handle errors inside the parts engine

### LINKED ISSUE #1400: Handle errors inside the parts engine gracefully
Image

MustafaMulla29:
/claim #1400
/closes #1400


seveibar (changes requested):
Don't silence errors- you have to show them to the user via circuit json errors

Also you probably want to return structured errors from the parts-engine so that we have something readable for the user

PR #1426, opened 2025-10-02
Fix invalid junction circles for same-network traces

Omar8345:
## Summary

- **Fixed invalid junctions** for same-network traces by editing `compute-junctions.ts` and related schematic trace logic. 
- **Updated** **`repro48`** snapshot and other related test snapshots that may be affected.

## Visual Reference

repro48-555-timer-schematic snap

/claim #1245


Omar8345:
@seveibar hope you check this, I thought of updating `repro48` right now until I get further instructions regarding the other failing tests, since in my opinion, I don't find `repro48` the only affected test

Omar8345:
@seveibar is like 12 snapshots a lot or not because without those tests wont pass

seveibar:
This removes valid junctions, it doesnt fix invalid junctions

PR #1429, opened 2025-10-02
Port hierarchy traversal to find parent chip PCB component

### LINKED ISSUE #1409: SchematicRect with traces test fails - Port PCB rendering issue with custom symbols
# Issue Description

The test SchematicRect with traces in `schematic-rect-with-traces.test.tsx` is failing with the error:

```
error: <port#69664 name=".pin1" /> has no parent pcb component, cannot render pcb_port (parent: <symbol#69665 />)
```

## Root Cause

The test is trying to create chips with both custom `symbol` definitions and `footprint` properties. The ports defined inside the custom symbol don't have access to the PCB component created by the footprint, causing the PCB port rendering to fail.

## Related Code Snippet

```tsx
      <chip
        name="U1"
        symbol={
          <symbol>
            <schematicrect
              center={{ x: 0, y: 0 }}
              width={2}
              height={2}
              isFilled={false}
            />
            <port name="pin1" direction="right" schX={1} schY={0} />
          </symbol>
        }
      />
```

## Expected Behavior

The test should be able to create chips with custom symbols that have footprints, and the ports should be able to render on both schematic and PCB sides.

saish9901:
/claim #1409
/closes #1409

the problem was that When you put a <port> inside a custom <symbol>, the port couldn't find its parent chip's PCB component.

What i fixed is that Instead of only checking the immediate parent, the port now climbs up the family tree until it finds an ancestor with pcb_component_id (the Chip).

Please review @seveibar


tests/components/primitive-components/schematic-rect-with-traces.test.tsx
imrishabh18, line 1: The test is failing, make sure the tests pass before requesting a review!

imrishabh18 (changes requested):

PR #1441, opened 2025-10-04
spicey integration

seveibar:
- **init spicey integration**
- **bake in spicey**


lib/spice/get-spicey-engine.ts
```diff
@@ -0,0 +1,37 @@
+import type { SpiceEngine } from "@tscircuit/props"
+import type { SimulationExperiment } from "circuit-json"
+
+export const getSpiceyEngine = (): SpiceEngine => {
+  return {
+    async simulate(spiceString: string) {
+      const { simulate, spiceyTranToVGraphs } = await import("spicey")
```
seveibar, line 7: need to fix this

PR #1444, opened 2025-10-04
boost converter circuit for spice simulation

ShiboSoftwareDev:
This should help track bugs and progress


seveibar (approved):
good example of "too many circles" in the graph making it more difficult to read too imo

PR #1445, opened 2025-10-04
feat: Add spice analysis test for RC charging voltage divider

ShiboSoftwareDev:
This PR adds a new test case for spice analysis featuring an RC charging       
circuit with a voltage divider.                                                

The new test file, spice-analysis03-rc-charging-voltage-divider.test.tsx,      
defines a circuit containing resistors, a capacitor, and a chip component      
acting as a power source. It runs an analogsimulation and asserts that the     
simulation output is generated, capturing the result in a snapshot.


ShiboSoftwareDev:
This is the first circuit I used to test spice sims

seveibar (approved):
yes, just as a reminder you can improve the schematic representation with `maxSchTraceLength` on the board, at least until https://github.com/tscircuit/schematic-trace-solver/issues/68 is solved

PR #1454, opened 2025-10-05
feat: added board title

jayantpranjal0:
ref https://github.com/tscircuit/tscircuit/issues/895

Need updation in package.json for tscircuit/props module


seveibar:
@jayantpranjal0 you need to add title support to circuit json in order to use this in a pinout diagram

seveibar:
Board already has a name, it doesn't need a title. I'm PR'ing to remove the title from props

seveibar:
Ok never mind, it's the difference between a display name and a variable name 👍

PR #1456, opened 2025-10-05
feat(spice): Add duration and timePerStep support to <analogsimulatio…

ShiboSoftwareDev:
…n />

This pull request adds support for duration and timePerStep props to the       
<analogsimulation /> component. This allows for customization of the SPICE     
transient analysis parameters.                                                 

 • <analogsimulation /> now accepts optional duration and timePerStep props.   
 • The SPICE simulation logic now uses these props to generate the .tran       
   directive. If not provided, it defaults to a 10ms duration and 1us time     
   step.                                                                       
 • A new test, spice-analysis04-rc-charging-voltage-divider-custom-time.test.tsx, has been added to verify that custom simulation times are correctly applied.


lib/components/primitive-components/Group/Group_doInitialSimulationSpiceEngineRender.ts
```diff
@@ -29,8 +29,20 @@ export function Group_doInitialSimulationSpiceEngineRender(group: Group<any>) {
...
+    const simulationProps = analogSims[0]?._parsedProps
+    const duration = simulationProps?.duration ?? "10ms"
+    const timePerStep = simulationProps?.timePerStep ?? "1us"
+
+    let spiceNetlistStr = circuitJsonToSpice(circuitJson).toSpiceString()
+
+    if (!spiceNetlistStr.includes(".tran")) {
+      spiceNetlistStr = spiceNetlistStr.replace(
+        /\.END/i,
+        `.tran ${timePerStep} ${duration}\n.END`,
```
seveibar, line 42: this seems hacky to me, why are we replacing? circuitJsonToSpice should be handling this

lib/components/primitive-components/AnalogSimulation.ts
```diff
@@ -1,5 +1,11 @@
-import { analogSimulationProps } from "@tscircuit/props"
+import { analogSimulationProps as baseAnalogSimulationProps } from "@tscircuit/props"
 import { PrimitiveComponent } from "../base-components/PrimitiveComponent"
+import { z } from "zod"
+
+const analogSimulationProps = baseAnalogSimulationProps.extend({
```
seveibar, line 5: why are props being define in this file

seveibar (changes requested):
circuitJsonToSpice needs to be able to handle this- we may need to pass in the experiment id etc. which would be generated by/contained within AnalogSimulation

seveibar (approved):

PR #1458, opened 2025-10-06
Fix invalid junctions between unconnected nets using connectivity mapping

### LINKED ISSUE #1245: Invalid Junction (small green circle indicating two traces cross) in repro48

saish9901:
/claim #1245 
/closes #1245

Junctions were incorrectly appearing where traces crossed but weren't actually connected to the same electrical net. The underlying issue was that the code was only checking if traces had the same source ID, rather than verifying if they were electrically connected to the same circuit.

So, Added proper net connectivity checking using the subcircuit connectivity map key to determine electrical relationships between traces. Now the system only creates junctions between traces that are actually part of the same electrical circuit, eliminating false junction dots while preserving legitimate connections.

Please review @seveibar


saish9901:
> Could you please add updated snapshots???


1.  voltage regulator circuit example7-
image

the voltage regulator circuit shows the correctly connected traces, the junctions are legitimate in this circuit, so my fx actually is preserving all the valid junctions.



2. repro4-schematic-trace-overlap
image

the repro4-schematic-trace-overlap validates that crossings are handled correctly

The issue is that the schematic auto-router is struggling with circuits that have crossing traces.
The existing test snapshots look identical because most test circuits are simple enough that the auto-router avoids crossings entirely. The invalid junction bug occurs in more complex.

Can u possibly test the changes with the attached image - this is the same circuit mentioned in the problem.
image


Please let me know if any changes to be made.

@seveibar @techmannih

seveibar (changes requested):
no evidence of fix in previous snapshots (there must be evidence of bug + fix)

Issue #1467, opened 2025-10-09
Undefined behaviour for unnamed chips (pun intended)


Heinrich-XIAO:
When making a chip without a name, it randomly gets named either `unnamed_chip1` or `undefined`.

Repro: weird caching stuff makes it hard to do it, but incognito fixes most of this. (some browsers will persist cached data if the incognito instances are not all closed)

Heinrich-XIAO:
It's not that it shouldn't be named "unnamed. It's that it should be named in a defined way. Right now, it randomly chooses either "unnamed_chip" or "undefined_chip".

PR #1480, opened 2025-10-12
repro: unnamed chips receive sequential unnamed_chip fallback names

devroy10:
repro for #1467


Heinrich-XIAO:
I think I made a typo in the original PR. It's not "undefined" and "unnamed_chip". Its "undefined_chip" or "unnamed_chip".

Heinrich-XIAO:
But also, now I can't seem to get it to do anything other than "undefined_chip".

PR #1510, opened 2025-10-15
feat: enhance CadModel URL handling and add tests for relative paths

ArnavK-09:
- Updated CadModel to use constructAssetUrl for better URL management.
- Improved cache-busting logic for model URLs.
- Added tests to verify CadModel handles relative paths correctly.

### DIFF
--- a/lib/utils/constructAssetUrl.ts
+++ b/lib/utils/constructAssetUrl.ts
@@ -1,3 +1,9 @@
+const joinUrlPath = (base: string, path: string): string => {
+  const trimmedBase = base.replace(/\/+$/, "")
+  const trimmedPath = path.replace(/^\/+/, "")
+  return `${trimmedBase}/${trimmedPath}`
+}
+
 export const constructAssetUrl = (targetUrl: string, baseUrl?: string) => {
   if (!baseUrl) {
     return targetUrl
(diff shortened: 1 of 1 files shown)


lib/utils/constructAssetUrl.ts
```diff
@@ -8,8 +8,22 @@ export const constructAssetUrl = (targetUrl: string, baseUrl?: string) => {
   }
 
   try {
-    const base = new URL(baseUrl)
-    const resolved = new URL(targetUrl, base.origin)
+    const baseUrlObj = new URL(baseUrl)
+
+    // If the targetUrl already starts with the base URL path, treat it as absolute from domain root
```
seveibar:
```suggestion
    // If the targetUrl already starts with the base URL path, treat it as absolute from domain root
    // e.g. if....
```

tests/components/normal-components/chip-cad-model-glb-relative-path.test.tsx
```diff
@@ -0,0 +1,62 @@
...
+    "http://localhost:3020/api/files/static/soic8.glb",
+  )
+  const cadComponent2 = circuit.db.cad_component.list()[1]
+  expect(cadComponent2.model_glb_url).toBe(
+    "http://localhost:3020/api/files/static/soic8.glb",
+  )
+  const cadComponent3 = circuit.db.cad_component.list()[2]
+  expect(cadComponent3.model_glb_url).toBe(
+    "http://localhost:3020/api/files/static/glbs/soic8.glb",
+  )
```
seveibar: this test looks correct to me!

lib/utils/constructAssetUrl.ts
```diff
@@ -8,8 +8,22 @@ export const constructAssetUrl = (targetUrl: string, baseUrl?: string) => {
...
+    ) {
+      const resolved = new URL(targetUrl, baseUrlObj.origin)
+      return resolved.toString()
+    }
+
+    // Otherwise, append to the base URL path
+    const baseUrlWithTrailingSlash = baseUrl.endsWith("/")
+      ? baseUrl
+      : baseUrl + "/"
+    const resolved = new URL(targetUrl.substring(1), baseUrlWithTrailingSlash)
```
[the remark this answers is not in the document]
seveibar: alternatively if you're going to normalize the url `baseUrlWithTrailingSlash` just do it at the beginning, not in the middle where logic is

PR #1514, opened 2025-10-16
fix: board width height when outline is passed

imrishabh18:
`doInitialPcbBoardAutoSize()` method was overwriting these calculated values with auto-calculated dimensions based on component placement


seveibar (approved):
Its a bit ambiguous if width/height are also present if outline is present. The width and height become the bounds. I’d rather have width/height be undefined if an outline is present i think but we could add a “shape” property to board in the future to help clear up ambiguity

PR #1524, opened 2025-10-17
fix: maintain accurate CAD rotation alignment with PCB components during packing and layer flips

### LINKED ISSUE #1518: Cad Models aren't being rotated after packing (only pcb pads/components are being rotated)
Image

/bounty $75

Abse2001:
Fixed CAD rotation logic to stay consistent with PCB component rotation during packing.
Added proper normalization and handling for bottom-layer flips to ensure accurate 3D alignment.
Updated packing flow to apply rotation changes directly to corresponding CAD components.
Added test and snapshot to verify CAD and PCB rotation match across layers.

/claim #1518
/closes #1518

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@tscircuit/core",
   "type": "module",
-  "version": "0.0.796",
+  "version": "0.0.797",
   "types": "dist/index.d.ts",
   "main": "dist/index.js",
   "module": "dist/index.js",


lib/components/primitive-components/Group/Group_doInitialPcbLayoutPack/applyPackOutput.ts
```diff
@@ -4,6 +4,32 @@ import { transformPCBElements } from "@tscircuit/circuit-json-util"
...
+  db: any,
+  pcbComponentId: string,
+  rotationDegrees: number,
+  layer: string | undefined,
+) => {
+  if (!rotationDegrees) return
+  if (!db?.cad_component?.list) return
+
+  const cad = db.cad_component
+    .list()
```
seveibar: you should be using `.getWhere` instead of `.list`, it has some optimizations

lib/components/primitive-components/Group/Group_doInitialPcbLayoutPack/applyPackOutput.ts
```diff
@@ -87,6 +114,7 @@ export const applyPackOutput = (
             "pcb_component_id" in elm && elm.pcb_component_id === componentId,
         )
       transformPCBElements(related as any, transformMatrix)
+      updateCadRotation(db, componentId, rotationDegrees, pcbComponent.layer)
```
seveibar: 1. named parameters when there are more than 2 args
2. fully qualified id variable names _always_, `componentId` should be `pcbComponentId`
3. Don't change the name of the variables throughout the code, i see `memberId` in some places as well!

lib/components/base-components/NormalComponent/NormalComponent.ts
```diff
@@ -1313,6 +1313,23 @@ export class NormalComponent<
     const accumulatedRotation =
       (decomposedTransform.rotation.angle * 180) / Math.PI
 
+    const normalizeDegrees = (angle: number) => {
+      let normalized = angle % 360
+      if (normalized < 0) normalized += 360
+      return normalized
+    }
```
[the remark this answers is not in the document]
seveibar: (move to the global scope)

seveibar (commented):
looks good, but code quality concerns

lib/components/base-components/NormalComponent/NormalComponent.ts
```diff
@@ -1313,6 +1319,29 @@ export class NormalComponent<
...
+        ? pcb_component.rotation
+        : accumulatedRotation
+
+    // NEW: inherit rotation from parent group if defined
+    let groupRotation = 0
+    if (
+      this.parent?.props?.pcbRotation &&
+      typeof this.parent.props.pcbRotation === "number"
+    ) {
+      groupRotation = this.parent.props.pcbRotation
```
seveibar: This isnt right. There is a method to get the transform if the parent. This will fail with nesting

seveibar: If you need the rotation from the transform, there is a decompose method in transformation – matrix

lib/components/base-components/NormalComponent/NormalComponent.ts
```diff
@@ -1313,6 +1314,19 @@ export class NormalComponent<
     const accumulatedRotation =
       (decomposedTransform.rotation.angle * 180) / Math.PI
 
+    const pcbRotation =
+      pcb_component?.rotation !== undefined
+        ? pcb_component.rotation
+        : accumulatedRotation
```
```suggestion
    const pcbRotation = pcb_component.rotation ?? 0
```

lib/components/primitive-components/Group/Group_doInitialPcbLayoutPack/applyPackOutput.ts
```diff
@@ -1,9 +1,46 @@
...
+  })
+  if (!cadComponent) return
+
+  const delta =
+    layer?.toLowerCase?.() === "bottom" ? -rotationDegrees : rotationDegrees
+
+  const currentRotationZ = cadComponent.rotation?.z ?? 0
+  const nextRotation = {
+    ...(cadComponent.rotation ?? { x: 0, y: 0, z: 0 }),
+    z: (currentRotationZ + delta) % 360,
```
seveibar: why not use the normalizeRotation method

seveibar (approved):

PR #1565, opened 2025-10-23
Fix the autolayout for the pinHeader component having no explicit facingDirection

imrishabh18:
# Before

image

# After

image

### DIFF
--- a/lib/components/primitive-components/Group/Group_doInitialSchematicLayoutMatchPack.ts
+++ b/lib/components/primitive-components/Group/Group_doInitialSchematicLayoutMatchPack.ts
@@ -774,15 +774,9 @@ export function Group_doInitialSchematicLayoutMatchPack<
               originalDirection,
               placement.ccwRotationDegrees,
             )
-            // Update side_of_component to match the new facing direction
-            const facingToSideMap = {
-              left: "left",
-              right: "right",
-              up: "top",
-              down: "bottom",
-            } as const
             port.side_of_component =
-              facingToSideMap[port.facing_direction] || port.side_of_component
+              (port.facing_direction as "left" | "right" | "top" | "bottom") ||
+              port.side_of_component
           }
 
           // Also rotate text positions


lib/components/normal-components/PinHeader.ts
```diff
@@ -2,9 +2,17 @@ import { pinHeaderProps, type SchematicPortArrangement } from "@tscircuit/props"
...
-import type { BaseSymbolName } from "lib/utils/constants"
+import type { z } from "zod"
 
 export class PinHeader extends NormalComponent<typeof pinHeaderProps> {
+  constructor(props: z.input<typeof pinHeaderProps>) {
+    const propsWithDefaults = {
+      ...props,
+      facingDirection: props.facingDirection ?? "right",
+    }
+    super(propsWithDefaults)
```
seveibar: don't modify props (props are immutable)

Instead, create a function called `_getFacingDirection()` and force anything using facing direction to use that. This allows much cleaner overrides and preserves the user's definitions (which makes errors and lots of other stuff more clear)

lib/components/primitive-components/Group/Group_doInitialSchematicLayoutMatchPack.ts
```diff
@@ -119,6 +120,14 @@ function convertTreeToInputProblem(
         // If schFacingDirection is set, don't allow rotation
         availableRotations = [0]
       }
+      // Check for components with _getFacingDirection method (e.g., PinHeader with defaults)
+      if (
+        (component as PinHeader)._getFacingDirection &&
+        typeof (component as PinHeader)._getFacingDirection === "function"
```
seveibar: this is really incorrect

seveibar:
```tsx
if ((component as PinHeader)?._getFacingDirection()) {
```

seveibar: or even better

```tsx
if (component._getSchFacingDirection()) {
```

then define in the PrimitiveComponent `_getFacingSchFacingDirection() { return null }`

seveibar (approved):

PR #1567, opened 2025-10-24
Fix selector resolution with async KiCad footprints

Aqil-Ahmad:
Fixes tscircuit/tscircuit#1149
/claim tscircuit/tscircuit#1149


Aqil-Ahmad:
@seveibar can you please review!

tests/repros/repro-kicad-footprint-pcbpath-selector.test.tsx
```diff
@@ -0,0 +1,99 @@
...
+  const pcbTrace = circuit.db.pcb_trace.list()[0]
+  expect(pcbTrace).toBeDefined()
+  expect(pcbTrace.route.length).toBeGreaterThanOrEqual(3)
+
+  await expect(circuit).toMatchPcbSnapshot(
+    `${import.meta.path}-kicad-footprint-selectors`,
+  )
+})
+
+test("trace pcbPath selectors work with regular footprints (baseline)", async () => {
```
seveibar: remove

seveibar:
A bit complex still since you're relying on a rerender. Did you check if asyncPhaseDependencies can handle this? Seems to me like it would?

Aqil-Ahmad:
>  Did you check if asyncPhaseDependencies can handle this? Seems to me like it would?

you're right! but i think asyncPhaseDependencies alone wont work. this only checks a component's own tree. since R1 and R2 are siblings, we will have to look from root - the render system itself would need to be modified.

Aqil-Ahmad:
> > Did you check if asyncPhaseDependencies can handle this? Seems to me like it would?
> 
> you're right! but i think asyncPhaseDependencies alone wont work. this only checks a component's own tree. since R1 and trace are siblings, we will have to look from root - the render system itself would need to be modified.

@seveibar what do you suggest?

seveibar:
> i think asyncPhaseDependencies alone wont work. this only checks a component's own tree. since R1 and trace are siblings, we will have to look from root - the render system itself would need to be modified.

Hmm interesting, i think it should check the entire tree and look from root. The intent with those dependencies was to make the entire system wait and not let certain parts of the tree move forward.

There's an exception for subcircuits which are supposed to be able to independently render but we can ignore that for now

seveibar:
@Aqil-Ahmad nice insights

seveibar:
@Aqil-Ahmad you seem to understand the code pretty well, definitely reach out to me on discord if you're interested in doing more work together: https://tscircuit.com/join

Aqil-Ahmad:
hi @seveibar! 
should i close this pr if its not the intended solution or would you prefer i adjust it?

PR #1577, opened 2025-10-24
Fix trace issue with KiCad footprints

techmannih:


seveibar (changes requested):
not the correct solution

PR #1578, opened 2025-10-24
Fix schpinspacing width

### LINKED ISSUE #1576: Fix `schPinSpacing` changing the width of the schematic box
- [ ] Reproduce issue in test (you must PROVE you reproduced the issue in the PR description)
- [ ] Fix (width of schematic component box should not change)

/bounty $40

RaghavArora14:
/Fixes #1576
/claim #1576
## Reproduction

Before Fix:
image
image
Small spacing - width: 0.4 height: 1
Large spacing - width: 1.6 height: 4
Test FAILED
## Fix

Changed width/height calculation in getAllDimensionsForSchematicBox.ts to use fixed MIN_PADDING (0.4) instead of schPinSpacing * 2.
After Fix: 
image
## Result

After fix, box width stays constant at 0.4 regardless of schPinSpacing value.

```
Small spacing - width: 0.4 height: 1
Large spacing - width: 0.4 height: 2.8
Tests PASS
```

### DIFF
--- a/lib/components/base-components/NormalComponent/NormalComponent.ts
+++ b/lib/components/base-components/NormalComponent/NormalComponent.ts
@@ -1237,7 +1237,7 @@ export class NormalComponent<
    * Extract pin labels from ports using existing Port logic
    */
   _getPinLabelsFromPorts(): Record<string, string> {
-    const ports = this.selectAll("port")
+    const ports = this.selectAll("port") as Port[]
     const pinLabels: Record<string, string> = {}
 
     for (const port of ports) {


tests/repros/repro-schpinspacing-width.test.tsx
```diff
@@ -0,0 +1,73 @@
...
+  console.log("Top/Bottom Small spacing - width:", sizeSmall.width, "height:", sizeSmall.height)
+  console.log("Top/Bottom Large spacing - width:", sizeLarge.width, "height:", sizeLarge.height)
+
+  // The width SHOULD be different because pins are on top/bottom sides
+  // and schPinSpacing affects the spacing between those pins
+  expect(sizeSmall.width).not.toBe(sizeLarge.width)
+
+  // The height should be the same because there are no pins on left/right sides
+  expect(sizeSmall.height).toBe(sizeLarge.height)
+})
```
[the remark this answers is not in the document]
RaghavArora14: should i fix this test file as graphite says?

seveibar:
@RaghavArora14 request your assigned maintainer for review. Work with them directly, join discord if needed https://tscircuit.com/join

https://docs.google.com/spreadsheets/d/1BU96V97WA6Xv9WyjIyN1qmdWvoJgLEir8mdqz--4B_o/edit?usp=sharing

RaghavArora14:
hey @ShiboSoftwareDev can i get a review on this please, all the tests are now passing, have reproduced the issue and added test snapshots too

reproduce-issue.md
```diff
@@ -0,0 +1,67 @@
+# Reproducing the schPinSpacing Width Bug
```
ShiboSoftwareDev: delete this file, a screenshot of the repro in the PR body is enough

lib/utils/schematic/getAllDimensionsForSchematicBox.ts
```diff
@@ -346,9 +346,11 @@ export const getAllDimensionsForSchematicBox = (
   // Use lengths to determine schWidth and schHeight
   let schWidth = params.schWidth
   if (schWidth === undefined) {
+    const DEFAULT_PIN_SPACING = 0.2
```
ShiboSoftwareDev: you introduced a variable but haven't used it?

ShiboSoftwareDev:
also, if you type something in the contributor channel in discord it will automatically create a thread, I prefer to chat/review on discord

RaghavArora14:
sure, pinged on discord

tests/examples/__snapshots__/example7-voltage-regulator-with-connections-schematic.snap.svg
```diff
@@ -1,4 +1,4 @@
-<svg xmlns="http://www.w3.org/2000/svg" width="1200" height="600" style="background-color: rgb(245, 241, 237)" data-real-to-screen-transform="matrix(73.1707317073,0,0,-73.1707317073,643.987804878,383.8243902439)" data-software-used-string="@tscircuit/core@0.0.769"><style>
```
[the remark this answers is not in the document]
seveibar: eh maybe not

seveibar (commented):
Why is repro3 messed up?

seveibar:
i think repro03 is showing that we introduced a bug no? Shouldn't it have a wider box because it has pin labels?

RaghavArora14:
> i think repro03 is showing that we introduced a bug no? Shouldn't it have a wider box because it has pin labels?

Repro3 doesn't have pin labels, so the box width is correctly calculated based on pin positions (sideLengths) + fixed padding. The box is narrower now because the old code incorrectly added schPinSpacing to the padding - the fix ensures padding stays constant at 0.4mm regardless of pin spacing.

If users need a wider box, they can explicitly set schWidth prop or the code already handles pin labels by adding extra width when labels are present

RaghavArora14:
> you broke more snapshots, your logic doesn't account for pin labels

The minimum width check should only apply when there are NO pin labels (or when the calculated width from labels is less than the minimum). The current logic applies the minimum width even when pin labels already calculated a larger width
image
what about this?

RaghavArora14:
The core issue was that the original code used params.schPinSpacing * 2 for padding, which made box dimensions depend on schPinSpacing (the bug). My fix changed this to a fixed DEFAULT_SCHEMATIC_BOX_PADDING_MM (0.4mm), which correctly makes padding independent of schPinSpacing. However, this created a problem: repro3 has schPinSpacing={0.75}, so it previously got 1.5mm of padding and now only gets 0.4mm, making it appear too narrow. 

I tried several approaches to fix this: (1) adding a minimum width check of 1.0mm for chips with left/right pins - this broke 28 snapshots because it overrode the pin label width calculations, (2) calculating default label width based on pin numbers when no explicit pinLabels are defined - this broke the original behavior where chips without pinLabels got no extra width, and (3) increasing DEFAULT_SCHEMATIC_BOX_PADDING_MM from 0.4mm to 0.6mm - this broke 45 tests.

 The final trade-off is: the bug is fixed (schPinSpacing no longer affects dimensions), but chips with large schPinSpacing values like repro3 will appear narrower than before since they no longer get proportionally larger padding. What do you think - should we accept this trade-off, or do you have any suggestions on how to provide adequate width for repro3 without breaking the other tests?

tests/components/normal-components/__snapshots__/chip-layer-flip-schematic.snap.svg
RaghavArora14: I think this happened because the test has a chip with a manual footprint that has portHints={["1"]}, portHints={["2"]}, portHints={["3"]}. These are NOT generic "pin1", "pin2" format - they're just numbers "1", "2", "3".

Our filter if (!hint.match(/^pin\d+$/)) only skips hints that match the pattern "pin" followed by digits. So "1", "2", "3" would pass through and be used as labels!

This means the chip's schematic box width changed because it now thinks it has labels "1", "2", "3" to display, triggering the 0.5mm minimum width logic.

RaghavArora14:
@ShiboSoftwareDev, thanks for all your help and patience throughout this pr, and yes i will keep a lookout for more issues and try to contribute more. Will keep on learning.

PR #1582, opened 2025-10-24
Fix trace issue with KiCad footprints

techmannih:
fixes #https://github.com/tscircuit/tscircuit/issues/1149
/claim https://github.com/tscircuit/tscircuit/issues/1149


lib/components/base-components/Renderable.ts
```diff
@@ -332,6 +332,12 @@ export abstract class Renderable implements IRenderable {
   }
 
   runRenderPhaseForChildren(phase: RenderPhase): void {
+    const deps = asyncPhaseDependencies[phase] || []
+    for (const depPhase of deps) {
+      if (this._hasIncompleteAsyncEffectsInSubtreeForPhase(depPhase)) {
+        return
+      }
+    }
```
seveibar: i'm confused because this should be handled on line 310

techmannih:
Yes, It’s still working even after I removed it from line 311. I need to dig in. Sorry I will come on it tomorrow.

PR #1591, opened 2025-10-26
feat(ci): add test matrix segmentation to speed up CI

### LINKED ISSUE #1592: Improve CI Test Time with Parallel Matrix Execution
Current CI runs all tests sequentially, making pipelines slow.
Implement a GitHub Actions matrix to run tests in parallel across directories.

Ayushjhawar8:
- Implement parallel test execution using GitHub Actions matrix strategy across 19 test directories
- Add validation script to ensure all test directories are covered by matrix
- Add retry logic (up to 4 attempts) for intermittent segmentation faults (exit codes 139/132)
- Add 3D snapshot artifact upload on failure for easier debugging
- Based on proven implementation from eval package
- Each test file runs in isolation to prevent cascading failures

fix #1592

### DIFF
--- a/.github/workflows/bun-test.yml
+++ b/.github/workflows/bun-test.yml
@@ -27,9 +27,6 @@ jobs:
     needs: validate-test-matrix
     strategy:
       matrix:
-        # Each pattern is evaluated in order by index. Files matched by earlier patterns
-        # are excluded from later patterns, enabling fine-grained test distribution.
-        # The catchall pattern at the end captures any remaining files.
         include:
           - pattern: "tests/repros/repro[1-9]-*.test.tsx tests/repros/repro10-*.test.tsx"
             index: 0


seveibar (commented):
This is a good start but the directory breakdown is not ideal, some of the workflows run in 10s some in 70s, instead of specifying directories can you enable “glob patterns”, then have a catchall glob pattern for any test that wasn’t captured? This will require a change in architecture fwiw but will allow us to split up the repros directory.

The tricky thing im not so sure about is you basically have to exclude any preceding glob when determining which tests should tun for a given matrix node

i think the best way to do this, is to have EVERY matrix node read the workflow and determine all the specific test files to run by evaluating which files already have been matched by previous global patterns

alternatively, we keep your existing approach but redesign the test directory to limit test runtime (meh) or use bun concurrency (also meh)

hopefully you see the issue… ideally our test runs run in just 10-20s but that will require restructuring all directories or smart globs

PR #1613, opened 2025-11-01
Ensure render phases wait for async dependencies across the entire component tree

ShiboSoftwareDev:
This commit provides a systemic fix for a race condition in the rendering pipeline where dependent render phases would
execute before all asynchronous tasks from their dependency phases were complete.                                     

The root cause was that the dependency check was too localized, only considering async effects within a component's   
own subtree. This led to issues where, for example, a `<CopperPour>` would render before a sibling `<group>` had finished asynchronous autorouting.                                                                                             

This has been resolved by refactoring the Renderable class:                                                           

 • A new getTopLevelRenderable method now correctly traverses the component hierarchy to find the top-level Renderable component.                                                                                                         
 • The runRenderPhase method has been updated to use this method, ensuring that it checks for incomplete asynchronous tasks across the entire component tree before executing a phase with dependencies.                                 

This change ensures that phases like PcbCopperPourRender correctly wait for all PcbTraceRender operations to finish,  
regardless of where they are in the component tree, making the rendering process more robust and predictable.


seveibar (changes requested):
This is working around the render system- probably not a good idea

seveibar (approved):
I beleieve thisbis the right fix

PR #1614, opened 2025-11-02
fix: board width and height should be undefined when outline is provided

### LINKED ISSUE #1516: Board with outline should have the width and height as undefined
Take reference from this comment - https://github.com/tscircuit/core/pull/1514#pullrequestreview-3345535623

Enity300:
Fixes #1516
image

Issue: When outline property is opted, previously width and height were also being updated to what the outline bounds tells it. This was needed to be fixed to not cause any ambiguity for the user in the future. 

Fix: When a board has an outline property, width and height are now left as undefined instead of being calculated from the outline bounds. This removes ambiguity about whether the board shape is defined by width/height or by the outline.

Changes:
- Board.doInitialPcbComponentRender: Don't calculate width/height when outline exists
- Board.doInitialPcbBoardAutoSize: Skip width/height update when outline is present
- Added repro70-board-outline.test.tsx to verify width/height are undefined with outline
- Board-outline-dimension.test.tsx: this test file was created for issue 1514, changed the test to expect width as undefined rather than excepting 16

image


imrishabh18 (changes requested):

MustafaMulla29 (commented):
Type check and other workflows are failing

nailoo (commented):
Easy to fix, please update snapshots

Enity300:
Hi! While implementing the shape property for outlined boards, I ran into a TypeScript compilation error.

The issue is that the PCBBoard interface in the circuit-json package currently requires width and height to be numbers. However, when a board uses a custom outline, the width and height properties shouldn't be present at all - the board dimensions come from the outline points instead.

Right now, I'm trying to conditionally set width and height to undefined when an outline is present, but TypeScript is rejecting this because it expects those fields to always be numbers, not undefined.

Could you update the circuit-json package to make the width and height properties optional on the PCBBoard interface? This would allow rectangular boards to specify width and height normally, while outlined boards can omit these properties and rely solely on the outline.

This is currently blocking the PR because the type-check tests are failing, which is in turn causing build test failure

tests/fixtures/extend-expect-3d-matcher.ts
```diff
@@ -56,25 +56,62 @@ export async function resolvePoppyglOptions(
...
   }
 
   if (!resolvedOpts.camPos && board) {
-    resolvedOpts.camPos = [
-      board.width! / 2,
-      (board.width! + board.height!) / 2,
-      board.height! / 2,
-    ]
+    // Handle outlined boards
+    if ((board as any).shape === "polygon" && board.outline) {
```
imrishabh18:
```suggestion
    if (board.shape === "polygon" && board.outline) {
```

lib/components/normal-components/Board.ts
```diff
@@ -267,6 +272,9 @@ export class Board extends Group<typeof boardProps> {
     const pcbBoard = this.root!.db.pcb_board.get(this.pcb_board_id!)
     if (!pcbBoard) return
 
+    // Only add board information for rectangular boards with width/height
+    if (!pcbBoard.width || !pcbBoard.height) return
```
seveibar: introduces bug

seveibar:
this issue is valid but i don't think it's important enough where we can risk the upstream breakage w.r.t. undefined width and height. Reduce scope to just add the `shape` property

seveibar:
@Enity300 no it's not necessary to revert the changes, i'm saying that we can leave tscircuit/core out-of-spec. But you can add `shape` to `board` to bring us closer to spec. Sometimes we wait until a big release to fix things that are out of spec so that we're not breaking too much at once

PR #1621, opened 2025-11-05
feat: Refactor copper pour to use @tscircuit/copper-pour-solver

ShiboSoftwareDev:
This PR replaces the internal copper pour generation logic with the new @tscircuit/copper-pour-solver package.                            

The new implementation uses convertCircuitJsonToInputProblem to prepare the data for the solver and then utilizes CopperPourPipelineSolver to generate the BRep shapes for the pour.                                                                                                 

This change delegates complex geometry calculations to a more robust, dedicated package.                                                  

 • Replaced pour generation logic in CopperPour.ts                                                                                        
 • Added @tscircuit/copper-pour-solver dependency                                                                                         
 • Updated and added tests to reflect the new implementation and verify margin functionality


seveibar (approved):
very exciting

PR #1646, opened 2025-11-12
Fix PCB pin label rotation to keep text readable

### LINKED ISSUE #1645: verify whether pcb pinlabels are rendering correctly
they look strange in this example:
Image
repro: https://github.com/tscircuit/sparkfun-boards/blob/main/boards/SparkFun-IR-Array-Breakout-55Degree-FOV-MLX90640/__snapshots__/SparkFun-IR-Array-Breakout-55Degree-FOV-MLX90640.circuit-pcb.snap.svg

RaghavArora14:
## Fix PCB pin label rotation to keep text readable

Fixes #1645

Pin labels on rotated pinrow components were appearing at odd angles (45°, 135°, etc.), making them difficult to read. This fix normalizes pin label rotation to always be horizontal (0°) or vertical (90°), never upside-down or diagonal.

### Changes
- Modified `createPinrowSilkscreenText.ts` to snap rotation to nearest 90° increment
- Ensures text is never upside-down by converting 180° → 0° and 270° → 90°
- Added test case with pinrows at 0°, 45°, 90°, and 135° rotations

### Before
Pin labels rotated with component (e.g., 45° diagonal text)
image

### After
Pin labels always horizontal or vertical, readable from standard viewing angles
image


RaghavArora14:
updated:
image

RaghavArora14:
updated:
image

PR #1656, opened 2025-11-13
feat: Add support for inflating traces from subcircuit circuitJson

ShiboSoftwareDev:
This change introduces the ability for <subcircuit> components to inflate source_trace elements from a circuitJson
prop. This allows for defining pre-wired subcircuits and having those internal connections be respected upon      
instantiation.                                                                                                    

A new inflator, inflateSourceTrace, has been added to handle the creation of <Trace> components from source_trace 
definitions. The logic correctly traverses the group hierarchy within the subcircuit's JSON to construct valid    
hierarchical selectors for trace paths.                                                                           

A new test case has been added to verify that a trace connecting two resistors within a subcircuit's circuitJson  
is correctly inflated and rendered.


seveibar (approved):
nice

PR #1657, opened 2025-11-13
feat: Add support for inflating plated holes from subcircuit circuitJson

Abse2001:
This PR adds full plated hole inflation support inside subcircuits.
It introduces a new `inflateSourcePlatedHole` inflator, registers it in the Subcircuit pipeline, and ensures that any pcb_plated_hole objects inside a subcircuit’s circuitJson are:

correctly inflated into PlatedHole primitives

inserted into the main circuit DB

included in the final rendered circuit JSON

A new test (`subcircuit-circuit-json07-plated-hole-inflation`) verifies correct inflation and persistence.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@tscircuit/core",
   "type": "module",
-  "version": "0.0.865",
+  "version": "0.0.866",
   "types": "dist/index.d.ts",
   "main": "dist/index.js",
   "module": "dist/index.js",


seveibar (changes requested):
when a source component etc. is inflated, it's footprint is inflated, only source elements should be inflated within the subcircuit function, source elements cause all their direct relationships to be inflated

there are some exceptions to this- there are pcb components without a parent component, those cases should be explicitly/specially handled, e.g. inflateFootprintElementsWithoutComponent

PR #1661, opened 2025-11-15
Auto-Inflate PCB-Only Components Without Source Data (KiCad/External JSON Support)

Abse2001:
added support for inflating PCB components that exist only in the PCB section of circuit JSON (e.g., imported from KiCad) and do not have matching source components.

- Detects pcb_component entries with ports but no corresponding source_component.

- Creates a placeholder Chip component to host their PCB primitives.

- Uses the new inflatePcbComponent path for clean integration.

- Ensures mixed schematic/PCB or PCB-only circuit JSONs fully inflate without missing components.

This significantly improves compatibility with external PCB-only inputs and prevents missing or invisible components during subcircuit inflation.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@tscircuit/core",
   "type": "module",
-  "version": "0.0.867",
+  "version": "0.0.868",
   "types": "dist/index.d.ts",
   "main": "dist/index.js",
   "module": "dist/index.js",


tests/features/subcircuit-circuit-json/subcircuit-circuit-json07.test.tsx
```diff
@@ -0,0 +1,57 @@
...
+    )
+  })
+
+  const pcbTraces = circuitJson.filter((c) => c.type === "pcb_trace")
+  const pcbTraceR1R2 = pcbTraces.find(
+    (c: any) => c.source_trace_id === (sourceTraceR1R2 as any).source_trace_id,
+  )
+
+  expect(circuit).toMatchPcbSnapshot(import.meta.path)
+})
```
baeoc: @Abse2001 test is failing, But workflow is not failing. Why?

PR #1665, opened 2025-11-20
feat: add support for the `soldermask_margin` in smtpads, hole and plated holes

imrishabh18:


tests/components/primitive-components/smtpad-with-solder-mask-margin.test.tsx
```diff
@@ -0,0 +1,63 @@
...
+    <board width="12mm" height="6mm">
+      <chip name="U1" layer="top" footprint={footprint} />
+    </board>,
+  )
+
+  circuit.render()
+
+  const circuitJson = circuit.getCircuitJson()
+  const pcb_smtpad = circuitJson.filter((elm) => elm.type === "pcb_smtpad")
+  console.log(pcb_smtpad)
```
seveibar:
```suggestion
```

lib/components/primitive-components/Hole.ts
```diff
@@ -48,6 +49,10 @@ export class Hole extends PrimitiveComponent<typeof holeProps> {
     const { _parsedProps: props } = this
     const subcircuit = this.getSubcircuit()
     const position = this._getGlobalPcbPositionBeforeLayout()
+    const soldermaskMargin = props.solderMaskMargin
+      ? distance.parse(props.solderMaskMargin)
```
seveibar: yea you shouldn't need to call distance.parse, props must not have put in `distance` properly, this is already parsed with `_parsedProps`

PR #1688, opened 2025-11-28
Add support for single layer board

techmannih:


tests/components/normal-components/board-single-layer.test.tsx
```diff
@@ -0,0 +1,35 @@
...
+      />
+    </board>,
+  )
+
+  await circuit.renderUntilSettled()
+
+  // Verify that allLayers returns only ["top"]
+  const boardComponent = circuit.firstChild
+  expect(boardComponent).not.toBeNull()
+  if (boardComponent && "allLayers" in boardComponent) {
```
seveibar: don't put this type of assertion in a test never use `if` conditions with `expect` in tests

seveibar (changes requested):
this should throw an error, smtpads aren't allowed on the bottom layer when only the top layer has copper

when components are put on the bottom layer, they must be through hole (plated holes)

PR #1692, opened 2025-11-28
Fix: repro73 trace coming out on other side of pad

MustafaMulla29:


seveibar (approved):

imrishabh18 (changes requested):
Nahh, I don't think this is correct. It's just making it too complicated. Why do you have to clip the trace?

Also, thinking more about this that trace going out the other end of the pad doesn't make the circuit invalid!

cc @seveibar

imrishabh18:
Yes, go ahead!

PR #1709, opened 2025-12-03
Prevent nested board deeper than 1 level

Ayushjhawar8:
- Added ` _getDescendantBoard()`  method to check if a component has any board descendants
- Enhanced  add()  method with two-way validation:
- Checks if the component being added is a board with a board ancestor (existing check)
- Checks if we're adding to a board and the component contains board descendants (new check - needed because React reconciler builds trees bottom-up)
- Added test file  `tests/features/nested-board-prevention.test.tsx`  to verify the validation works


seveibar (changes requested):
PrimitiveComponent is very sensitive to pollution, dont add any LoC for this there

PR #1719, opened 2025-12-05
fix(panel): correct logic that always overwrote user-provided width/height with autosized values

AnasSarkiz:
### **Overview**

This PR fixes a critical issue where the `Panel` component **ignored explicit `width` and `height` props** and **always replaced them with auto-sized values**, preventing users from defining custom panel dimensions.

### **Root Cause**

`doInitialPanelLayout()` applied auto-calculated dimensions **unconditionally**, overwriting any user-provided width/height values on every initial render.

### **Fix**

The Panel logic is updated so **explicit width/height props always take precedence**, and auto-sizing only occurs when the props are *not* provided.

Key changes:

1. **`doInitialPanelLayout()`**

   * Now applies auto-calculated width/height **only when the props are `undefined`**.
   * Prevents auto-sizing from overwriting explicit user input.

2. **`doInitialPcbComponentRender()`**

   * Uses `0` as a placeholder for width/height when they are not provided, enabling proper auto-sizing in layout.

3. **`updatePcbComponentRender()`**

   * Preserves existing (possibly auto-calculated) dimensions when props remain undefined.
   * Avoids overwriting user-defined dimensions or clobbering stable auto-sized values.

### **What This Fixes**

* Users can now pass `<Panel width="100mm" height="80mm" />` and have the dimensions respected.
* Auto-sizing still works when width/height are intentionally omitted.
* Prevents unexpected layout changes caused by previously forced auto-sizing.

### **Files Changed**

* `lib/components/normal-components/Panel.ts`

  * Updated dimension-handling logic across:

    * `doInitialPanelLayout()`
    * `doInitialPcbComponentRender()`
    * `updatePcbComponentRender()`

### **Tests**

* Added: `panel-explicit-dimensions.test.tsx`

  * Verifies that explicit width/height props are now honored.
  * Ensures auto-sizing only triggers when dimensions are omitted.

### **Note**

Full autosize support (omitting width/height entirely) requires an updated release of `@tscircuit/props` where width/height fields are optional.
Local props already include this (`distance.optional()`), but core is still using **v0.0.424**, where width/height remain required.

### DIFF
--- a/lib/components/normal-components/Panel.ts
+++ b/lib/components/normal-components/Panel.ts
@@ -146,26 +146,26 @@ export class Panel extends Group<typeof panelProps> {
         maxY = Math.max(maxY, top)
       }
 
-      if (isFinite(minX)) {
+      // Skip auto-calculation if both dimensions are explicitly provided
+      const hasExplicitWidth = this._parsedProps.width !== undefined
+      const hasExplicitHeight = this._parsedProps.height !== undefined
+
+      if (hasExplicitWidth && hasExplicitHeight) {
+        db.pcb_panel.update(this.pcb_panel_id!, {
+          width: distance.parse(this._parsedProps.width),
+          height: distance.parse(this._parsedProps.height),
+        })
+      } else if (isFinite(minX)) {
         const boundsWidth = maxX - minX
         const boundsHeight = maxY - minY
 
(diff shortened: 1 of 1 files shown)


AnasSarkiz:
@seveibar updated

ShiboSoftwareDev (approved):

PR #1734, opened 2025-12-09
Add packing lifecycle events and improve packing error reporting

Abse2001:
## Summary

- Introduces new lifecycle events: `packing:start`, `packing:end`, and `packing:error`.
- Emits metadata for each event, including `subcircuit_id` and `componentDisplayName`, enabling better UI hooks and debugging.
- Wraps the packing operation in a try/catch to surface errors through `packing:error` before rethrowing.
- Adds `example36` tests and snapshot outputs to verify correct event firing behavior.

### DIFF
--- a/package.json
+++ b/package.json
@@ -106,7 +106,7 @@
   "dependencies": {
     "@flatten-js/core": "^1.6.2",
     "@lume/kiwi": "^0.4.3",
-    "calculate-packing": "0.0.68",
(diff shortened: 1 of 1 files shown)


Abse2001:
@seveibar I'm trying to prevent packing and auto routing from being called when we use circuitJson inflation, so I thought this will be nice to have , your use is better tbh.

PR #1737, opened 2025-12-10
Implement nested board restriction and add corresponding test case

Ayushjhawar8:
```
<board>
<group>
<board/>
</group>
</board>
``` 
throws not supported


seveibar (changes requested):
error'ing at wrong time, plus throwing a repeated error message makes it extremely difficult to track down errors

https://github.com/tscircuit/core/blob/9a2299f3caab355f528c1a9502275d0dbe8f8195/lib/components/base-components/PrimitiveComponent/PrimitiveComponent.ts#L668

PR #1745, opened 2025-12-11
fix interconnect typing and props version

0hmX:
ref: https://github.com/tscircuit/core/pull/1741


seveibar (commented):
Whats the fix?

0hmX:
lib/components/normal-components/Interconnect.ts 

I do not know why the CI is green, but in my VSCode, I got type errors when I open the file lib/components/normal-components/Interconnect.ts 

This fixes that type error

1. The comment to ignore the error was one line above, from what it was supposed to be. 
2. Also, I was getting a type error tscircuit/props, so I updated the version

seveibar:
hmmmmmmmmm

seveibar:
weird

seveibar:
CI is throwing for me as well- something might be _weird_

PR #1754, opened 2025-12-11
Prevent autorouting and packing when circuitJson is provided, and add circuitJson-based PCB trace insertion with full Group integration and snapshot tests

Abse2001:
This PR adds full support for inserting PCB traces directly from circuitJson, including port remapping, coordinate translation,
and consistent placement inside the Group lifecycle.
Autorouting and packing are automatically disabled when circuitJson is provided, ensuring user-defined routing and geometry remain authoritative.
The new insertion module maps source ports to target pcb_ports and reconstructs trace routes with correct offsets.
Group and layout-pack behavior now align with externally supplied circuitJson without altering component positioning.
A new example with snapshot tests validates deterministic trace reproduction and rendered output.
Rendering fixtures were updated to use renderUntilSettled for stable async lifecycle completion.
Overall, this enables reliable PCB generation from pre-routed circuitJson data.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@tscircuit/core",
   "type": "module",
-  "version": "0.0.926",
+  "version": "0.0.927",
   "types": "dist/index.d.ts",
   "main": "dist/index.js",
   "module": "dist/index.js",


lib/components/primitive-components/Group/Group.ts
```diff
@@ -631,6 +631,7 @@ export class Group<Props extends z.ZodType<any, any, any> = typeof groupProps>
 
   doInitialPcbTraceRender() {
     const debug = Debug("tscircuit:core:doInitialPcbTraceRender")
+    if (this.getInheritedProperty("circuitJson")) return
```
seveibar: hmm i think its more clear as...

wdyt? you can define that function as `return Boolean(this.getInheritedProperty("circuitJson"))`
```suggestion
    if (this.isPrerendered()) return
```

lib/components/primitive-components/Group/Subcircuit/inflators/inflatePcbTrace.ts
```diff
@@ -0,0 +1,22 @@
...
+export const inflatePcbTrace = (
+  pcbTrace: PcbTraceElement,
+  inflatorContext: InflatorContext,
+) => {
+  const { subcircuit } = inflatorContext
+
+  subcircuit.add(
+    new PcbTrace({
+      route: pcbTrace.route,
+      source_trace_id: pcbTrace.source_trace_id,
```
seveibar: this isn't right- you need to create a `Trace` not a `PcbTrace`, and you can't use the source_trace_id of the original circuitJson (that wouldn't work at all, there is nothing to create that source trace)

PR #1765, opened 2025-12-14
feat: Add fiducial primitive component

ShiboSoftwareDev:
This PR introduces the Fiducial primitive component.                                                   

Key features:                                                                                              

 • Creates a circular pcb_smtpad to represent a fiducial mark on the PCB.                                  
 • The soldermaskPullback prop can be used to    
   control the solder mask margin.                                                                         
 • Includes tests to verify the functionality and visual output.


lib/components/normal-components/Fiducial.ts
```diff
@@ -0,0 +1,75 @@
...
+      shape: "circle",
+      x: pcbX,
+      y: pcbY,
+      radius: padDiameter / 2,
+      layer: props.layer ?? "top",
+      pcb_component_id: this.pcb_component_id,
+      subcircuit_id: this.getSubcircuit().subcircuit_id ?? undefined,
+      is_covered_with_solder_mask: true,
+      soldermask_margin: soldermaskPullback,
+    } as PcbSmtPadCircle)
```
seveibar: add a child `new SmtPad`

lib/components/normal-components/Fiducial.ts
```diff
@@ -0,0 +1,75 @@
...
+  doInitialSourceRender() {
+    const { db } = this.root!
+    const source_component = db.source_component.insert({
+      ftype: "simple_fiducial",
+      name: this.name,
+    } as Omit<SourceSimpleFiducial, "source_component_id" | "type">)
+    this.source_component_id = source_component.source_component_id
+  }
+
+  doInitialSchematicComponentRender() {
```
seveibar: just set `noSchematicRepresentation` on this- (don't set a prop- props are not allowed to be touched, it depends on how the implementation works but something like this...

```
get hasSchematicRepresenation() {
  return !this._parsedProps.noSchematicRepresentation
}
```

in NormalComponent then an override in `Fiducial` is a good idea

seveibar (commented):
generally good, i'd like to see the SmtPad added as a child

lib/components/normal-components/Fiducial.ts
seveibar: it's also possible @ShiboSoftwareDev that this is not a NormalComponent, this may be a PrimitiveComponent, in which case you don't really have to do this.

PR #1767, opened 2025-12-14
Anchor-Based PCB Group Positioning with Deterministic Offsets

Abse2001:
This PR introduces explicit anchor-based positioning for PCB groups, enabling precise placement relative to configurable anchor points.
Group anchor alignment is now derived consistently from group geometry, outlines, or child component bounds when explicit dimensions are unavailable.
A shared utility computes anchor positions deterministically from group center, size, and selected nine-point anchor.
PCB group metadata is updated to persist both anchor alignment and computed anchor position.
This ensures stable offsets, predictable layout behavior, and accurate overlay visualization across boards.
New snapshot tests validate anchor positioning, offsets, and rendered group bounds.

### DIFF
--- a/lib/components/primitive-components/Group/Group.ts
+++ b/lib/components/primitive-components/Group/Group.ts
@@ -126,13 +126,9 @@ export class Group<Props extends z.ZodType<any, any, any> = typeof groupProps>
           y: distance.parse(point.y),
         }))
       : undefined
-
+    const ctx = this.props
     const anchorPosition = this._getGlobalPcbPositionBeforeLayout()
-    const center = computeCenterFromAnchorPosition(
-      anchorPosition,
-      { width: this.props.width, height: this.props.height },
-      this.props.pcbAnchorAlignment,
-    )
+    const center = computeCenterFromAnchorPosition(anchorPosition, ctx)
 
     const pcb_group = db.pcb_group.insert({
       is_subcircuit: this.isSubcircuit,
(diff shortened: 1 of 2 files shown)


seveibar (changes requested):
don't use IIFEs

also the center is derived from the anchor, not the anchor from the center

lib/components/primitive-components/Group/Group.ts
```diff
@@ -140,6 +140,7 @@ export class Group<Props extends z.ZodType<any, any, any> = typeof groupProps>
             trace_clearance: props.autorouter.traceClearance,
           }
         : undefined,
+      anchor_alignment: props.pcbAnchorPosition ?? "center",
```
seveibar: Ugh what is happening
```suggestion
      anchor_alignment: props.pcbAnchorAlignment ?? null,
```

lib/components/primitive-components/Group/Group.ts
```diff
@@ -126,12 +127,19 @@ export class Group<Props extends z.ZodType<any, any, any> = typeof groupProps>
         }))
       : undefined
 
+    const anchorPosition = this._getGlobalPcbPositionBeforeLayout()
+    const center = computeCenterFromAnchorPosition(
+      anchorPosition,
+      { width: this.props.width, height: this.props.height },
+      this.props.pcbAnchorAlignment,
```
seveibar: Two parameter rule… use named parameters

PR #1771, opened 2025-12-17
Implement circuit-json trace inflation and disable autorouting and packing for inflations

Abse2001:
This PR adds first-class support for inflating traces directly from provided circuit-json data.
Source traces and PCB traces are now converted into `Trace` components during inflation.
When PCB routes are present, traces are treated as pre-routed and their geometry is preserved.
Boards and groups backed by circuit-json are marked as prerendered to prevent layout mutations.
Autorouting and PCB packing are skipped to avoid modifying externally defined routing.
Includes updated examples and snapshot coverage to validate the new behavior.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@tscircuit/core",
   "type": "module",
-  "version": "0.0.936",
+  "version": "0.0.937",
   "types": "dist/index.d.ts",
   "main": "dist/index.js",
   "module": "dist/index.js",


lib/utils/circuit-json/inflate-circuit-json.ts
```diff
@@ -81,7 +81,54 @@ export const inflateCircuitJson = (
   }
 
   const sourceTraces = injectionDb.source_trace.list()
+  const pcbTraces = injectionDb.pcb_trace.list()
+
+  const pcbTracesBySourceTraceId = new Map<string, typeof pcbTraces>()
+  for (const pcbTrace of pcbTraces) {
+    const sourceTraceId = (pcbTrace as any).source_trace_id as
```
ShiboSoftwareDev: don't use any casting

lib/components/normal-components/Board.ts
```diff
@@ -558,25 +559,28 @@ export class Board
...
-        if (childOldCenter) {
-          child._repositionOnPcb({
-            x: childOldCenter.x + deltaX,
-            y: childOldCenter.y + deltaY,
-          })
+        if (
+          !("subcircuit_id" in elm) ||
+          typeof elm.subcircuit_id !== "string"
+        ) {
+          return false
```
seveibar: Lets not change this function with special cases, they arent obvious, all the changes here seem like a hack- why?

lib/utils/circuit-json/inflate-circuit-json.ts
```diff
@@ -81,7 +81,50 @@ export const inflateCircuitJson = (
...
+
+  const pcbTracesBySourceTraceId = new Map<string, typeof pcbTraces>()
+  for (const pcbTrace of pcbTraces) {
+    const sourceTraceId = pcbTrace.source_trace_id as string | undefined
+    if (!sourceTraceId) continue
+    const list = pcbTracesBySourceTraceId.get(sourceTraceId) ?? []
+    list.push(pcbTrace)
+    pcbTracesBySourceTraceId.set(sourceTraceId, list)
+  }
+
```
seveibar: Naming unclear. There are two types of ids, some from the inflation db and others from the circuit, so you need to prefix. Also this datastructure is unnecessary, just use the inflation db

seveibar: Dont create id maps because more variables adds complexity for little gain

lib/utils/circuit-json/inflate-circuit-json.ts
```diff
@@ -81,7 +81,50 @@ export const inflateCircuitJson = (
...
+  const hasAnyRoutedPcbTrace = pcbTraces.some(
+    (t) => Array.isArray(t.route) && t.route.length > 0,
+  )
+  if (sourceTraces.length > 0 && !hasAnyRoutedPcbTrace) {
+    const anyTarget = target
+    if (
+      typeof anyTarget?.setProps === "function" &&
+      !anyTarget?.props?.autorouter
+    ) {
+      anyTarget.setProps({ ...anyTarget.props, autorouter: "sequential-trace" })
```
seveibar: Wat why

PR #1779, opened 2025-12-18
Normalize board coordinates for consistent autorouting in panels

ShiboSoftwareDev:
When a `<board>` is placed inside a `<panel>`, its position is often offset from the origin (0,0). The local          
autorouting logic (_runLocalAutorouting) generates a simpleRouteJson based on the global coordinates of the       
components. However, some parts of the routing pipeline implicitly assume the problem space is centered around the
origin. This discrepancy caused routing to fail or produce incorrect results for boards inside panels.            

This PR introduces a normalization step for local autorouting within a board context:                             

This change ensures that the autorouter sees the same relative geometry for a board whether it's standalone or    
placed within a panel, leading to consistent and correct routing results.                                         

A new test (tests/components/normal-components/panel-vs-board.test.tsx) has been added to verify this fix by      
snapshotting the same complex board both standalone and inside a panel, confirming the routing is now identical.

This also fixes the repro10-usbc-flashlight-panel-auto-layout.test.tsx test, ensuring that all boards within the  
panel are routed identically to a single standalone instance.


ShiboSoftwareDev:
@seveibar board caching is not the fix, since this issue will be present even if you had to route just one board, since my fix is not valid then I guess I just have to fix the autorouter to handle these boards correctly

PR #1794, opened 2025-12-24
test: add autorouting test for one-sided board with top layer traces

Ayushjhawar8:


seveibar (changes requested):
we already have tests like this, and you're using a sort of deprecated autorouter

Ayushjhawar8:
Unstale

seveibar (approved):

PR #1817, opened 2026-01-06
keyboard issue investigation

ShiboSoftwareDev:
the issue was that the parent group wasn't positioned therefore it was packed awkwardly, I believe this is the expected behavior so there is no issue correct?


seveibar (changes requested):
the inner group should have pcbX={0} and pcbY={0} iiuc

PR #1818, opened 2026-01-06
add connectsTo array for net

techmannih:


seveibar (changes requested):
netProps shouldn't even be defined here.

tests/components/primitive-components/net-connectsTo-array.test.tsx
```diff
@@ -0,0 +1,52 @@
...
+  const traces = circuit.db.source_trace
+    .list()
+    .map((t) => t.display_name)
+    .sort()
+  expect(traces).toMatchInlineSnapshot(`
+    [
+      "net.VCC to net.VCC_MAIN",
+      "net.VCC_REGULATED to net.VCC_MAIN",
+    ]
+  `)
```
techmannih: now i am verify net-to-net connectivity inline test,, not schematic visuals

seveibar (changes requested):
i don't really feel like this test shows that it works, you need a pcb snapshot

seveibar:
@techmannih please just introduce one test, and think carefully about how to make sure it works

PR #1819, opened 2026-01-08
fix(panel): Ensure panel layout waits for async footprints to resolve

ShiboSoftwareDev:
This change corrects an issue where panel layout would run before asynchronous footprints (e.g., from an external      
library) were fully loaded. This caused incorrect board sizing and placement within the panel.                               

By adding a render phase dependency, panel layout now waits for all footprints to resolve, ensuring accurate board dimensions and correct positioning. A new test case has been added to verify this behavior.


seveibar (approved):
YEAAAAA

PR #1825, opened 2026-01-09
Prevent all trace routing outside board boundaries

ShiboSoftwareDev:
Implements board boundary validation preventing invalid cross-board traces in both manual and automatic routing scenarios.

Key improvements:
- Added arePortsWithinBoardOutline utility with polygon/bounds containment validation
- Manual traces now validate port locations before PCB trace creation with error reporting
- Fixed auto-routing bounds calculation to respect board boundaries in multi-board layouts
- Test coverage demonstrating boundary validation behavior

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@tscircuit/core",
   "type": "module",
-  "version": "0.0.957",
+  "version": "0.0.958",
   "types": "dist/index.d.ts",
   "main": "dist/index.js",
   "module": "dist/index.js",


lib/components/primitive-components/Trace/Trace_doInitialPcbManualTraceRender.ts
```diff
@@ -201,6 +202,18 @@ export function Trace_doInitialPcbManualTraceRender(trace: Trace) {
     end_pcb_port_id: otherPort.pcb_port_id!,
   })
 
+  if (!arePortsInSameBoard(ports, db)) {
```
seveibar: this doesn't seem like the right fix, why are we even retrieving ports from a different board to begin with? I'm looking through the code

seveibar (changes requested):
see discord comment here https://discord.com/channels/1233487248129921135/1458902999903633522/1459293401785696330

PR #1829, opened 2026-01-10
feat: Emit circuit JSON warnings for conflicting solder mask properties

ShiboSoftwareDev:
Circuit JSON warnings when coveredWithSolderMask is true and solderMaskMargin is also set on solder mask controlling components (SmtPad, PlatedHole, Hole). The warnings are emitted as source_property_ignored_warning entries in the circuit JSON using the proper getParentNormalComponent() method to find the source component.
Updated corresponding tests to expect these warnings. Components now emit clear guidance when both properties are specified.


seveibar (approved):

seveibar (changes requested):
break out a common method

PR #2088, opened 2026-03-31
fix: correct courtyard rotation calculation for flipped components

techmannih2:


MustafaMulla29 (approved):
Can you add mor tests with different rotations.

Wait for seve!!

seveibar (approved):
good

PR #2118, opened 2026-04-09
Feat: add inner symbol for usbc

MustafaMulla29:


seveibar (approved):
make sure to set up the defaults for USBC such that there aren't overlaps

Also I think the schematic box itself is not arranged well, you can fix in a future PR, but I think typically the pins are labelled differently, in fact the snapshots show that the standard isn't actually working at all since both symbols have different pin names (incorrectly)

image

PR #2123, opened 2026-04-09
Update calculate-packing version to 0.0.71

rushabhcodes:
holes were added as obstacle in the newer version, i faced this issue while creating a sparkfun board which now can be fixed after this update.


seveibar (approved):
include motivation in your title/description. Why are you updating it? What feature do you need?

PR #2127, opened 2026-04-10
Add phased subcircuit autorouting

ShiboSoftwareDev:


lib/components/primitive-components/Group/Group.ts
```diff
@@ -396,12 +402,119 @@ export class Group<Props extends z.ZodType<any, any, any> = typeof groupProps>
     return traces.length > 0
   }
 
+  _getRoutingPhasePlans(): RoutingPhasePlan[] {
```
seveibar: should this be in the subcircuit class?

lib/components/primitive-components/Group/Group.ts
```diff
@@ -553,153 +706,180 @@ export class Group<Props extends z.ZodType<any, any, any> = typeof groupProps>
...
+      pads: Array<{
+        center: { x: number; y: number }
+        width: number
+        height: number
+        layer: string
+      }>
+    }> = []
+    let obstacleCircuitJson: AnyCircuitElement[] = []
+
+    const solvePhase = async (phasePlan: RoutingPhasePlan) => {
```
seveibar: so generally, named closures are a huge antipattern, for some reason codex loves making them, but we'll rarely accept a PR with them, and staff are instructed to make sure we don't merge code with named closures especially into core

tests/subcircuits/subcircuit-phased-autorouting-power-first.test.tsx
```diff
@@ -0,0 +1,103 @@
...
+import { getTestFixture } from "tests/fixtures/get-test-fixture"
+
+test("subcircuit phased autorouting routes thick power traces before signals", async () => {
+  const { circuit } = getTestFixture()
+
+  circuit.add(
+    <board width="24mm" height="20mm">
+      <subcircuit
+        name="POWER_STAGE"
+        autorouter={{ local: true, groupMode: "subcircuit" }}
```
seveibar:
```suggestion
```

tests/subcircuits/subcircuit-phased-autorouting-power-first.test.tsx
```diff
@@ -0,0 +1,103 @@
+import { expect, test } from "bun:test"
+import { getTestFixture } from "tests/fixtures/get-test-fixture"
+
+test("subcircuit phased autorouting routes thick power traces before signals", async () => {
+  const { circuit } = getTestFixture()
+
+  circuit.add(
+    <board width="24mm" height="20mm">
+      <subcircuit
```
seveibar: remove the subcircuit- why is there a subcircuit?

seveibar (commented):
this is too confusing as written, but it works so it's a good first attempt

the diff is really hard to read because you used closures. You should request changes for any PR that has closures- we don't accept them because they cause a codebase to rapidly become unmaintanable because of scope leaking

seveibar:
@ShiboSoftwareDev you might want to do 2 PRs, in the first PR you can just "set up" the autorouting system to avoid some kind of large refactor. The diff is impossible to read as is and this is sensitive logic. If the code was broken up properly there would not be this issue, so you might want to clean up the functions and do basic methods in the first PR, then do the actual implementation in a second PR. Being careful and keeping this code clean will save us hundreds of hours in the long term

PR #2131, opened 2026-04-13
Standardize USB-C connector schematic pin labels

MustafaMulla29:
@seveibar should the `SHELL` handling be removed?


lib/utils/connectors/getStandardUsbCPinLabels.ts
```diff
@@ -0,0 +1,76 @@
...
+  { label: "DM2", aliases: ["DN2"] },
+  { label: "DP2", aliases: [] },
+  { label: "CC2", aliases: [] },
+  { label: "VBUS2", aliases: [] },
+  { label: "GND2", aliases: [] },
+  { label: "SHELL1", aliases: ["MH1", "EH1", "MOUNT1"] },
+  { label: "SHELL2", aliases: ["MH2", "EH2", "MOUNT2"] },
+  { label: "SHELL3", aliases: ["MH3", "EH3", "MOUNT3"] },
+  { label: "SHELL4", aliases: ["MH4", "EH4", "MOUNT4"] },
+]
```
seveibar: this is good

lib/components/normal-components/Connector.ts
```diff
@@ -7,18 +7,47 @@ import {
...
 
 export class Connector<
   PinLabels extends string = never,
 > extends Chip<PinLabels> {
+  /**
+   * Labels derived from the fetched part's circuit JSON (e.g. standard
+   * USB-C signal names). Applied as additional per-pin aliases in initPorts
+   * so they flow to source_port.port_hints and the schematic display.
+   */
+  private _derivedStandardPinLabels: Record<string, string> | null = null
```
seveibar: this is an oxymoron, you cannot be both derived and standard, it confuses the direction of information

seveibar (commented):
do not confuse the direction of information. The standard pin labels are not derived.

seveibar:
The code needs to be reoriented to avoid the confusion- it is written as if we use the circuit json to get pin labels. The pin labels do not come from the circuit json, we match the circuit json to standard pin labels

seveibar (changes requested):
Same comment as before. Why is getStandardPinLabels taking in circuit json- it doesnt make sense. You need to start over

seveibar (approved):

seveibar (changes requested):
Why arent thr stmbols the same tho?

seveibar (approved):
Thr symbols need to be identical

seveibar:
You can fix in the next pr

PR #2132, opened 2026-04-14
Deduplicate logical ports from composite footprint copper

ShiboSoftwareDev:
Fixes composite footprints where multiple copper primitives share the same logical pin,
  such as an SMT pad plus plated hole with the same portHints.

  Previously, footprint port discovery returned one port per copper primitive, so repeated
  pin1/pin2 hints could inflate the inferred pin count and create extra logical pins like
  pin3 and pin4.

  This adds a shared helper to normalize footprint portHints, merge aliases, and dedupe
  ports by logical pin number for both string and JSX footprints.

### DIFF
--- a/lib/components/base-components/NormalComponent/utils/getLogicalPortsFromPortHintGroups.ts
+++ b/lib/components/base-components/NormalComponent/utils/getLogicalPortsFromPortHintGroups.ts
@@ -6,6 +6,18 @@ type PortHintGroup = {
   originDescription: string
 }
 
+/**
+ * Converts footprint primitive port hints into logical component ports.
+ *
+ * Footprints can have multiple copper primitives for one logical pin, for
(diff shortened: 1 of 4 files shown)


lib/components/base-components/NormalComponent/utils/getLogicalPortsFromPortHintGroups.ts
```diff
@@ -0,0 +1,55 @@
+import { Port } from "lib/components/primitive-components/Port"
+import { getPortFromHints } from "lib/utils/getPortFromHints"
+
+type PortHintGroup = {
+  hints: string[]
+  originDescription: string
+}
+
+export function getLogicalPortsFromPortHintGroups(
```
seveibar, line 21: Can you summarize with a jsdoc, im trying to understand what this does, maybe some example inputs/outputs.

seveibar (approved):

PR #2134, opened 2026-04-14
Fix: connector schematic symbol layout

MustafaMulla29:


lib/components/normal-components/Connector.ts
```diff
@@ -120,6 +174,111 @@ export class Connector<
...
+        mergedSchPinStyle,
+        allPinLabels,
+      ),
+      pinCount,
+      schPortArrangement: schPortArrangement ?? undefined,
+      pinLabels: allPinLabels,
+    })
+  }
+
   get config() {
```
seveibar, line 291: We should maybe create a separate class for UsbCConnector 🤔

seveibar (approved):

PR #2136, opened 2026-04-14
Fix: Avoid false missing-footprint error for USB-C connectors

MustafaMulla29:


imrishabh18 (approved):
LGTM! But wait for Seve's review

seveibar (approved):

PR #2139, opened 2026-04-14
Add phased local autorouting by routingPhaseIndex

ShiboSoftwareDev:
Implements phased local autorouting for groups. Traces and nets are grouped by
  routingPhaseIndex, numbered phases route first, and null routes last. After each
  phase, routed trace geometry is added as obstacles for later phases so higher-priority
  routes are preserved.


seveibar (approved):
scared but let's do it

PR #2143, opened 2026-04-16
Emit a warning when manufacturerPartNumber is missing for the connector

MustafaMulla29:


lib/components/normal-components/Connector.ts
```diff
@@ -299,17 +299,28 @@ export class Connector<
...
 
     this.source_component_id = source_component.source_component_id!
+
+    if (props.standard && !manufacturerPartNumber) {
+      db.source_missing_manufacturer_part_number_warning.insert({
+        source_component_id: this.source_component_id,
+        standard: props.standard,
+        subcircuit_id: this.getSubcircuit()?.subcircuit_id ?? undefined,
+        warning_type: "source_missing_manufacturer_part_number_warning",
+        message: `${this.getString()} has standard="${props.standard}" but no manufacturerPartNumber (mfn). Add mfn if you do not want the USB-C part to change in future.`,
```
imrishabh18, line 321:
```suggestion
        message: `${this.getString()} has standard="${props.standard}" but no manufacturerPartNumber (mfn). Add mfn if you do not want the part to change in future.`,
```
Is the connector always usb-c?

MustafaMulla29: Yes currently only usbc is implemented

MustafaMulla29: Although your suggestion sounds good

seveibar (approved):

PR #2144, opened 2026-04-16
  Support implicit PCB ports for composite footprint pins

ShiboSoftwareDev:
This changes repeated non-overlapping portHints from an ambiguity error into multiple pcb_port records tied to the same logical source_port_id.
  This supports single electrical terminals with multiple physical copper features, such as battery clips with multiple pads and plated holes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@tscircuit/core",
   "type": "module",
-  "version": "0.0.1176",
+  "version": "0.0.1177",
   "types": "dist/index.d.ts",
   "main": "dist/index.js",
   "module": "dist/index.js",


lib/components/primitive-components/PlatedHole.ts
```diff
@@ -24,6 +25,10 @@ export class PlatedHole extends PrimitiveComponent<typeof platedHoleProps> {
     }
   }
 
+  private _getMatchedPcbPortId(): string | undefined {
+    return this.matchedPort?.getPcbPortIdForMatchedComponent(this) ?? undefined
```
seveibar: too confusing

seveibar (commented):
a port has multiple port ids? No way- we would never be able to maintain this

seveibar:
This is not an easy problem to solve but we also can't hack it- otherwise we will never be able to fix this codebase. There are other solutions, for example introducing the concept of "internally connected ports"

seveibar:
So a source port can have multiple pcb ports? This also sounds like a hack no?

PR #2148, opened 2026-04-18
Add min via size support

mohan-bee:
This change makes minViaDiameter and minViaHole flow through core properly.

They now work from TSX on <board> and <group>, get stored in Circuit JSON/PCB data, pass into SimpleRouteJson, and are used when routed vias are created as pcb_via elements. It also removes the hardcoded 0.6 via size in phased autorouting so core uses the configured via size instead.


mohan-bee:
waiting for these PR's to merge 

https://github.com/tscircuit/props/pull/634
https://github.com/tscircuit/circuit-json/pull/542

rushabhcodes (commented):
needs seve's approval for this chanage

imrishabh18 (changes requested):
The names are confusing, `minViaHole` doesn't mean anything much.

PR #2159, opened 2026-04-20
fix: prevent duplicate net labels when multiple ports share same net

BossChaos:
Fixes repro61 test case where duplicate net labels were shown when multiple components share the same net (GND/VCC).

## Problem
When two capacitors C1 and C2 both connect to GND and VCC nets, the schematic renderer placed a net label at each connection point, resulting in duplicate labels (2x GND, 2x VCC).

## Solution
Add a `netsWithLabels` Set to track which nets already have a label placed. When processing ports or traces, skip placing a label if that net already has one.

## Changes
- `insertNetLabelsForPortsMissingTrace.ts`: Add netsWithLabels tracking
- `insertNetLabelsForTracesExcludedFromRouting.ts`: Add netsWithLabels tracking

## Test
Repro61 test case now shows only 2 labels (1x GND, 1x VCC) instead of 4.

### DIFF
--- a/lib/components/primitive-components/Group/Group_doInitialSchematicTraceRender/insertNetLabelsForPortsMissingTrace.ts
+++ b/lib/components/primitive-components/Group/Group_doInitialSchematicTraceRender/insertNetLabelsForPortsMissingTrace.ts
@@ -20,6 +20,11 @@ export const insertNetLabelsForPortsMissingTrace = ({
 }) => {
   const { db } = group.root!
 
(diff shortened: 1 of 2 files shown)


rushabhcodes (commented):
proof that this fixes it, iirc this should be fixed in the schematic-trace-solver

BossChaos:
Just checking in — is there anything else needed from my side to move this forward? Happy to make any adjustments! 👍

PR #2160, opened 2026-04-20
min_trace_width value should be 0.1mm

imrishabh18:


seveibar (commented):
i think there's an issue- the trace width is changing but the default nominal trace width should probably be larger

seveibar (approved):
idk the trace width issue could be due to the autorouter not inflating

PR #2162, opened 2026-04-21
Infer internal footprint connections for split pins and shared aliases

ShiboSoftwareDev:
This makes repeated non-overlapping footprint contacts behave like implicit internally connected pins instead of ambiguous PCB targets. Separate physical contact clusters now become separate source ports with one PCB port each, and they are tied together through existing internal-connection machinery with no schema change. Shared non-numeric aliases like A1 also infer the same internal connection, so traces can target the alias directly.

  It also updates pad and plated-hole matching to assign each primitive to the correct inferred port, removes the old ambiguous-port expectations from imported footprint tests, and adds repro coverage with PCB SVG snapshots for split-pin and battery clip cases.


lib/components/base-components/NormalComponent/NormalComponent.ts
```diff
@@ -385,15 +389,31 @@ export class NormalComponent<
...
+              port._parsedProps.pinNumber !== undefined &&
+              p._parsedProps.pinNumber !== port._parsedProps.pinNumber
+                ? false
+                : p.isMatchingAnyOf(port.getNameAndAliases()),
             ) ??
             portsToCreate.find((p) =>
-              p.isMatchingAnyOf(port.getNameAndAliases()),
+              p._parsedProps.pinNumber !== undefined &&
+              port._parsedProps.pinNumber !== undefined &&
+              p._parsedProps.pinNumber !== port._parsedProps.pinNumber
```
seveibar: hmm why is this so messy? I wonder if there's a way to make it cleaner somehow- e.g.

```
const eitherPortIsUndefined = p._parsedProps.pinNumber === undefined ||
              port._parsedProps.pinNumber === undefined
```

lib/components/base-components/NormalComponent/NormalComponent.ts
```diff
@@ -1166,7 +1193,15 @@ export class NormalComponent<
...
   getPortsFromFootprint(opts?: {
     additionalAliases?: Record<string, string[]>
+    collectInferredInternallyConnectedPins?: boolean
   }): Port[] {
+    const logicalPortOpts = {
+      ...opts,
+      inferredInternallyConnectedPinNames:
+        opts?.collectInferredInternallyConnectedPins === true
+          ? this._inferredInternallyConnectedPinNames
+          : undefined,
```
seveibar: can you move this outside of a ternary? I'm thinking about panning ternaries for readability, i get more sensitive about it when the code is tricky to understand

```
let inferredInternallyConnectedPins = undefined
if (opts?. collectInferredInternallyConnectedPins) {
  inferredInternallyConnectedPins = ...
}
```

lib/components/base-components/NormalComponent/NormalComponent.ts
```diff
@@ -1373,7 +1409,9 @@ export class NormalComponent<
 
     const portsFromFootprint = this.getPortsFromFootprint()
     if (portsFromFootprint.length > 0) {
-      return portsFromFootprint.length
+      return portsFromFootprint.filter(
+        (port) => !port._isInferredInternalPhysicalPort,
```
seveibar: interesting, this makes me think we should rename this method `_getPinCount()` to `_getUniqueNetPinCount`, but not a huge issue (i think)- i think we're establishing that a pin may have multiple ports which is acceptable!

lib/components/base-components/NormalComponent/utils/inferInternallyConnectedPinNamesFromPorts.ts
```diff
@@ -0,0 +1,79 @@
...
+  const sortedA = a.toSorted()
+  const sortedB = b.toSorted()
+
+  return (
+    sortedA.length === sortedB.length &&
+    sortedA.every((name, index) => name === sortedB[index])
+  )
+}
+
+const getRawPinName = (port: Port): string | null =>
```
seveibar: not important, but "Raw Pin Name" isn't an obvious concept

```suggestion
const getPinNumberedName = (port: Port): string | null =>
```

lib/components/base-components/NormalComponent/utils/inferInternallyConnectedPinNamesFromPorts.ts
```diff
@@ -0,0 +1,79 @@
+import { Port } from "lib/components/primitive-components/Port"
+
+const hasSameMembers = (a: string[], b: string[]): boolean => {
```
seveibar:
```suggestion
const hasExactSameMembers = (a: string[], b: string[]): boolean => {
```

lib/components/base-components/NormalComponent/utils/inferInternallyConnectedPinNamesFromPorts.ts
```diff
@@ -0,0 +1,79 @@
...
+export function inferInternallyConnectedPinNamesFromPorts(
+  ports: Port[],
+  inferredInternallyConnectedPinNames: string[][],
+): void {
+  const portsByAlias = new Map<string, Port[]>()
+
+  for (const port of ports) {
+    for (const alias of new Set(port.getNameAndAliases())) {
+      if (/^(pin)?\d+$/.test(alias)) continue
+      if (alias.startsWith("__")) continue
```
seveibar: what's an example of a "__" in a pin alias? Just trying to make sure we're not hacking strings

lib/components/primitive-components/Port/Port.ts
```diff
@@ -39,6 +39,8 @@ export class Port extends PrimitiveComponent<typeof portProps> {
 
   schematicSymbolPortDef: SchSymbol["ports"][number] | null = null
   matchedComponents: PrimitiveComponent[]
+  _isInferredInternalPhysicalPort = false
```
seveibar: I don't understand this

Can you clarify:

1. What does it mean to be inferred
2. What does it mean to be an internal physical port?

I _think_ this could be renamed:

```
_isSecondaryPort = false
```

This is an interesting idea, to have primary and secondary ports.

I don't know if `inferred` is necessary as a concept here.

Also you could simplify this by just have `_isPrimaryReference` and `_primaryPortNumber`

This concept of a "Primary Reference" is interesting, the primary port takes precedent. This is good conceptually

seveibar (commented):
Yea this is a good concept, i just have some comments then i'd like to re-review, nice work!

lib/components/base-components/NormalComponent/NormalComponent.ts
```diff
@@ -257,8 +260,23 @@ export class NormalComponent<
     } = {},
   ) {
     if (this.root?.schematicDisabled) return
+    this._inferredInternallyConnectedPinNames = []
     const { config } = this
     const portsToCreate: Port[] = []
+    const canMergePortDefinitions = (a: Port, b: Port) => {
```
seveibar: named closures often have scope leaking, but ig this is ok (named closures should be put into separate files as an imported function generally)

lib/components/base-components/NormalComponent/NormalComponent.ts
```diff
@@ -413,7 +435,7 @@ export class NormalComponent<
     // Add ports that we know must exist because we know the pin count and
     // missing pin numbers, and they are inside the pins array of the
     // schPortArrangement
-    const requiredPinCount = opts.pinCount ?? this._getPinCount() ?? 0
+    const requiredPinCount = opts.pinCount ?? this._getLogicalPinCount() ?? 0
```
```suggestion
    const requiredPinCount = opts.pinCount ?? this._getUniqueNetPinCount() ?? 0
```

lib/components/base-components/NormalComponent/NormalComponent.ts
```diff
@@ -1166,7 +1195,18 @@ export class NormalComponent<
...
     additionalAliases?: Record<string, string[]>
+    collectInferredInternallyConnectedPins?: boolean
   }): Port[] {
+    let inferredInternallyConnectedPinNames: string[][] | undefined = undefined
+    if (opts?.collectInferredInternallyConnectedPins) {
+      inferredInternallyConnectedPinNames =
+        this._inferredInternallyConnectedPinNames
+    }
+
+    const logicalPortOpts = {
```
seveibar: what does it mean to be logical?

lib/components/base-components/NormalComponent/NormalComponent.ts
```diff
@@ -1360,7 +1401,7 @@ export class NormalComponent<
     )
   }
 
-  _getPinCount(): number {
+  _getLogicalPinCount(): number {
```
```suggestion
  _getUniqueNetPinCount(): number {
```

seveibar: OOOOH OR _getPrimaryPinCount() (that could be good!)

lib/components/base-components/NormalComponent/utils/getLogicalPortsFromPortHintGroups.ts
```diff
@@ -20,10 +24,16 @@ type PortHintGroup = {
  */
 export function getLogicalPortsFromPortHintGroups(
```
seveibar:
```suggestion
export function getPrimaryPortsFromPortHintGroups(
```

seveibar (commented):
do my suggestions work? I'm thinking Logical -> Primary? That makes wayyy more sense to me?

seveibar (approved):

PR #2165, opened 2026-04-21
Wire footprint insertion direction into PCB components and align board DRC fields

ShiboSoftwareDev:
Propagate footprint.insertionDirection onto pcb_component.insertion_direction using the component’s global pre-layout rotation plus bottom-side mirroring, and make primitive footprint flipping respect originalLayer so bottom-authored footprints and footprint constraints behave correctly. Bump @tscircuit/circuit-json-util to rely on its built-in PCB component transform updates.

  Also align board manufacturing defaults and tests with the new *Clearance field names in circuit-json, replace the stale JLC tolerance import with a local helper, and add regression coverage for transformed insertion directions and originalLayer="bottom" footprint layout.


lib/components/primitive-components/Group/Group_doInitialPcbLayoutPack/applyPackOutput.ts
```diff
@@ -212,6 +223,15 @@ export const applyPackOutput = (
     }
 
     transformPCBElements(relatedElements as any, transformMatrix)
+    for (const elm of relatedElements) {
+      if (elm.type === "pcb_component") {
+        updatePcbInsertionDirection({
```
seveibar: transformPcbElements should perform this operation automatically

seveibar (commented):
I think transformPcbElements should automatically perform this

seveibar (commented):
Can you take a look over this, you removed the ysage of the jlc spec module

seveibar (approved):
Undo jlc spec modifications

PR #2191, opened 2026-04-27
Unconnected obstacles outside the board outline are ignored and the board outline is not inflated

imrishabh18:


seveibar (commented):
this fix should be in the autorouter, not core

PR #2193, opened 2026-04-27
implement a fix for 45 degree rects

Sang-it:


lib/utils/obstacles/getObstaclesFromCircuitJson.ts
```diff
@@ -88,21 +85,17 @@ export const getObstaclesFromCircuitJson = (
...
-            connectedTo: withNetId([element.pcb_smtpad_id]),
-          })
-        }
+        obstacles.push({
+          type: "rect",
+          layers: [element.layer],
+          center: rect.center,
+          width: rect.width,
+          height: rect.height,
+          ...(singleRect ? {} : { ccwRotationDegrees: element.ccw_rotation }),
```
seveibar, line 96: bad ternary, doesn't make sense. singleRect does not imply uprightRect

lib/utils/obstacles/getObstaclesFromCircuitJson.ts
```diff
@@ -1,10 +1,7 @@
 import { getObstaclesFromRoute } from "./getObstaclesFromRoute"
 import type { ConnectivityMap } from "circuit-json-to-connectivity-map"
 import type { AnyCircuitElement } from "circuit-json"
-import {
-  generateApproximatingRects,
```
seveibar, line 5: is this function still used anywhere?

seveibar (changes requested):
This will break the autorouter- (that's why traces changed) there should be no snapshot updates- you need to port the approximating rect implementation over to the autorouter prior to making this change

PR #2197, opened 2026-04-28
Add schematic box rendering for groups

ShiboSoftwareDev:
Implements showAsSchematicBox by rendering collapsed groups as regular box-style schematic_components, with direct group ports exposed as schematic pins and internal schematic elements suppressed.


lib/components/base-components/PrimitiveComponent/PrimitiveComponent.ts
```diff
@@ -576,6 +576,20 @@ export abstract class PrimitiveComponent<
     return null
   }
 
+  getCollapsedSchematicBoxAncestor(): PrimitiveComponent | null {
```
seveibar, line 584: can you maybe add a docstring here, i think i understand it, it's the first parent/ancestor that has showAsSchematicBox

seveibar (approved):
wow

tests/groups/group-schematic-box.test.tsx
```diff
@@ -24,19 +24,32 @@ test("group schematic box", () => {
...
           "y": 0,
         },
-        "height": 0,
+        "height": 0.4,
         "is_subcircuit": undefined,
         "name": "G1",
-        "schematic_component_ids": [],
+        "schematic_component_ids": [
+          "schematic_component_0",
+        ],
```
imrishabh18, line 32: Shouldn't both the components's id be present?

lib/components/primitive-components/Group/Group_doInitialSchematicBoxComponentRender.ts
```diff
@@ -0,0 +1,127 @@
...
+    source_group_id: group.source_group_id!,
+    schematic_group_id: group.schematic_group_id ?? undefined,
+    subcircuit_id: group.subcircuit_id ?? group.getSubcircuit()?.subcircuit_id,
+    is_schematic_group: true,
+    is_box_with_pins: true,
+    port_arrangement: underscorifyPortArrangement(schPortArrangement!),
+    pin_spacing: props.schPinSpacing ?? 0.2,
+    pin_styles: underscorifyPinStyles(props.schPinStyle, portLabels),
+    port_labels: portLabels,
+  } as any)
```
imrishabh18:
```suggestion
  })
```

lib/components/primitive-components/Group/Group.ts
```diff
@@ -1009,7 +1073,12 @@ export class Group<Props extends z.ZodType<any, any, any> = typeof groupProps>
     })
     this.schematic_group_id = schematic_group.schematic_group_id
 
+    if (props.showAsSchematicBox) {
+      Group_doInitialSchematicBoxComponentRender(this as any)
```
imrishabh18:
```suggestion
      Group_doInitialSchematicBoxComponentRender(this)
```

imrishabh18 (approved):
Nice work!

tests/groups/group-schematic-box.test.tsx
ShiboSoftwareDev: @imrishabh18 this is the schematic box we are going to actually draw not the internal ones

ShiboSoftwareDev: we do not keep the internal schematic components because otherwise circuit-to-svg will draw them

imrishabh18: Ok

PR #2198, opened 2026-04-28
Fix Connector binding for usb_c footprint resolution

MustafaMulla29:


lib/components/normal-components/Connector.ts
```diff
@@ -118,35 +118,38 @@ export class Connector<
   }
 
   private async _tryFetchPartCircuitJson(
-    fetchPartCircuitJson: NonNullable<PartsEngine["fetchPartCircuitJson"]>,
+    partsEngine: PartsEngine & {
+      fetchPartCircuitJson: NonNullable<PartsEngine["fetchPartCircuitJson"]>
```
seveibar, line 122: Why is this necessary? This is a hack?

seveibar (commented):
this seems like a hack?

MustafaMulla29:
@seveibar `this` context was getting lost, this pr was trying to do partsEngine.fetchPartCircuitJson. what could be a better pattern?

PR #2202, opened 2026-04-28
Add `supplier_footprint_mismatch_warning` when footprint string does not match with the jlcpcb footprint

imrishabh18:


lib/components/base-components/NormalComponent/NormalComponent.ts
```diff
@@ -1865,6 +1867,17 @@ export class NormalComponent<
     }
   }
 
+  doInitialSupplierFootprintMismatchWarning(): void {
+    NormalComponent_doInitialSupplierFootprintMismatchWarning(
+      this,
+      (name, effect) => this._queueAsyncEffect(name, effect),
```
seveibar: no second arguments for these methods

lib/components/base-components/NormalComponent/NormalComponent_doInitialSupplierFootprintMismatchWarning.ts
```diff
@@ -0,0 +1,191 @@
...
+}
+
+const supplierNames: SupplierName[] = [
+  "jlcpcb",
+  "macrofab",
+  "pcbway",
+  "digikey",
+  "mouser",
+  "lcsc",
+]
```
seveibar: can you import this from props? Is this the best place to define this?

seveibar (approved):
see comment

PR #2205, opened 2026-04-29
Add async manifold initialization phase for copper pours

seveibar:
## Summary
- Add a dedicated `InitializeAsyncModules` render phase and make copper pour rendering wait on it.
- Initialize `manifold-3d` once per process through `CopperPour`, with shared in-flight state and cached failure handling.
- Keep `CopperPourPipelineSolver` in `SOLVERS` and surface clear render-time errors when manifold initialization or solver setup fails.

## Testing
- Focused unit tests for copper pour rendering and solver event emission passed.
- Typecheck and package build completed successfully.

### DIFF
--- a/lib/components/base-components/Renderable.ts
+++ b/lib/components/base-components/Renderable.ts
@@ -19,6 +19,7 @@ export const orderedRenderPhases = [
   "FetchPartFootprint",
   "InitializePortsFromChildren",
   "CreateNetsFromProps",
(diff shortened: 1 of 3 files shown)


seveibar:
@ShiboSoftwareDev @AnasSarkiz actually the old code is fine- do we just need to add manifold-3d as a tscircuit dependency?

PR #2207, opened 2026-04-29
Fix: incorrect prop value autorouter was being used for the trace input to simpleRouteJson

imrishabh18:
Update the autorouter to the latest


lib/components/primitive-components/Group/Group.ts
```diff
@@ -643,8 +643,8 @@ export class Group<Props extends z.ZodType<any, any, any> = typeof groupProps>
     const { simpleRouteJson: baseSimpleRouteJson } =
       getSimpleRouteJsonFromCircuitJson({
         db,
-        minTraceWidth: this.props.autorouter?.minTraceWidth ?? 0.15,
-        nominalTraceWidth: this.props.nominalTraceWidth,
+        minTraceWidth: Number(props.minTraceWidth ?? 0.15),
+        nominalTraceWidth: Number(props.nominalTraceWidth ?? 0.15),
```
imrishabh18:
```suggestion
        nominalTraceWidth: Number(props.nominalTraceWidth),
```

PR #2230, opened 2026-05-05
Normalize generic resistor footprints to resistor-specific packages

techmannih:


seveibar (changes requested):
keep resistor specific features in resistor etc. this is a hack

seveibar:
you shouldn't need to modify NormalComponent for this AFAIK

PR #2231, opened 2026-05-05
Add `minViaEdgeToPadEdgeClearance` in the srj input

imrishabh18:


seveibar (commented):
this doesn't make sense-

seveibar (approved):
oh nvm

seveibar (commented):
wait no what is happening

lib/utils/autorouting/SimpleRouteJson.ts
```diff
@@ -71,6 +71,7 @@ export interface SimpleRouteJson {
   min_via_pad_diameter?: number
 
   minTraceToPadEdgeClearance?: number
+  minPadEdgeToPadEdgeClearance?: number
```
seveibar: why would this be an SRJ input?

seveibar (changes requested):
pad edge to pad edge clearance isn't related to SRJ?

lib/utils/autorouting/SimpleRouteJson.ts
imrishabh18: How would the `globalDrcForceImproveSolver ` know about the minimum requirement of clearance?

PR #2236, opened 2026-05-05
Add autorouting phase component support

ShiboSoftwareDev:
Adds a self-closing <autoroutingphase /> primitive for configuring autorouter behavior per existing routingPhaseIndex phase. Routing phase plans now attach matching phase autorouter configs, local autorouting uses those configs per phase, and net-derived trace phases reuse the existing net phase lookup.


seveibar (approved):
MY MAN

PR #2237, opened 2026-05-05
Fix named schPinArrangement ports for group schematic boxes

ShiboSoftwareDev:
Allow group schematic boxes to use named child ports in schPinArrangement before child port labels are available. Centralizes explicit arrangement pin extraction with getPinsFromPortArrangement, handles string pins when computing pin counts, and adds coverage for side placement/facing behavior.


tests/groups/group-schematic-box-sch-port-arrangement.test.tsx
```diff
@@ -0,0 +1,55 @@
...
+          },
+          rightSide: {
+            pins: ["VOUT"],
+            direction: "top-to-bottom",
+          },
+        }}
+      >
+        <port name="VIN" direction="right" />
+        <port name="GND" direction="right" />
+        <port name="VOUT" direction="left" />
```
seveibar, line 24: can you make sure these are optional if the schPinArrangement is provided as well?

seveibar (approved):

PR #2244, opened 2026-05-06
Normalize resistor footprints to resistor-specific packages

techmannih:
## Summary

This PR keeps generic passive footprint normalization scoped to `Resistor` instead of moving that behavior into `NormalComponent`.

## What changed

- normalize generic resistor footprints like `0402`, `0603`, and `1206` to resistor-specific packages such as `res0402`
- keep capacitor and LED footprints unchanged
- refresh affected PCB and 3D snapshots, plus the updated inline snapshot in `repro44`


seveibar (changes requested):
We dont have any code with a pattern like this

PR #2247, opened 2026-05-06
fix(NormalComponent): respect pcbPositionMode prop in pcb_component output

### LINKED ISSUE #2: MVP TODOs
- [x] PCB Traces
- [ ] PCB Trace Hinting
- [ ] Schematic Traces
- [ ] PCB Rotation for Primitives
- [ ] PCB Group Rotation
- [ ] Group
- [ ] Group PCB & Schematic Translation
- [ ] Various Small Elements (Diode, Inductor etc.)

### LINKED ISSUE #2242: pcbPositionMode prop typechecks but has no runtime effect — position_mode stays 'packed'
## Summary

Setting `pcbPositionMode="relative_to_board_anchor"` (or any of the other `PcbPositionMode` enum values: `"auto"`, `"relative_to_group_anchor"`, `"relative_to_component_anchor"`) on a chip/component in JSX has **no observable runtime effect**. The resulting `pcb_component`'s `position_mode` field stays `"packed"` (the auto-placer's collision-avoidance mode), and the component lands wherever the placer puts it — not at the requested `pcbX/pcbY`.

The prop typechecks via `PcbLayoutProps.pcbPositionMode` in `@tscircuit/props`, so users assume it works. The runtime silently ignores it.

## Repro

```tsx
// index.circuit.tsx — board 22 × 20 mm, U1 should be at top edge
export default () => (
  <board width="22mm" height="20mm" layers={2}>
    <chip
      name="U1"
      footprint="dip8"
      pcbX={0}
      pcbY={4}
      pcbPositionMode="relative_to_board_anchor"  // ← typechecks but no-op
    />
  </board>
)
```

```sh
tsci build
jq '.[] | select(.type=="pcb_component") | {center, position_mode}' dist/index/circuit.json
# → { "center": { "x": 0, "y": 0 }, "position_mode": "packed" }
# Expected: center.y near 4, position_mode "relative_to_board_anchor"
```

## Workaround

Add `pcbX={0} pcbY={0}` to the parent `<group>`. That presence-of-prop (regardless of value) flips the *group* into `"relative_to_group_anchor"` mode, and child `pcbX/pcbY` are then honoured. Discovered by JSON inspection on a real board after burning ~5 turns of debugging.

```tsx
<board ...>
  <group pcbX={0} pcbY={0}>      {/* ← required */}
    <chip name="U1" pcbX={0} pcbY={4} />  {/* now lands at (0, 4) */}
  </group>
</board>
```

## Why this matters

Component-level `pcbPositionMode` is the documented escape hatch for "place this exactly where I asked, ignore the auto-placer." When it silently doesn't work, there's no error, no warning — the component just shows up somewhere else. Without JSON inspection (`jq` against `circuit.json`), the failure mode is "I set pcbY=4 and the chip is at y=0 and I don't know why."

## Suggested fix (any of)

1. **Implement** the prop — wire it through to the placer so `relative_to_board_anchor` actually skips the packed-mode collision-avoidance for that component.
2. **Or warn**: at JSX-prop-parse time, log a `[tscircuit] pcbPositionMode is set on a component but is not yet implemented; positioning will fall back to "packed" mode unless the parent group has explicit pcbX/pcbY.`
3. **Or remove from typings** until implemented, so TypeScript catches the use.

Hit while placing an InsightSiP ISP3080-UX module on a daughter board — wanted U1 flush against the top edge. The pcbPositionMode prop on the chip seemed like the obvious knob; took several turns of bisection to find the group-level workaround.

gsdali:
## Summary

Closes #2242.

The `pcbPositionMode` prop was declared in `@tscircuit/props`'s `PcbLayoutProps` and typechecked, but the runtime never consumed it — `NormalComponent._initialPcbDisplayOffset` always stamped `position_mode: "relative_to_group_anchor"`, regardless of what the user requested. Setting `pcbPositionMode="relative_to_board_anchor"` on a chip produced no observable change in the circuit JSON, AND no change in actual placement.

This PR fixes both halves of the problem. (Two commits — landable as one.)

## Fix #1 — respect the prop in `pcb_component.position_mode`

`NormalComponent._initialPcbDisplayOffset` now uses `this._parsedProps.pcbPositionMode ?? "relative_to_group_anchor"` when stamping `position_mode`. The user's prop value is honoured if set; the default is unchanged.

`NormalComponent.isRelativelyPositioned()` also returns true when `pcbPositionMode` is set — explicit positioning intent (defaulting to `(0, 0)`) the auto-placer should treat as static.

## Fix #2 — cascade through ungrouped parents

When a chip declares `pcbPositionMode="relative_to_board_anchor"`, the auto-placer must not shift its parent group as a unit — otherwise the chip's "absolute" coordinates end up shifted by whatever offset the parent's pack assigned to the group.

`Group.isRelativelyPositioned()` now overrides the inherited check to also return true when any descendant component has `pcbPositionMode="relative_to_board_anchor"`. The parent's pack solver then adds the group's `pcb_group_id` to `excludedPcbGroupIds` and pulls all descendants into `staticPcbComponentIds`, leaving the chip at its requested board-absolute coordinates.

`Group._hasDescendantWithBoardAnchorPositioning()` walks the children recursively. Not memoised — children populate during render-tree construction, so a stale-but-cached `false` is worse than re-walking on each call.

## Test plan

- [x] `tests/components/pcb/pcb-position-mode-prop-respected.test.tsx` — two tests:
   - `pcbPositionMode="relative_to_board_anchor"` reflected in `pcb_component.position_mode`
   - prop unset preserves the historical default
- [x] `tests/components/pcb/pcb-position-mode-cascade.test.tsx` — two tests:
   - chip with `pcbPositionMode='relative_to_board_anchor'` inside a group with no `pcbX/pcbY` lands at its requested coords (the daughter-board scenario from the issue)
   - siblings without `pcbPositionMode` still get auto-placed (cascade is opt-in)
- [x] Full-suite delta vs `main`: same flaky-test profile (1 fail in `tests/repros/repro81-panel-unnamed-pcb-group.test.tsx` which flakes under full-suite load on `main` too — passes in isolation; not caused by this change).

## Scope

| Scenario | Status |
|---|---|
| Component-level `pcbPositionMode` reflected in output | Fixed |
| Component-level `isRelativelyPositioned()` includes `pcbPositionMode` | Fixed |
| Single-level cascade (chip in a group with no `pcbX/pcbY`) | Fixed (daughter-board case) |
| Backward compat for components without the prop | Preserved |
| Nested-group cascade (chip > inner > outer > board) | Deferred — requires deeper changes to `applyPackOutput`'s cluster handling |

Documented the deferred case in the commit message; happy to spin up a follow-up if maintainers want to tackle it.

### DIFF
--- a/lib/components/base-components/NormalComponent/NormalComponent.ts
+++ b/lib/components/base-components/NormalComponent/NormalComponent.ts
@@ -2085,7 +2085,12 @@ export class NormalComponent<
             : undefined)
 
     db.pcb_component.update(this.pcb_component_id, {
-      position_mode: "relative_to_group_anchor",
+      // Honour an explicit `pcbPositionMode` prop if the user set one;
+      // otherwise default to relative_to_group_anchor (the historical
+      // behaviour). This makes the prop have an observable effect on
+      // the resulting pcb_component, which it previously did not.
+      position_mode:
+        this._parsedProps.pcbPositionMode ?? "relative_to_group_anchor",
       positioned_relative_to_pcb_group_id: positionedRelativeToGroupId,
       positioned_relative_to_pcb_board_id: positionedRelativeToBoardId,
       display_offset_x: resolvedPcbX as any,
@@ -2124,6 +2129,10 @@ export class NormalComponent<
       this._parsedProps.pcbRightEdgeX !== undefined ||
       this._parsedProps.pcbTopEdgeY !== undefined ||
       this._parsedProps.pcbBottomEdgeY !== undefined ||
+      // An explicit pcbPositionMode counts as evidence that the user
+      // intends absolute positioning, even if the coordinates default
+      // to (0, 0). The packer should treat the component as static.
+      this._parsedProps.pcbPositionMode !== undefined ||
       rawProps.pcbLeftEdgeX !== undefined ||
       rawProps.pcbRightEdgeX !== undefined ||
       rawProps.pcbTopEdgeY !== undefined ||
--- a/lib/components/primitive-components/Group/Group.ts
+++ b/lib/components/primitive-components/Group/Group.ts
@@ -74,6 +74,61 @@ export class Group<Props extends z.ZodType<any, any, any> = typeof groupProps>
   schematic_group_id: string | null = null
   subcircuit_id: string | null = null
 
+  /**
+   * Override of `NormalComponent.isRelativelyPositioned()` that also
+   * returns true if any descendant component has
+   * `pcbPositionMode="relative_to_board_anchor"`.
+   *
+   * Why: when a deeply-nested chip declares board-absolute positioning,
+   * none of its parent groups should be moved by the auto-placer —
+   * otherwise the chip's "absolute" coordinates end up shifted by
+   * whatever offset the parent's pack assigned to the group as a unit.
+   * Adding the cascade here marks every ancestor group up to the board
+   * as static-for-packer, so the chip lands at the requested
+   * (pcbX, pcbY) relative to the board.
+   */
+  override isRelativelyPositioned(): boolean {
+    if (super.isRelativelyPositioned()) return true
+    return this._hasDescendantWithBoardAnchorPositioning()
+  }
+
+  private _boardAnchorDescendantCache: boolean | null = null
+  /**
+   * Recursively check children for a component with
+   * `pcbPositionMode="relative_to_board_anchor"`. Memoised per-instance
+   * so repeated calls during a single pack pass don't re-walk the tree.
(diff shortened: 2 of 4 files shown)


gsdali:
Bumping per the stale-bot tag — this fix is still in our active patch set on a downstream board (locally `bun patch`'ed against @tscircuit/core 0.0.1215). All 10 CI checks pass and the PR is mergeable / clean. Happy to address review comments or split the cascade-through-ungrouped-parents change out into a separate PR if you'd prefer the single-level fix to land first.

PR #2248, opened 2026-05-06
fix(schematic): respect schTraceAutoLabelEnabled={false} to suppress port-net auto-labels

### LINKED ISSUE #2243: schTraceAutoLabelEnabled={false} has no observable effect on schematic rendering
## Summary

Setting `schTraceAutoLabelEnabled={false}` on either a `<group>` or `<board>` element has **no observable effect** on the schematic output. The schematic placer continues to emit `schematic_net_label` elements for every node on a named net, regardless of the prop's value. Setting it on the parent group, the board, or both produces identical SVG output.

The prop typechecks via `SubcircuitGroupProps.schTraceAutoLabelEnabled` (described as *"If true, net labels will automatically be created for complex traces"*), so users assume `false` will suppress them. It doesn't.

## Repro

```tsx
export default () => (
  <board width="20mm" height="20mm" schTraceAutoLabelEnabled={false}>
    <chip
      name="U1"
      footprint="soic8"
      pinLabels={{ pin1: "VCC", pin2: "GND" }}
      connections={{ VCC: "net.V3_3", GND: "net.GND" }}
    />
    <resistor
      name="R1"
      resistance="10k"
      footprint="0402"
      connections={{ pin1: "net.V3_3", pin2: "net.GND" }}
    />
  </board>
)
```

```sh
tsci build
jq '[.[] | select(.type=="schematic_net_label")] | length' dist/index/circuit.json
# → ≥ 4  (V3_3 labels at U1.VCC, R1.pin1; GND labels at U1.GND, R1.pin2)
# Expected: 0 with schTraceAutoLabelEnabled={false}
```

The same script with the prop removed produces the same number of labels.

## Why this matters

Named-net rendering as labels (instead of drawn wires) is the **canonical** tscircuit schematic style and matches commercial schematic conventions. So most users won't want to disable it. But:

1. Users *find* this prop in the docs / type definitions and assume it does what it says.
2. Some specific cases (single-page schematics with only a few power connections, certain didactic examples) genuinely benefit from drawn wires over labels — and the prop existing implies the override is supported.
3. Silent no-ops are confusing; users assume the prop's value is fighting their other settings.

## Suggested fix (any of)

1. **Implement** — when `false`, force the placer to draw `schematic_trace` elements for all source_traces on the affected group/board, even when the source_trace's net has multiple references. (Acknowledging that this can cause visual spaghetti for complex power nets — that's the user's choice.)
2. **Or warn**: log when the prop is set but not yet implemented.
3. **Or remove from typings** until implemented.

Hit while trying to make a decoupling cap chain visibly wire to the chip pin on a schematic — the cap chain shared `net.V3_3` with the chip, so the placer rendered both endpoints as labels rather than drawing a connecting wire. Tried `schTraceAutoLabelEnabled={false}` as a documented escape hatch; no change. Eventually accepted that named nets rendering as labels is by design and there's no per-trace override — but the existence of this prop suggested otherwise.

Related: this gap is now documented in our project's [`docs/tscircuit-diagnostic-playbook.md`](https://github.com/gsdali/wanman-electronics/blob/main/docs/tscircuit-diagnostic-playbook.md) so future sessions don't burn time on it.

gsdali:
## Summary

Closes #2243.

`schTraceAutoLabelEnabled` is declared as a tri-state prop in the type definitions, but only the `true` case was wired up at runtime — for opting INTO net labels at complex multi-junction traces (in `Trace_doInitialSchematicTraceRender.ts`). Setting the prop to `false` had no observable effect, because the only consumer treats falsy (undefined or false) as "not opted in".

The user-facing pain: named-net connections (every chip pin tied to `net.V3_3` / `net.GND` / etc.) automatically get a `schematic_net_label` drawn at every endpoint — the canonical "label, not wire" rendering. Users wanted a way to opt out and got none.

## Fix

Wire `schTraceAutoLabelEnabled === false` (explicit, not undefined) into `Group_doInitialSchematicTraceRender` so that `insertNetLabelsForPortsMissingTrace` is skipped when the user explicitly opts out. Both call sites (early-return when no routable connections, and post-routing) honour the opt-out.

### Tri-state semantics

| value | behaviour |
|---|---|
| `undefined` (default) | existing behaviour, auto labels emitted |
| `true` | opt INTO labels for complex traces (existing, in `Trace_doInitialSchematicTraceRender`) |
| `false` | opt OUT of port-net auto-labels (NEW) |

## Test plan

- [x] New `tests/components/sch/sch-trace-autolabel-disabled.test.tsx`:
   - `false` → `0` `schematic_net_label` records for a chip on `net.V3_3` / `net.GND`
   - unset (default) → `≥1` `schematic_net_label` (existing behaviour preserved)
- [x] `bun test tests/components/sch`: 2/2 pass
- [x] Full-suite delta vs `main`: same flaky-test profile (1 fail in `tests/repros/repro81-panel-unnamed-pcb-group.test.tsx` which flakes under full-suite load on `main` too — passes in isolation; not caused by this change)

## What this PR does NOT solve

- The `true` case behaviour is unchanged from current semantics (only opts into complex-trace labels).
- This PR only suppresses labels at port-missing-trace junctions (the most common user pain). Other auto-label sources (e.g. labels emitted by the trace solver inside `applyNetLabelPlacements`) are not affected. Reasonable next step is to extend the opt-out to those paths too, but it'd touch more code and is left as follow-up.

## Why this scope

The most surprising part of the original report was *the prop has zero observable effect when set to false*. After this PR, setting `false` produces an immediately-visible difference (no more port-net auto-labels). Users who want "named nets render as wires, not labels" still need other mechanisms (port-to-port connections via array `connections`), but at least they have one knob that makes a visible difference.

### DIFF
--- a/lib/components/primitive-components/Group/Group_doInitialSchematicTraceRender/Group_doInitialSchematicTraceRender.ts
+++ b/lib/components/primitive-components/Group/Group_doInitialSchematicTraceRender/Group_doInitialSchematicTraceRender.ts
@@ -19,6 +19,17 @@ export const Group_doInitialSchematicTraceRender = (group: Group<any>) => {
   if (!group.isSubcircuit) return
   if (group.root?.schematicDisabled) return
 
+  // `schTraceAutoLabelEnabled` is a tri-state:
+  //   - undefined (default): existing behaviour
+  //   - true : opt INTO net labels for complex traces (handled
+  //            in Trace_doInitialSchematicTraceRender)
+  //   - false: opt OUT of auto-generated net labels for ports
+  //            connected to a net via no explicit trace. Useful
+  //            when you'd rather see drawn wires (or accept
+  //            visually disconnected ports) over auto labels.
+  const autoLabelExplicitlyDisabled =
+    group._parsedProps.schTraceAutoLabelEnabled === false
+
   // Prepare the solver input and context
   const {
     inputProblem,
@@ -42,14 +53,16 @@ export const Group_doInitialSchematicTraceRender = (group: Group<any>) => {
     inputProblem.netConnections.length > 0
 
   if (!hasRouteableSchematicConnections) {
-    insertNetLabelsForPortsMissingTrace({
-      group,
-      allSourceAndSchematicPortIdsInScope,
-      schPortIdToSourcePortId,
-      sckToSourceNet,
-      pinIdToSchematicPortId,
-      schematicPortIdsWithPreExistingNetLabels,
-    })
+    if (!autoLabelExplicitlyDisabled) {
+      insertNetLabelsForPortsMissingTrace({
+        group,
+        allSourceAndSchematicPortIdsInScope,
+        schPortIdToSourcePortId,
+        sckToSourceNet,
+        pinIdToSchematicPortId,
+        schematicPortIdsWithPreExistingNetLabels,
+      })
+    }
     return
   }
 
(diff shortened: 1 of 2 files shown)


gsdali:
Bumping per the stale-bot tag — happy to address review comments. PR is mergeable / clean (10/10 checks). Use case: tscircuit's named-net schematic style means every net-connected pin gets a label, which is sometimes too noisy for a first-pass look at routed power/ground. This opt-out keeps the default behaviour while making the cosmetic suppression possible without ripping out per-trace named nets.

PR #2257, opened 2026-05-07
Fix invalid Circuit JSON routes from through-obstacle autorouting

### LINKED ISSUE #290: NaN in place of coordinates is causing it to fail.
files: https://tscircuit.com/0hmX/diy-power-delivery-trigger

Image

mohan-bee:
closes https://github.com/tscircuit/circuit-json-to-kicad/issues/290

This fixes a routing export bug where internal through_pad route points could leak into pcb_trace.route.

through_pad uses start and end, but Circuit JSON traces need normal wire/via route points with top-level x and y. When the internal shape leaked, KiCad export could produce NaN coordinates.

This change converts through_pad points before writing pcb_trace:

creates a via at the start point
adds a wire on the target layer when needed
leaves existing wire and via points unchanged
A repro test was added to make sure rendered autorouter output writes valid pcb_trace routes.


lib/utils/autorouting/mergeRoutes.ts
```diff
@@ -1,9 +1,45 @@
...
+        x: point.end.x,
+        y: point.end.y,
+        width: point.width,
+        layer: point.to_layer as LayerRef,
+      })
+    }
+
+    return circuitJsonRoutePoints
+  })
+}
```
seveibar, line 41: this is incorrect, should be `through_pad`

seveibar (changes requested):
see comment

seveibar (changes requested):
didn't address my comments

PR #2260, opened 2026-05-07
When reroute autoroutingphase is used without phaseIndex, add a default autorouting phase before

imrishabh18:


seveibar (approved):
Not exactly what i had in mind, the jsx order should be the phase order if its unspecified, the implementation will then worker for multiple phases.

seveibar:
I would probably reimplement w the implied jsx order but you can revert this pr in that pr if it’s easier

implied jsxOrder is always higher than the highest userDefinedPhaseOrder

PR #2262, opened 2026-05-07
Adds support for preserving imported pcb_trace elements during circuitJson inflation

imrishabh18:


seveibar (commented):
can you get a real circuitJson repro? This doesn't seem like it'll work with the real kicad data

seveibar (approved):

PR #2288, opened 2026-05-11
Add explicit show_as_bounding_box to cad_component when there is no cad model

MustafaMulla29:
This adds support for explicit fallback CAD components using cad_component.show_as_bounding_box.

Core now creates a cad_component in three cases:

  1. Explicit CAD model
     <chip cadModel={{ glbUrl: "..." }} />Emits a normal cad_component with the model URL.
     

  2. <chip footprint="soic8" />
      Emits a normal cad_component with
     footprinter_string.

  3. Fallback bounding box
     <chip name="U1" />

     If no CAD model and no footprinter
     CAD source exists, core emits:

     {
       "type": "cad_component",
       "show_as_bounding_box": true
     }

This tells renderers to draw a generic 3D box using the associated pcb_component bounds.

Core does not create a fallback cad_component when CAD is explicitly disabled:

  <chip name="U1" cadModel={null} />

  That still emits no cad_component.

This makes fallback boxes explicit in Circuit JSON instead of requiring renderers like circuit-json-to-gltf to infer them from bare pcb_components.

### DIFF
--- a/lib/components/base-components/NormalComponent/NormalComponent.ts
+++ b/lib/components/base-components/NormalComponent/NormalComponent.ts
@@ -1670,7 +1670,7 @@ export class NormalComponent<
     const isBottomLayer = computedLayer === "bottom"
 
     if (!cadModel && !footprint) {
-      const cad_model = db.cad_component.insert({
+      const cad_component = db.cad_component.insert({
         position: {
           x: bounds.center.x,
           y: bounds.center.y,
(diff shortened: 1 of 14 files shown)


MustafaMulla29:
@seveibar what changed in circuit-json? type check is failing

lib/components/base-components/NormalComponent/NormalComponent.ts
```diff
@@ -1670,6 +1669,32 @@ export class NormalComponent<
         : preLayoutRotation
     const isBottomLayer = computedLayer === "bottom"
 
+    if (!cadModel && !footprint) {
+      const cad_model = db.cad_component.insert({
```
seveibar: Bad naming will definitely create confusion!

```suggestion
      const cad_component = db.cad_component.insert({
```

seveibar (approved):
See important comment

PR #2296, opened 2026-05-12
Fix invalid source trace IDs in region reroute output

ShiboSoftwareDev:
Parses reroute output source trace references from _reroute_ and merged __ connection
  names, then only assigns pcb_trace.source_trace_id when it resolves to exactly one real
  source trace. Reuses the parser for existing trace replacement cleanup and updates the
  Arduino Uno invalid-source-id repro to assert the fixed invariant.


seveibar (approved):
i know i created the original issue, forgot to fix it etc. but we should stop encoding data inside of ids 😭

PR #2297, opened 2026-05-13
Fix default nominal trace width for krt autorouter routes

MustafaMulla29:


lib/components/primitive-components/Group/Group.ts
```diff
@@ -790,11 +790,14 @@ export class Group<Props extends z.ZodType<any, any, any> = typeof groupProps>
     const isAutoJumperPreset = this._isAutoJumperAutorouter(autorouterConfig)
     const isSingleLayerBoard = this._getSubcircuitLayerCount() === 1
 
+    const minTraceWidth = Number(props.minTraceWidth ?? 0.15)
+    const nominalTraceWidth = Number(props.nominalTraceWidth ?? 0.15)
```
imrishabh18, line 794: @seveibar Should the `nominalTraceWidth` be `0.15`?

imrishabh18 (approved):
Just confirm from Seve about the default value for this, or if there shouldn't be any default value

lib/components/primitive-components/Group/Group.ts
seveibar: I think so! But we should import this frome somewhere maybe so that we have it encoded… is it jlcpcb recommended trace size or something?

imrishabh18: jlcpcb capabilities page don't have anything as optimal size, it's just they have `0.10` as min trace width. So, I think `0.15` is okay

seveibar: ok

seveibar (approved):

PR #2305, opened 2026-05-16
Fix jlcpcb CAD fallback for library footprints

techmannih:
## What changed
This updates CAD fallback generation for library-style footprints so `cad_component.footprinter_string` is only emitted for direct footprinter strings such as `0402` or `soic8`.

For library references like `jlcpcb:C51950748`, KiCad library refs, URLs, and static asset paths, the fallback now emits a bounding-box CAD component instead of forwarding the library reference as a footprinter string.

## Why this changed
The 3D viewer was trying to parse `jlcpcb:...` as a raw footprinter function when no CAD model was available. That produced runtime errors like:

`Invalid footprint function, got "jlcpcb"`

The underlying footprint could still load for PCB rendering, but the CAD fallback path was sending an invalid string into the 3D engine.

## Impact
This fixes 3D rendering for library footprints that do not provide a CAD model by falling back cleanly to a bounding box instead of surfacing a parser error.

## Root cause
`NormalComponent.doInitialCadModelRender()` treated any string footprint as a valid `footprinter_string` fallback. Library refs such as `jlcpcb:...` are valid footprint lookup keys, but they are not valid raw footprinter expressions.

## Validation
- `bun test tests/footprint/footprint-library-map-jlcpcb-cadmodel.test.tsx`
- `bun test tests/components/normal-components/cad-component-bounding-box.test.tsx`

## Notes
Repo-wide build/type generation still has unrelated pre-existing DTS failures in `lib/components/primitive-components/PcbTrace.ts`.

### DIFF
--- a/lib/components/base-components/NormalComponent/NormalComponent.ts
+++ b/lib/components/base-components/NormalComponent/NormalComponent.ts
@@ -1621,12 +1621,18 @@ export class NormalComponent<
       cadModelProp === undefined ? this._asyncFootprintCadModel : cadModelProp
     const footprint =
       this.getFootprinterString() ?? this._getImpliedFootprintString()
-    const footprintString = typeof footprint === "string" ? footprint : null
-    const canUseFootprinterStringFallback =
-      !!footprintString &&
-      !parseLibraryFootprintRef(footprintString) &&
-      !isHttpUrl(footprintString) &&
-      !isStaticAssetPath(footprintString)
+    let footprintString: string | undefined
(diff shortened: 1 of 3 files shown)


seveibar (commented):
bad naming, the fix is valid but there are issues:

1) `canUseFootprintStringFallback` is unclear. There is no fallback. Better naming is needed e.g. `footprintIsFootprinterString` is more clear
2) Ternary usage is confusing
3) `!!` is never allowed

seveibar (approved):

PR #2310, opened 2026-05-19
Add automatic breakout point generation for fanout groups

MustafaMulla29:
Implements automatic breakout point generation for `<breakout>`/`<fanout>` groups. When a group does not define explicit `<breakoutpoint>` or `<fanoutpoint>` children, PCB ports inside the group are assigned generated breakout targets along the group boundary. The target distance is controlled by the group bounds and padding, while routing and via placement remain handled by the existing autorouter.

Explicit breakout points continue to take precedence when provided.


MustafaMulla29:
@seveibar not sure if its the correct breakout

seveibar (commented):
so this is a good start but i don't think this is the right method for breakouts

We want to only break out pins that are actually used- this is why breakouts shouldn't be subcircuits (or at least, shouldn't be normal subcircuits) because they use information from their parent to determine how they should break out

PR #2311, opened 2026-05-19
Add breakout repros and autorouting end-phase stack snapshots

MustafaMulla29:


lib/components/primitive-components/Group/Group.ts
```diff
@@ -965,6 +965,18 @@ export class Group<Props extends z.ZodType<any, any, any> = typeof groupProps>
...
+          traces,
+        }
+
+        this.root?.emit("autorouting:end", {
+          type: "autorouting:end",
+          subcircuit_id: this.subcircuit_id,
+          componentDisplayName: this.getString(),
+          srj: outputSimpleRouteJson,
+          simpleRouteJson: outputSimpleRouteJson,
+        })
```
seveibar, line 978: are you sure this doesn't already exist?

lib/events/index.ts
```diff
@@ -48,6 +48,10 @@ export interface AutoroutingProgressEvent {
 
 export interface AutoroutingEndEvent {
   type: "autorouting:end"
+  subcircuit_id: string
+  componentDisplayName: string
+  srj: SimpleRouteJson
+  simpleRouteJson: SimpleRouteJson
```
seveibar, line 53: why are you including simple route json twice?

lib/events/index.ts
```diff
@@ -48,6 +48,10 @@ export interface AutoroutingProgressEvent {
 
 export interface AutoroutingEndEvent {
   type: "autorouting:end"
+  subcircuit_id: string
+  componentDisplayName: string
+  srj: SimpleRouteJson
```
seveibar:
```suggestion
```

lib/components/primitive-components/Group/Group.ts
```diff
@@ -965,6 +965,18 @@ export class Group<Props extends z.ZodType<any, any, any> = typeof groupProps>
...
+        const outputSimpleRouteJson = {
+          ...simpleRouteJson,
+          traces,
+        }
+
+        this.root?.emit("autorouting:end", {
+          type: "autorouting:end",
+          subcircuit_id: this.subcircuit_id,
+          componentDisplayName: this.getString(),
+          srj: outputSimpleRouteJson,
```
```suggestion
```

seveibar (approved):
I think we need to see the autorouting inputs in addition to the outputs in the snapshot (both the start and end simpleRouteJson)

lib/components/primitive-components/Group/Group.ts
MustafaMulla29: Yes, event was existing but it was not emitted

PR #2312, opened 2026-05-20
Add autorouting start SRJ stack snapshots for breakout repros

MustafaMulla29:


seveibar (approved):
are you sure routes are being draw for the start phases? They're missing in the snapshots

MustafaMulla29:
> are you sure routes are being draw for the start phases? They're missing in the snapshots

autorouting:start is not drawing traces

PR #2314, opened 2026-05-20
Implement automatic breakout point routing with snapshot coverage

MustafaMulla29:


seveibar (commented):
- The routing snapshots don't make sense- i don't ever see previous routes before i see new routes
- i think the breakout solver should really be in a separate repo so that we can have tests dedicated to it. Right now i have no idea how it works without carefully reading the implementation- the snapshots do not show how it works, and there's no way for me to build visual intuition
- breakout phases are running AFTER the default phase instead of before it. That doesn't make sense to me- we should be routing the breakouts first no?

seveibar:
@ShiboSoftwareDev what do you think about my comments above?

lib/components/primitive-components/Breakout/Breakout.ts
```diff
@@ -1,8 +1,11 @@
 import { breakoutProps } from "@tscircuit/props"
 import { Group } from "../Group/Group"
 import type { z } from "zod"
+import { Breakout_insertAutomaticBreakoutPoints } from "./Breakout_insertAutomaticBreakoutPoints"
 
 export class Breakout extends Group<typeof breakoutProps> {
+  _hasInsertedAutomaticBreakoutPoints = false
```
seveibar, line 7: this is good

lib/components/primitive-components/Breakout/Breakout_insertAutomaticBreakoutPoints.ts
```diff
@@ -0,0 +1,209 @@
...
+          source_trace_id: sourceTrace.source_trace_id,
+          x: boundaryPoint.x,
+          y: boundaryPoint.y,
+        })
+      }
+    }
+  }
+
+  breakout._hasInsertedAutomaticBreakoutPoints = true
+}
```
seveibar, line 209: i need some kind of visualization that helps me build intuition about this function. The algorithm parts are mixed in with the logistics (data pipelining)- we should avoid this by calling out to a separate repo for the actual breakout algorithm

lib/utils/autorouting/getAutomaticBreakoutPointPosition.ts
```diff
@@ -0,0 +1,225 @@
...
+  const dx = point.x - obstacle.center.x
+  const dy = point.y - obstacle.center.y
+  const localX = dx * cos - dy * sin
+  const localY = dx * sin + dy * cos
+
+  return (
+    Math.abs(localX) <= obstacle.width / 2 + margin &&
+    Math.abs(localY) <= obstacle.height / 2 + margin
+  )
+}
```
seveibar, line 43: you're only handling rect obstacles here iiuc- i think this file contains a lot of bugs and there's not really a good way to test it against a tdataset from this repo

lib/utils/autorouting/getAutomaticBreakoutPointPosition.ts
```diff
@@ -0,0 +1,225 @@
...
+      const x = from.x + t * dx
+      if (t > 0 && x >= boundary.left && x <= boundary.right) {
+        candidates.push({ t, point: { x, y } })
+      }
+    }
+  }
+
+  candidates.sort((a, b) => a.t - b.t)
+  return candidates[0]?.point ?? null
+}
```
seveibar, line 82: what is happening here?

lib/utils/autorouting/getAutomaticBreakoutPointPosition.ts
```diff
@@ -0,0 +1,225 @@
...
+
+const getBoundaryPerimeterCandidates = ({
+  boundary,
+  step,
+}: {
+  boundary: AutomaticBreakoutBoundary
+  step: number
+}) => {
+  const candidates: AutomaticBreakoutPoint[] = []
+  const addCandidate = (candidate: AutomaticBreakoutPoint) => {
```
seveibar, line 128: closure

lib/utils/autorouting/getAutomaticBreakoutPointPosition.ts
```diff
@@ -0,0 +1,225 @@
...
+      edge,
+      boundary,
+      step,
+    })) {
+      addCandidate(candidate)
+    }
+  }
+
+  return candidates
+}
```
seveibar, line 152: I get the general idea of this algorithm, but we need to make this a SOLVER- algorithms should not be in these non-debuggable formats. Move to a separate repo and make it a solver

lib/utils/autorouting/getAutomaticBreakoutPointPosition.ts
```diff
@@ -0,0 +1,225 @@
...
+
+export const getAutomaticBreakoutPointPosition = ({
+  insidePortPosition,
+  outsideTargetPosition,
```
seveibar, line 156: what is a target? Rename this?

lib/utils/autorouting/getAutomaticBreakoutPointPosition.ts
```diff
@@ -0,0 +1,225 @@
...
+}): AutomaticBreakoutPoint | null => {
+  const rayIntersection = getRayBoundaryIntersection({
+    from: insidePortPosition,
+    to: outsideTargetPosition,
+    boundary,
+  })
+  if (!rayIntersection) return null
+  if (boundaryPointSpacing <= 0) return rayIntersection
+
+  const hasConflict = (candidate: AutomaticBreakoutPoint) =>
```
seveibar, line 181: closures!

lib/utils/autorouting/getAutomaticBreakoutPointPosition.ts
```diff
@@ -0,0 +1,225 @@
...
+      step,
+    })) {
+      addCandidate(candidate)
+    }
+  }
+
+  return candidates
+}
+
+export const getAutomaticBreakoutPointPosition = ({
```
seveibar, line 154: ???????? I have no idea what these mean!!!! There are no docstrings! This algo isn't inside a solver! No way to debug!

PR #2316, opened 2026-05-21
Normalize through_obstacle autorouter segments before persisting pcb_trace.route

rushabhcodes:
## Summary

  This PR fixes a core autorouting bug where local autorouter output could write through_obstacle segments directly into
  pcb_trace.route, even though the public route format should expose those segments as through_pad.

  ## What changed

  - Normalize through_obstacle route points to through_pad before inserting pcb_trace records from autorouter output
  - Apply the same normalization in the reroute/jumper trace-splitting flow so downstream route handling sees the public route shape
    consistently
  - Add a regression test that reproduces the bug with a custom autorouter returning a through_obstacle segment and verifies the
    persisted route no longer contains it

  ## Why this matters

  This is a tscircuit/core behavior fix, not just a test update. It improves the public route data model exposed by autorouting and
  prevents internal autorouter-specific segment types from leaking into persisted PCB trace data.


lib/components/primitive-components/Group/Group.ts
```diff
@@ -1208,6 +1208,18 @@ export class Group<Props extends z.ZodType<any, any, any> = typeof groupProps>
     // Apply each routed trace to the corresponding circuit trace
     // const circuitTraces = this.selectAll("trace") as Trace[]
     for (const routedTrace of routedTraces) {
+      const publicRoute = routedTrace.route.map((point) => {
```
```suggestion
      const cjRoute = routedTrace.route.map((point) => {
```

lib/components/primitive-components/Group/Group.ts
```diff
@@ -1216,7 +1228,7 @@ export class Group<Props extends z.ZodType<any, any, any> = typeof groupProps>
       // TODO use upsert to make sure we're not re-creating traces
       const pcb_trace = db.pcb_trace.insert({
         subcircuit_id: this.subcircuit_id!,
-        route: routedTrace.route as any,
+        route: publicRoute as any,
```
```suggestion
        route: cjRoute as any,
```

lib/components/primitive-components/Group/Group.ts
```diff
@@ -1301,12 +1313,24 @@ export class Group<Props extends z.ZodType<any, any, any> = typeof groupProps>
             db.source_net.get(sourceTraceId)?.subcircuit_id)
           : undefined) ?? this.subcircuit_id!
 
+      const publicRoute = pcb_trace.route.map((point: any) => {
```
```suggestion
      const cjRoute = pcb_trace.route.map((point: any) => {
```

tests/features/autorouting-through-obstacle-public-route.test.tsx
```diff
@@ -0,0 +1,73 @@
...
+test("through_obstacle route points are normalized before writing pcb_trace.route", async () => {
+  const { circuit } = getTestFixture()
+
+  circuit.add(
+    <board
+      width="20mm"
+      height="20mm"
+      autorouter={{
+        local: true,
+        groupMode: "subcircuit",
```
seveibar:
```suggestion
```

seveibar (approved):

PR #2331, opened 2026-05-24
fix: parse scientific notation in pcb coordinate strings

asapsav:
## Summary

Footprint primitives with CAD-imported coordinates in scientific notation (e.g. `pcbX="-1.1368683772161603e-13mm"`) were failing with:

```
source_invalid_component_property_error: Invalid pcbY value for SmtPad:
Invalid calc() expression. expression="1.1368683772161603e-13mm"
```

This happens when footprints are auto-imported from EasyEDA/JLC/KiCad sources that serialize near-zero values as IEEE float dust.

**Root cause:** In `evaluateCalcString.ts`, `tokenize()` only consumed digits and `.` for numbers. For `1.1368683772161603e-13mm` it parsed `1.1368683772161603`, stopped at `e`, and treated `e` as a unit suffix → `Unknown unit: "e"`, which upstream is reported as an invalid calc expression.

## Changes

- Extend the number lexer in `tokenize()` to accept optional scientific notation (`e`/`E`, optional sign, exponent digits) before the optional unit suffix (`mm`, etc.)
- Normalize `-0` to `0` in the final result (avoids negative zero from unary `-` on dust values)

## Testing

**Unit tests (`evaluateCalcString`):**
- `1.5e-2mm` → `0.015`
- `2E+3mm` → `2000`
- `calc(1e-13mm + 2mm)` → `2`

**Integration test (`<smtpad>`):**
- Footprint with `pcbX="-1.1368683772161603e-13mm"` and `pcbY="0mm"` renders without `source_invalid_component_property_error`; pad `x` is ~`0`

**Regression checks:**
- Existing calc expression tests pass (`calc(board.minX + 1mm)`, legacy `board.minx`, etc.)
- RP2040 decoupling capacitors repro passes (uses the same scientific-notation coordinate in the RP2040 footprint fixture)

## to reproduce error with and without the fix:

git checkout 34770277 -- lib/utils/evaluateCalcString.ts
bun test tests/utils/evaluate-calc-string.test.ts tests/components/primitive-components/smtpad-scientific-notation-coordinates.test.tsx
git checkout HEAD -- lib/utils/evaluateCalcString.ts   # restore fix

### DIFF
--- a/lib/utils/evaluateCalcString.ts
+++ b/lib/utils/evaluateCalcString.ts
@@ -137,11 +137,6 @@ function tokenize(expr: string, units: Record<string, number>): Token[] {
         throw new Error(`Invalid number: "${numberText}"`)
       }
 
-      // Snap sub-nm float dust to zero (common in CAD-imported footprints)
-      if (Math.abs(num) < 1e-9) {
-        num = 0
-      }
-
       // Optional unit directly after number (e.g. "1mm")
       const unitStart = i
       while (i < expr.length && /[A-Za-z]/.test(expr[i])) {
--- a/lib/utils/evaluateCalcString.ts
+++ b/lib/utils/evaluateCalcString.ts
@@ -38,7 +38,7 @@ export function evaluateCalcString(
   const tokens = tokenize(expr, units)
   const result = parseExpression(tokens, knownVariables)
 
-  return result
+  return result === 0 ? 0 : result
 }
 
 export function extractCalcIdentifiers(input: string): string[] {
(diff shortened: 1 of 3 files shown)


imrishabh18 (changes requested):
I am not sure if this is the right place for the fix, can you create a test in here to demonstrate the fix

seveibar (commented):
needs a test

asapsav:
added tests back

to reproduce error with and without the fix:
```
git checkout 34770277 -- lib/utils/evaluateCalcString.ts
bun test tests/utils/evaluate-calc-string.test.ts tests/components/primitive-components/smtpad-scientific-notation-coordinates.test.tsx
git checkout HEAD -- lib/utils/evaluateCalcString.ts   # restore fix
```

@imrishabh18  where is the right place to fix? new to the repo

PR #2338, opened 2026-05-27
Add arduino breakout test and update deps

MustafaMulla29:


imrishabh18 (commented):
Can you screenshot and point which traces are overlapping?

PR #2344, opened 2026-05-28
Initial integration of breakout-point-solver

MustafaMulla29:
This is the partial integration of breakout-point-solver, next pr will be to fix overlapping and duplicate traces in the breakout snapshots.


lib/components/base-components/Renderable.ts
```diff
@@ -57,6 +57,8 @@ export const orderedRenderPhases = [
   "PcbLayout",
   "PcbBoardAutoSize",
   "PanelLayout",
+  "PcbAutoBreakoutPointRender",
+  "PcbBreakoutPointRender",
```
MustafaMulla29, line 61: Kept two phases, instead of duplicating the db insert records for pcb_breakout_point in both Breakout.ts and BreakoutPoint.ts

tests/breakout/__snapshots__/breakout-sot23-regulator-power-rail-pcb.snap.svg
imrishabh18, line 1: Can you run the drc check for both before and after? Let us know what is the difference

lib/components/base-components/Renderable.ts
seveibar: Rename to something like PcbFindAutoBreqkoutPoints

lib/components/primitive-components/BreakoutPoint.ts
```diff
@@ -3,13 +3,30 @@ import { PrimitiveComponent } from "../base-components/PrimitiveComponent"
...
   typeof breakoutPointProps
 > {
   pcb_breakout_point_id: string | null = null
   matchedPort: Port | null = null
   matchedNet: Net | null = null
   isPcbPrimitive = true
+  private resolvedTarget: ResolvedBreakoutPointTarget | null = null
+
+  setResolvedTarget(target: ResolvedBreakoutPointTarget): void {
+    this.resolvedTarget = target
```
seveibar, line 24: ??????

lib/components/primitive-components/Group/region-replacement/delete-existing-pcb-traces-replaced-by.ts
```diff
@@ -22,6 +22,10 @@ function addPossibleReplacementSourceTraceId(
   }
 }
 
+function isRerouteReplacementName(value: unknown): boolean {
+  return typeof value === "string" && value.includes("_reroute_")
```
seveibar, line 26: Huge anti pattern

seveibar (changes requested):
Make it smaller, we NEED to merge incrementally!!!

PR #2359, opened 2026-05-29
fix: When `<chip />` and `<smtpad />` both have layer as `bottom` then the footprint should render it on the `bottom` layer

imrishabh18:


seveibar (approved):
I get the idea here- i wonder if we need a footprint prop that basically says something like "primaryLayer" so that its more clear what layer the user was specifying the footprint for

imrishabh18:
> I get the idea here- i wonder if we need a footprint prop that basically says something like "primaryLayer" so that its more clear what layer the user was specifying the footprint for

image

Even for this case when we have the `originalLayer` as `bottom` the pads are rendered on `top`, so this is another bug

PR #2360, opened 2026-05-30
Fix breakout point SRJ construction to route through boundary points

MustafaMulla29:
- Filter directTraceConnections by subcircuit_id so internal traces routed inside a breakout are not duplicated in the parent's SRJ
- Add breakout point coordinates as additional points in pointsToConnect for cross-boundary traces, ensuring the autorouter routes through the boundary crossing
- Skip breakout points that are already handled by cross-boundary traces; only create direct [port, bp] connections for manual breakout points without a cross-boundary trace


MustafaMulla29:
@seveibar tried to solve it by following your suggestions, but i think there are still issues.

tests/breakout/__snapshots__/breakout-qfp16-with-header-and-passives-pcb.snap.svg
MustafaMulla29, line 1: Here it creates 2 traces for VCC connection but one stops at breakoutpoint at top

tests/breakout/__snapshots__/breakout-sot23-regulator-power-rail-autorouting-srj.snap.svg
MustafaMulla29, line 1: here as well, extra trace to the breakout point, instead of going through it.

tests/breakout/__snapshots__/fanout-soic8-sensor-to-i2c-header-pcb.snap.svg
MustafaMulla29, line 1: This one looks quite good

seveibar (commented):
I don’t think this is correct, breakout points should be used but their matched point should be excluded, this will fix traces not using the breakout point

MustafaMulla29:
@seveibar something or the other is getting regressed if I try to fix it

lib/utils/autorouting/getSimpleRouteJsonFromCircuitJson.ts
```diff
@@ -303,28 +338,29 @@ export const getSimpleRouteJsonFromCircuitJson = ({
...
+              },
           ...hintPoints,
-          {
-            x: portB.x!,
-            y: portB.y!,
-            layer: layerB,
-            pointId: portB.pcb_port_id,
-            pcb_port_id: portB.pcb_port_id,
-          },
+          replaceB
```
seveibar: this is a really bad pattern! I would call this sabotage, you're both introducing bad naming "replaceA" with a large ternary! Keep the code clean!

seveibar: This is core! This code is very high quality

seveibar (commented):
image

seveibar:
@MustafaMulla29 this is looking a lot better

lib/utils/autorouting/getSimpleRouteJsonFromCircuitJson.ts
```diff
@@ -319,15 +319,35 @@ export const getSimpleRouteJsonFromCircuitJson = ({
...
+      for (const spId of trace.connected_source_port_ids) {
+        const bp = sourcePortIdToBreakoutPoint.get(spId)
+        if (!bp || bp.subcircuit_id === subcircuit_id) continue
+        const target =
+          spId === trace.connected_source_port_ids[0] ? pointA : pointB
+        target.x = bp.x
+        target.y = bp.y
+        target.layer = "top"
+        delete (target as any).pointId
+        delete (target as any).pcb_port_id
```
seveibar: what is the target? Why are you setting the target layer to top? I don't understand this-

tests/breakout/__snapshots__/fanout-soic8-sensor-to-i2c-header-autorouting-srj.snap.svg
MustafaMulla29, line 1: Not sure what is happening here with the trace going at the top

seveibar (approved):

PR #2362, opened 2026-05-30
Add pcb_copper_text support to circuit-json footprint rehydration

anil08607:
## Changes

- add `pcb_copper_text` handling to `createComponentsFromCircuitJson`
- rehydrate imported copper text with its text, anchor position, rotation, layer, mirrored, and knockout fields

## Why

The footprint/import rehydration path supported only a subset of valid PCB circuit-json element types. This change adds support for `pcb_copper_text`, so imported footprint JSON can preserve copper text primitives during rehydration.


ShiboSoftwareDev (commented):
Add tests including an svg snapshot test

ShiboSoftwareDev (approved):

PR #2363, opened 2026-05-30
fix: `orginalLayer` and` layer` prop both have it as `bottom` but due to missing layer prop on `<chip />` it was rendering on top due to flip bug

imrishabh18:


seveibar (approved):
Its more like _getFootprintOriginalLayer

PR #2364, opened 2026-05-31
  Add custom SPICE subcircuit models via <spicemodel />

ShiboSoftwareDev:
Adds a Spicemodel component that can be passed through spiceModel={`<spicemodel />`} on normal
  components to emit simulation_spice_subcircuit circuit-json. The implementation uses spicets
  to parse .subckt definitions, maps SPICE pins to component ports through spicePinMapping or
  matching pin names.


lib/utils/spice/parse-spice-subckt.ts
```diff
@@ -0,0 +1,21 @@
+export interface ParsedSpiceSubckt {
+  modelName: string
+  pinNames: string[]
+}
+
+export function parseSpiceSubckt(source: string): ParsedSpiceSubckt | null {
```
seveibar: Remove (this is a hack)

seveibar (commented):
This would be a normal component feature, not just for chips fwiw

you didnt create a new class (all other elements have classes?)

I dont understand why youre parsing the subcircuit to get the name- it doesnt seem necessary?

lib/utils/spice/parse-spice-subckt.ts
seveibar: Why are we parsing for the name- why do we need it

lib/components/index.ts
```diff
@@ -90,3 +90,4 @@ export { SchematicCell } from "./primitive-components/SchematicCell"
 export { SymbolComponent as Symbol } from "./primitive-components/Symbol"
 export { AnalogSimulation } from "./primitive-components/AnalogSimulation"
 export { VoltageProbe } from "./primitive-components/VoltageProbe"
+export { Spicemodel } from "./primitive-components/Spicemodel"
```
seveibar: SpiceModel

seveibar (approved):

lib/components/base-components/NormalComponent/NormalComponent.ts
```diff
@@ -11,6 +11,7 @@ import type {
   CadModelWrl,
   FootprintInsertionDirection,
   SchematicPortArrangement,
+  SpicemodelElement,
```
seveibar: SpiceModelElement

lib/components/primitive-components/Spicemodel.ts
```diff
@@ -0,0 +1,108 @@
+import { spicemodelProps } from "@tscircuit/props"
+import type {
+  SimulationSpiceSubcircuitInput,
+  SourceInvalidComponentPropertyErrorInput,
+} from "circuit-json"
+import { parseSpiceNetlist } from "spicets"
+import { Port } from "./Port"
+import { PrimitiveComponent } from "../base-components/PrimitiveComponent"
+
+export class Spicemodel extends PrimitiveComponent<typeof spicemodelProps> {
```
seveibar: SpiceModel (qlso filename)

seveibar (approved):
Surprised this works, i think usually children elements need special handling to properly work w the render cycle

PR #2365, opened 2026-05-31
chore: update the autorouter version in core to the latest

imrishabh18:


imrishabh18 (commented):
@seveibar Some of the circuits are failing to route with the new autorouter, most of them are breakout boards

seveibar (approved):

seveibar:
@imrishabh18 the default autorouter pipeline should now be Pipeline7, and we should add fallbacks- cc @ShiboSoftwareDev @Abse2001 @0hmX re:regressions need investigation

we should allow the user to specify a version to go back to pipeline4 (autorouter=“v2”?)

PR #2367, opened 2026-06-01
Fix solderjumper bridged props PCB footprint rendering

mohan-bee:
This fix makes `bridged` and `bridgedPins` the source of truth for solder jumper PCB rendering.

Before this change, a user could write:

```tsx
<solderjumper footprint="solderjumper2" bridgedPins={[["1", "2"]]} />
```

but the PCB footprint still rendered as an open solder jumper because core passed `solderjumper2` directly to footprinter.

Now `SolderJumper` resolves those semantic bridge props into the matching footprinter variant before the normal footprint render path runs:

```txt
solderjumper2 + bridgedPins [["1", "2"]] -> solderjumper2_bridged12
solderjumper3 + bridgedPins [["2", "3"]] -> solderjumper3_bridged23
```

Existing explicit bridged footprints are left unchanged, so backward-compatible usage like this still works:

```tsx
<solderjumper footprint="solderjumper2_bridged12" />
```

A small PCB snapshot repro was added to verify that bridged solder jumper props now produce bridged copper on the PCB.


seveibar (commented):
Instead of modifying props, ideally we have a method like resolveFootprint that is used whenever the footprint is needed, this will require changes in a couple places, maybe start with the repro (pr 1), then introduce the resolveFootprint, then introduce the fix to solder jump


Alternatively instead of using super, try to modify a method like doInitialFootprintRender etc

using super is a bit of a hack here

mohan-bee:
fixed

PR #2370, opened 2026-06-01
Fix breakout source trace id assignment

MustafaMulla29:


lib/components/primitive-components/BaseBreakoutPoint.ts
```diff
@@ -86,10 +88,14 @@ export class BaseBreakoutPoint<
   _setPositionFromLayout(newCenter: { x: number; y: number }) {
     const { db } = this.root!
     if (!this.pcb_breakout_point_id) return
-    db.pcb_breakout_point.update(this.pcb_breakout_point_id, {
+    const update: { x: number; y: number; source_trace_id?: string } = {
```
seveibar: don't use verbs for variables, you need to use nouns for variables

e.g. `updatedProperties`

```suggestion
    const update: { x: number; y: number; source_trace_id?: string } = {
```

lib/components/primitive-components/Group/Group.ts
```diff
@@ -135,6 +135,8 @@ export class Group<Props extends z.ZodType<any, any, any> = typeof groupProps>
   _asyncAutoroutingResult: {
     output_simple_route_json?: SimpleRouteJson
     output_pcb_traces?: (PcbTrace | PcbVia)[]
+    pcb_trace_ids_to_delete?: string[]
```
```suggestion
    /* PCB traces that are being re-routed */
    pcb_trace_ids_to_be_replaced?: string[]
```

seveibar (approved):
see my recommended renaming

PR #2373, opened 2026-06-02
Fix type issues with switch intrinsic

MustafaMulla29:
React already defines lowercase <switch> as an SVG intrinsic element:
`switch: React.SVGProps<SVGSwitchElement>`

tscircuit also defines lowercase `<switch>` as a circuit component:
`switch: Props.SwitchProps`

Newer React/TypeScript typings reject that duplicate declaration because the two switch prop types are different.

The fix avoids redeclaring switch in JSX.IntrinsicElements. Instead, it removes switch from the tscircuit intrinsic extension and adds the tscircuit switch props onto React’s shared SVG attributes:

```tsx
type TscircuitElementsWithoutReactConflicts = Omit<TscircuitElements, "switch">
  type SwitchSvgAttributeProps = Partial<
    Omit<Props.SwitchProps, "children" | "type">
  >

  declare module "react" {
    interface SVGAttributes<T> extends SwitchSvgAttributeProps {}

    namespace JSX {
      interface IntrinsicElements extends TscircuitElementsWithoutReactConflicts {}
    }
}
```

This lets existing circuit JSX keep working:
`<switch name="SW1" spst schX={0} schY={0} />`

while avoiding the TypeScript merge conflict with React’s built-in SVG <switch>.

Downside

The downside is reduced SVG prop strictness. Because the extra switch props are added to SVGAttributes<T>, TypeScript will now allow switch-specific props on any SVG JSX element:
`<circle spst schX={1} />`
`<svg dpdt simStartClosed />`

Those are not meaningful SVG props, but they would typecheck.

So this is a compatibility workaround. It preserves the existing `<switch>` API without breaking users.


seveibar (commented):
Can you somehow disable react html builtins? Basically i want tscircuit to work assuming react dom hasn’t been loaded

MustafaMulla29:
@seveibar is this what you meant?

seveibar (commented):
Im pretty sure you broke all tscircuit type checking with this?

seveibar:
@MustafaMulla29 can you revert back to original but try to keep reqct compat code separate-

seveibar (approved):

PR #2377, opened 2026-06-02
Assign routed trace source IDs from source ports

MustafaMulla29:
Assigns source_trace_id to autorouted PCB trace segments from real source connectivity instead of relying on merged connection_name values.

The resolver now:
  - Preserves valid source_trace_ids already returned by the autorouter.
  - Looks up source_port_ids from routed endpoint pcb_port_ids.
  - Falls back to endpoint pad/plated-hole lookup when route points do not include port IDs.
  - Matches those source ports against source_trace.connected_source_port_ids.
  - Picks a deterministic source trace when multiple candidates are connected.
  - Handles reroute segments that terminate at existing trace cut points by inheriting from the touched
    trace.


tests/examples/__snapshots__/example39-unbroken-copper-pour-escape-vias-pcb.snap.svg
MustafaMulla29, line 1: Not sure if the updated snapshot is correct

seveibar (approved):

PR #2379, opened 2026-06-02
Fix missing net labels when accidentally removed because they were considered duplicates incorrectly

Sang-it:


lib/components/primitive-components/Group/Group_doInitialSchematicTraceRender/insertNetLabelsForPortsMissingTrace.ts
```diff
@@ -111,17 +181,49 @@ export const insertNetLabelsForPortsMissingTrace = ({
...
+          nl,
+          connKey,
+          sourceNet,
+          text,
+        })
       })
       if (existingAtPort) continue
     }
 
+    if (!sourceNet) {
```
seveibar, line 195:
```suggestion
    // If a label isn't attached to a port, and it's connected to a net that other labels are connected to then remove it
    if (!sourceNet) {
```

seveibar (approved):

PR #2385, opened 2026-06-03
Add normal component footprint resolver

mohan-bee:


lib/components/base-components/NormalComponent/NormalComponent.ts
```diff
@@ -500,10 +500,18 @@ export class NormalComponent<
     return null
   }
 
+  // Override to choose the effective footprint without rewriting user props.
+  resolveFootprint() {
+    return (
+      this._parsedProps.footprint ??
+      this.props.footprint ??
```
```suggestion
```

seveibar (approved):

lib/components/base-components/NormalComponent/NormalComponent.ts
mohan-bee: ig we need props.footprint because it is a user defined footprint ?

PR #2386, opened 2026-06-03
Fix srj to set breakout bounds

MustafaMulla29:
The root cause 
When the code builds the outer routing problem, it includes all the already-routed inner copper as obstacles, so the outer route avoids it. That part is correct and intended. 

The problem: the inner handoff wire ends exactly at the breakout point — its tip is the breakout point. And the breakout point is also the outer route's target endpoint. So:  The outer route's target lands right on top of the tip of a thin sliver of copper, on the same layer.

So the real defect is in how the breakout point and its handoff copper are positioned/modeled in the SRJ: the parent's routing target is buried under the tip of the inner copper, on the boundary edge, with no clear free cell to begin/end the route.


lib/utils/autorouting/getSimpleRouteJsonFromCircuitJson.ts
```diff
@@ -149,6 +165,20 @@ export const getSimpleRouteJsonFromCircuitJson = ({
...
+            y: number
+          }[]
+          if (wires.length === 0) return true
+          const first = wires[0]
+          const last = wires[wires.length - 1]
+          return (
+            !touchesBreakoutPoint(first.x, first.y) &&
+            !touchesBreakoutPoint(last.x, last.y)
+          )
+        })
```
seveibar: i don't understand the idea behind this filter, why does something have to not be touching a breakout point to be considered a descendent trace? Are you trying to filter out already-routed traces? Why? Descendent traces need to be included as obstacles even if they're used from breakout right? Or am i misunderstanding?

lib/utils/autorouting/getSimpleRouteJsonFromCircuitJson.ts
```diff
@@ -229,7 +259,23 @@ export const getSimpleRouteJsonFromCircuitJson = ({
 
   let bounds: { minX: number; maxX: number; minY: number; maxY: number }
 
-  if (board && !board.outline) {
+  // For non-board subcircuits (e.g. breakout regions), the pcb_group
+  // defines the routing boundary, not the parent board.
+  const useGroupBoundsAsPrimary = !!(
```
seveibar:
```suggestion
  const useGroupBoundsAsSrjBounds = !!(
```

lib/utils/autorouting/getSimpleRouteJsonFromCircuitJson.ts
```diff
@@ -80,13 +80,15 @@ export const getSimpleRouteJsonFromCircuitJson = ({
   )
 
   let board: PcbBoard | undefined | null = null
+  let boardMatchesSubcircuit = false
```
seveibar:
```suggestion
  let subcircuitIsBoard = false
```

lib/utils/autorouting/getSimpleRouteJsonFromCircuitJson.ts
seveibar:
```suggestion
  const shouldUseGroupBoundsAsSrjBounds = Boolean(
```

seveibar (commented):
this might be okay but can you help me understand why i'm not seeing some obstacles here? I think maybe there was a bug introduced

image

MustafaMulla29:
Screenshot 2026-06-05 at 2 10 09 AM


@seveibar not sure why it fails with image diff here in CI locally every test passes without any snapshot update:
Screenshot 2026-06-05 at 2 16 11 AM

Might be the routing is changing on CI(is it an OS issue)?

seveibar (commented):
@MustafaMulla29 can you try assigning the trace obstacle to the net (obstacle.connectedTo) rather than adjust the bounds? It should be more stable across trace widths

MustafaMulla29:
> @MustafaMulla29 can you try assigning the trace obstacle to the net (obstacle.connectedTo) rather than adjust the bounds? It should be more stable across trace widths

I think i tried it, but the autorouter is not respecting the connectedTo ig, claude was saying autorouter is returning connectedTo:[], let me try that again. I'll let you know.

seveibar (approved):

PR #2393, opened 2026-06-04
add oval hole support

rushabhcodes:
## Summary
- add `oval` handling to the `Hole` primitive so it emits a `pcb_hole` with `hole_shape: "oval"`
- treat oval holes like other width/height hole shapes when computing PCB bounds
- add a focused regression test and PCB snapshot for `<hole shape="oval" />`

## Why
`@tscircuit/props` already accepts `shape="oval"`, and nearby code paths already understand oval holes, but `lib/components/primitive-components/Hole.ts` still fell through to the circular-hole branch. That meant oval holes were not emitted correctly from the primitive component itself.

## Validation
- `bunx tsc --noEmit`
- `bun test tests/components/primitive-components/hole-oval.test.tsx`
- `bun test tests/components/primitive-components/hole.test.tsx`
- `bun test tests/components/primitive-components/hole-pill.test.tsx`
- `bun test tests/components/primitive-components/hole-rect.test.tsx`


techmannih (commented):
Extraction circuit json missing

seveibar (approved):

PR #2395, opened 2026-06-04
update regex for power/ground label detection

Sang-it:


seveibar (approved):
is V3_3 captured?

Sang-it:
The regex doesn't handle that. Let me add that. Never seen that before.

seveibar (approved):

seveibar:
V3_3 is how we typically annotate 3.3V, see https://github.com/tscircuit/sparkfun-boards

PR #2402, opened 2026-06-07
fix: manualEdits position double-counts parent group's anchor (#2281)

### LINKED ISSUE #2281: manualEdits position double-counts the parent group's anchor for components inside positioned regions
## Summary

`manualEdits.pcb_placements` lands a manually-placed component at the wrong absolute position when its enclosing `<group>` has a non-zero `pcbX/pcbY`. The group anchor gets applied twice — once when computing the parent's transform, again via `translate(manualPlacement.x, manualPlacement.y)` — even though `manualPlacement` is already in subcircuit-absolute coordinates.

## Repro

```tsx
<board width="40mm" height="20mm" manualEdits={{
  pcb_placements: [
    { selector: "C1", center: { x: 11.4, y: 4.3 }, relative_to: "group_center" },
  ],
}}>
  <group name="region" pcbX={16} pcbY={0}>
    <chip name="U1" footprint="soic8" pinLabels={{ pin1: "VCC", pin8: "GND" }} />
    <capacitor name="C1" capacitance="100nF" footprint="0402"
      connections={{ pin1: "U1.VCC", pin2: "U1.GND" }} />
  </group>
</board>
```

Expected: `C1.center = (11.4, 4.3)` — the user dragged it there.

Actual: `C1.center = (27.4, 4.3)`. The +16 region anchor was added on top of the manual position.

(With the static-marking fix from #TBD applied, the bug becomes more visible: `C1.center = (43.4, 4.3)` — the anchor is added a third time by `applyPackOutput`. Without the static fix, the packer overwrites the wrong position with its own different wrong position, masking this bug.)

## Cause

In `lib/components/base-components/PrimitiveComponent.ts`'s `_computePcbGlobalTransformBeforeLayout`:

```ts
const manualPlacement = this.getSubcircuit()._getPcbManualPlacementForComponent(this)
if (manualPlacement && this.props.pcbX === undefined && this.props.pcbY === undefined) {
  return compose(
    this.parent?._computePcbGlobalTransformBeforeLayout() ?? identity(),
    compose(
      translate(manualPlacement.x, manualPlacement.y),
      rotate(rotation * Math.PI / 180),
    ),
  )
}
```

`manualPlacement` comes from `_getPcbManualPlacementForComponent`:

```ts
const center = applyToPoint(
  this._computePcbGlobalTransformBeforeLayout(),  // subcircuit's transform
  position.center,
)
return center
```

For a board subcircuit (transform = identity), the returned `center` is just `position.center` — the value the user wrote into `manual-edits.json`, which the IDE recorded in **board-absolute** coordinates. So `manualPlacement` is already in the same frame as the final intended position.

But then `_computePcbGlobalTransformBeforeLayout` composes it on top of the **parent group's** transform. For a chip directly under the board (parent = board, transform = identity), this happens to work. For a component inside a positioned group, the parent's translate is applied AND the absolute manual position is applied as a translate, double-counting the group anchor.

## Fix sketch

When a manualPlacement exists, the component's transform should be the manualPlacement directly, NOT composed with the parent's transform:

```ts
if (manualPlacement && this.props.pcbX === undefined && this.props.pcbY === undefined) {
  const rotation = this._getPcbRotationBeforeLayout() ?? 0
  return compose(
    translate(manualPlacement.x, manualPlacement.y),
    rotate(rotation * Math.PI / 180),
  )
}
```

(The same applies to the schematic equivalent at line ~1815 if it has the same shape.)

## Impact

Anyone using the IDE drag-and-pin flow for components inside positioned regions sees their drags land at the wrong place after rebuild. This is the dominant pattern for any board with multiple chip regions — every cap, resistor, etc. inside a regional group becomes mis-placed.

The companion fix #TBD (this PR's parent issue) makes the packer respect `manualEdits` placements; this issue is what makes those placements land where the user actually dragged them. Both are needed for the IDE drag workflow to be correct end-to-end.

sagarmaurya64-ai:
This PR resolves #2281 where manual placement coordinates inside a positioned group or translated subcircuit were double-counting the parent group's/subcircuit's translation.

### Cause
In `PrimitiveComponent.ts`, `manualPlacement` coordinates (retrieved via `getSubcircuit()._getPcbManualPlacementForComponent(this)`) are already subcircuit-absolute (meaning the subcircuit's global transform has already been pre-multiplied).
However, `_computePcbGlobalTransformBeforeLayout()` composed this with `this.parent?._computePcbGlobalTransformBeforeLayout()`, adding the parent's translation offset a second time.

### Fix
Composed the manual placement coordinate with only the parent's global **rotation** (extracted using `decomposeTSR` on the parent global transform), rather than the entire parent global transform.

### Testing
- Created a standard unit test file `tests/repro-2281.test.tsx` verifying the positioning offset behaves correctly both in plain positioned groups and in rotated parent groups.

### DIFF
--- a/lib/components/base-components/PrimitiveComponent/PrimitiveComponent.ts
+++ b/lib/components/base-components/PrimitiveComponent/PrimitiveComponent.ts
@@ -28,6 +28,7 @@ import {
   type Matrix,
   applyToPoint,
   compose,
+  decomposeTSR,
   flipY,
   identity,
   rotate,
@@ -504,12 +505,14 @@ export abstract class PrimitiveComponent<
       this.props.pcbY === undefined
     ) {
       const rotation = this._getPcbRotationBeforeLayout() ?? 0
+      const parentGlobalTransform =
+        this.parent?._computePcbGlobalTransformBeforeLayout() ?? identity()
+      const decomposed = decomposeTSR(parentGlobalTransform)
+      const parentRotation = (decomposed.rotation.angle * 180) / Math.PI
+
       return compose(
-        this.parent?._computePcbGlobalTransformBeforeLayout() ?? identity(),
-        compose(
-          translate(manualPlacement.x, manualPlacement.y),
-          rotate((rotation * Math.PI) / 180),
-        ),
+        translate(manualPlacement.x, manualPlacement.y),
+        rotate(((parentRotation + rotation) * Math.PI) / 180),
       )
     }
 
--- a/tests/repro-2281.test.tsx
+++ b/tests/repro-2281.test.tsx
@@ -1,0 +1,75 @@
+import { test, expect } from "bun:test"
+import { RootCircuit } from "lib/RootCircuit"
+import "lib/register-catalogue"
+
+test("manual placement in positioned group does not double-count parent transform", () => {
+  const circuit = new RootCircuit()
+
+  circuit.add(
+    <board
+      width="40mm"
+      height="40mm"
+      manualEdits={{
+        pcb_placements: [
+          {
+            selector: ".R1",
+            center: { x: 5, y: 5 },
+          },
+        ],
+      }}
+    >
+      <group pcbX={10} pcbY={10}>
+        <resistor name="R1" resistance="10k" footprint="0402" />
+      </group>
+    </board>,
+  )
+
+  circuit.render()
+
+  const resistor = circuit.selectOne(".R1")
+  expect(resistor).not.toBeNull()
+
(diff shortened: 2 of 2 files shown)


sagarmaurya64-ai:
Please reopen this PR, the fix is fully valid and regression tests are included.

sagarmaurya64-ai:
@seveibar please reopen this PR! The fix for manualEdits double-counting parent group's anchor (#2281) is fully complete and includes regression tests.

sagarmaurya64-ai:
Please reopen this PR, the fix is fully valid and regression tests are included.

PR #2415, opened 2026-06-10
Add pcb silkscreen graphic support

techmannih:
## Summary

- add a `SilkscreenGraphic` primitive component for BRep-based PCB silkscreen graphics
- teach `createComponentsFromCircuitJson` to reconstruct `pcb_silkscreen_graphic` elements from circuit JSON
- cover the round-trip with a PCB snapshot test

## Why

- allow circuit-json silkscreen graphics to round-trip back into core components

## Impact

- users can import BRep silkscreen graphics from circuit JSON and render them through the existing PCB snapshot flow

## Validation

- `bun test tests/utils/createComponentsFromCircuitJson/pcb-silkscreen-graphic-snapshot.test.tsx`
- `bunx tsc --noEmit`


lib/components/primitive-components/SilkscreenGraphic.ts
techmannih, line 1: using a fast-xml-parser to parse the imported SVG image so we can extract its shapes, transforms, and fill rules, convert them into paths/polygons, and then generate internal BRep pcb_silkscreen_graphic geometry

seveibar (changes requested):
Clean up code. The silkscreengraphics file should not contain a bunch of functions, create a directory for silkscreengrqphics and move functions to files

PR #2419, opened 2026-06-10
change the default autorouter to AutoroutingPipelineSolver7_MultiGraph

ShiboSoftwareDev:


tests/features/subcircuit-circuit-json/__snapshots__/subcircuit-circuit-json02-pcb.snap.svg
imrishabh18, line 1: crossing traces?

tests/repros/__snapshots__/repro48-555-timer-pcb.snap.svg
imrishabh18, line 1: Even this looks wrong

imrishabh18 (changes requested):
I don't think we should merge this, until these are fixed.

cc @seveibar

PR #2426, opened 2026-06-11
Fix custom symbol schematic solver pin IDs and fallback netlabel names

MustafaMulla29:
Core was building schematic trace solver pin IDs from `schematic_port.pin_number` only. Custom symbol ports can be authored with `name` but no `pinNumber`, so multiple ports on the same custom symbol were sent to the solver as the same ID, e.g. `HOST.undefined`. That made the solver input ambiguous and caused incorrect routing/fallback behavior.

This fixes the solver input generation to use the schematic port’s numeric `pin_number` when available, and fall back to the attached `source_port.name` for named custom symbol ports. The repro now asserts custom symbol connections are emitted as `HOST.pin1`, `HOST.pin2`, etc.

Also fixed fallback netlabel text generation in Port.ts: ports nested inside `<symbol>` were using the immediate parent symbol name, producing labels like `undefined_pin3`. Netlabel text now uses the owning normal component name, so custom symbol labels resolve correctly, e.g. `HOST_pin3`.


tests/repros/__snapshots__/repro129-host-custom-symbol-passives-schematic-schematic.snap.svg
MustafaMulla29, line 1: WIll be looking into trace solver why it creates those other two netlabels instead of traces

lib/components/primitive-components/Port/Port.ts
```diff
@@ -811,6 +811,11 @@ export class Port extends PrimitiveComponent<typeof portProps> {
    * port, but appears at the port it connects to.
    */
   _getNetLabelText(): string | undefined {
-    return `${this.parent?.props.name}_${this.props.name}`
+    const componentName =
+      this.getParentNormalComponent()?.props.name ?? this.parent?.props.name
```
seveibar, line 815: is `this.getParentNormalComponent()?.props.name` not good enough?

lib/components/primitive-components/Port/Port.ts
```diff
@@ -811,6 +811,11 @@ export class Port extends PrimitiveComponent<typeof portProps> {
...
+
+    return componentName
```
seveibar, line 817: when is componentName not defined for a port?

seveibar (commented):
makes code worse by adding ambiguity and extra LoC unnecessarily for the net label text getter

PR #2438, opened 2026-06-12
chore: update schematic-trace-solver dep

MustafaMulla29:


tests/features/spice-analysis/__snapshots__/ngspice-spice-analysis05-buck-converter-simulation.snap.svg
MustafaMulla29, line 1: Solver detects that trace is crossing from the D1's bounding box and doesnt reroute it and leaves the previous routed trace. I suspects its the symbol sizing issue.

PR #2441, opened 2026-06-14
 Fix subcircuit inflation for imported LEDs and fiducials

rushabhcodes:
Adds missing circuit JSON inflators for simple_led and simple_fiducial source components so imported subcircuits can render without failing on unsupported ftypes.

  This fixes the Arduino Uno reroute repro where imported KiCad circuit JSON crashed during subcircuit inflation. The PR includes a focused regression test that reproduces
  the missing inflator coverage and verifies LED source inflation plus fiducial PCB pad recovery.

  Impact:

  - Fixes a real imported-board rendering/rerouting failure
  - Expands tscircuit/core circuit JSON inflation support
  - Preserves imported fiducial PCB geometry through extracted footprint primitives
  - Adds regression coverage for the repaired behavior


seveibar:
Needs snapshot updates?

PR #2450, opened 2026-06-16
fix: support updated KiCad imports and preserve imported trace topology during inflation

rushabhcodes:
This branch updates the KiCad import dependencies in `core` and adapts subcircuit inflation to the new circuit JSON shape they produce.

## What changed

- bumped `kicad-to-circuit-json` from `^0.0.60` to `^0.0.96`
- bumped `kicadts` from `^0.0.35` to `^0.0.45`
- added inflators for `simple_led` and `simple_fiducial`
- wired those new source component types into `inflate-circuit-json.ts`
- fixed imported trace inflation to handle `1 source_trace -> many pcb_trace` instead of assuming a 1:1 mapping
- updated affected repro expectations and PCB snapshots

## Why this branch was needed

The dependency update changed the imported KiCad circuit JSON in two important ways:

1. Imported boards can now contain `simple_led` and `simple_fiducial` source components, which previously crashed inflation because `core` had no inflators for those `ftype` values.
2. Imported routed copper is no longer guaranteed to be represented as one `pcb_trace` per `source_trace`. On the Arduino Uno repro, the new import output contains `109` `source_trace` records but `188` `pcb_trace` records. The old inflation logic used `getWhere({ source_trace_id })`, so it kept only one routed segment for each `source_trace_id` and silently dropped the rest.

### LED and fiducial inflation

- `inflateSourceLed` rehydrates imported LEDs with PCB placement and footprint data.
- `inflateSourceFiducial` rehydrates imported fiducials with placement and pad diameter derived from the imported SMT pad.

### Trace inflation

`inflateSourceTrace` now inflates one `Trace` per matching imported `pcb_trace`, while still preserving the straight-line fallback when an injected trace has no routed PCB geometry.

That restores the missing imported traces instead of collapsing them during subcircuit inflation.

## Tests and snapshots

Updated coverage and snapshots for the affected repros:

- `repro116-arduino-uno.test.tsx`
- `repro116-arduino-uno-reroute.test.tsx`
- `repro116-arduino-uno-reroute-center.test.tsx`
- `repro116-arduino-uno-reroute-lower-mid.test.tsx`
- `repro117-schematic-disabled-inflation.test.tsx`


MustafaMulla29 (commented):
Ok, lets fix the failing tests first.

tests/repros/__snapshots__/repro116-arduino-uno-reroute-lower-mid-pcb.snap.svg
MustafaMulla29, line 1: Some of the traces are missing in this!
Screenshot 2026-06-16 at 1 33 44 PM

MustafaMulla29 (approved):
Take a review from rishabh as well!

imrishabh18 (commented):
why are the traces missing for those circuits?

ShiboSoftwareDev:
why are you modifying/disabling the tests?

PR #2451, opened 2026-06-16
fix: support imported rotated_rect SMT pads from circuit json

anil08607:
## Summary

This fixes `createComponentsFromCircuitJson` so imported `pcb_smtpad` elements with `shape: "rotated_rect"` are rendered correctly.

Previously, rotated rectangular SMT pads were skipped during import, which meant:
- no `rotated_rect` pad was created in the PCB db
- the snapshot rendered an empty board instead of the imported pad

This change:
- adds `rotated_rect` handling in `createComponentsFromCircuitJson`
- preserves imported pad geometry and metadata, including `width`, `height`, `ccw_rotation`, `corner_radius`, `layer`, and `port_hints`
- updates the PCB snapshot to reflect the rendered rotated pad


mohan-bee (approved):
LGTM

imrishabh18 (approved):

PR #2452, opened 2026-06-16
fix: missing board labels (A0, SCL, DIGITAL-PWM~) after subcircuit inflation from imported KiCad PCB

rushabhcodes:
## Problem

When importing a KiCad PCB (e.g. Arduino Uno) as a `<subcircuit circuitJson={...} />`, board labels like `A0`, `A1`, `SCL`, `SDA`, and `DIGITAL - PWM~` were silently dropped from the output. The silkscreen and copper text elements simply never appeared in the inflated circuit.

## Root Cause (two bugs in `inflateStandalonePcbPrimitives.ts`)

**Bug 1 — Empty-string `pcb_component_id` was not recognized as standalone:**
The filter that identifies standalone primitives (board-level elements not attached to any component) only accepted `null` or `undefined` as "no component":

```ts
// before — misses elements where pcb_component_id === ""
elm.pcb_component_id === null || elm.pcb_component_id === undefined
```

The KiCad converter emits `pcb_component_id: ""` for board-level labels, so every one of those elements was filtered out. Fixed by treating any falsy value as "no component":

```ts
// after
!elm.pcb_component_id
```

**Bug 2 — `pcb_copper_text` was absent from `standalonePrimitiveTypes`:**
`pcb_copper_text` was never included in the list of types eligible for standalone inflation, so copper-layer board labels (e.g. "UNO TH RevC") were unconditionally skipped regardless of their `pcb_component_id`.

## Fix

- Added `"pcb_copper_text"` to `standalonePrimitiveTypes`
- Changed the component-id check from strict null/undefined to a falsy guard

## Reproduction & Tests

- **New test** `repro131-arduino-uno-labels-missing` inflates the full Arduino Uno circuit JSON and asserts that `A0`, `A1`, `SCL`, `SDA`, `DIGITAL - PWM~`, and the copper-text label are all present in the inflated output
- Updated snapshots for `repro116` and `subcircuit-circuit-json08` which now correctly render the previously-missing labels


MustafaMulla29 (approved):
Take approval from rishabh as well before merging

tests/repros/__snapshots__/repro116-arduino-uno-reroute-lower-mid-pcb.snap.svg
imrishabh18, line 1: Can you add an actual kicad snapshot of this kicad_pcb file? The text is just cluttering.

PR #2453, opened 2026-06-16
Remove '_pin' from fallback connections netlabel

MustafaMulla29:


imrishabh18 (approved):
This is removing the `pin_` prefix from the netlabels connection, the PR title could be better.

Take Seve's review

tests/repros/__snapshots__/repro118-trace-label-overlap-schematic.snap.svg
rushabhcodes, line 1: cant figure out why this changed?

MustafaMulla29: Snapshot is the same nohing changed, not sure. Maybe because of the `BUN_UPDATE_SNAPSHOTS`

rushabhcodes: ooh the keys changed in svg
image

rushabhcodes (approved):

PR #2461, opened 2026-06-17
Fix subcircuit inflation for imported KiCad LEDs, fiducials, and split traces

ShiboSoftwareDev:
Updates kicad-to-circuit-json/kicadts and fixes the core inflation gaps exposed by the newer
  output.

  Adds inflators for simple_led and simple_fiducial, preserving imported footprint primitives where
  available. 
  
  Also fixes imported trace inflation when one logical source_trace has multiple physical
  pcb_trace geometries, so subcircuits no longer collapse split KiCad routes down to a single trace.


rushabhcodes (approved):

lib/components/primitive-components/Trace/Trace.ts
```diff
@@ -54,6 +54,7 @@ export class Trace
   pcb_trace_id: string | null = null
   schematic_trace_id: string | null = null
   _inflatedPcbTrace?: PcbTrace
```
imrishabh18: Refactor so that we have only _inflatedPcbTraces

imrishabh18: We don't want confusion with both these being present

imrishabh18 (approved):

PR #2463, opened 2026-06-17
Add support for `<silkscreengraphics />`

imrishabh18:


seveibar (approved):

seveibar:
@imrishabh18 we should use bitmap based filter methods

seveibar:
Ie we should NOT extract paths from svgs- but this is ok in the short term

PR #2464, opened 2026-06-17
add @tscircuit/eecircuit-engine as a dev dep

ShiboSoftwareDev:


imrishabh18 (approved):
Is this supposed to be added as a tar ball?

PR #2468, opened 2026-06-18
fix pcbSx silkscreen visibility

mohan-bee:
Motivation: the "& silkscreentext" selector's visibility hidden not working.


techmannih (commented):
repro doesnt make sense for it without confirmation, if we have to support the visibility for silkscreen so direct add support

mohan-bee:
i though repro would much better .

techmannih:
Make sure take review from seve, not sure for the visibility support for silkscreen.

PR #2476, opened 2026-06-19
fix: rebuild selector cache so fallback net labels include async-loaded ports

MustafaMulla29:
## Problem
When a fallback net label names every port on a net (e.g. `U1_OUT/R1_pin1`), a trace between a resistor and a `<connector standard="usb_c">` only produced `R2_pin1` — the connector's pin was missing from the label.

## Root Cause
The connector adds its ports via an async parts-engine fetch that resolves in a later render cycle. The `OptimizeSelectorCache` phase memoizes `selectAll("port")` into `_cachedSelectAllQueries` in the first cycle, before those ports exist. When the ports arrive, `_markDirty` correctly re-marks the phase dirty, but there was no `updateOptimizeSelectorCache` method — so the dispatcher's `update${phase}?.()` call silently no-op'd and the stale, connector-less cache was reused.

## Fix
Add the missing `updateOptimizeSelectorCache` lifecycle method (every phase follows the `doInitial`/`update` pattern, and the dispatcher already invokes it when a phase is dirty). It clears the memoized selector caches and re-warms them from the current tree, so `selectAll("port")` reflects async-loaded ports.


tests/repros/__snapshots__/repro-usbc-trace-too-long-schematic.snap.svg
MustafaMulla29, line 1: This is not routing because this test has R2 in a separate section and the last commit in the trace-solver removed routing from different sections. It routes when i remove R2 from the section. Basically this test is not properly written.

imrishabh18 (approved):

PR #2489, opened 2026-06-21
add repro for imported pcb_courtyard_rect rotation

anil08607:


ShiboSoftwareDev:
is it is a bug, it should be called a repro not a regression test

anil08607:
@ShiboSoftwareDev Thanks, I've updated the PR title. Could you check it now?

ShiboSoftwareDev:
I mean the filename should also reflect it's purpose

anil08607:
@ShiboSoftwareDev Done

ShiboSoftwareDev (approved):

PR #2501, opened 2026-06-24
Fix duplicate same-net labels from schematic trace fallback insertion

mohan-bee:
### Motivation

Named nets such as `BAT` could render duplicate overlapping schematic net labels near the same port. This happened in the fallback logic for ports missing a visible schematic trace: when a nearby label already represented the same source connection, the fallback did not safely reuse it for non-ground nets, so another label could be inserted for the same net.

In `repro132`, this showed up around `R8_BAT_TOP`, where `BAT` label could overlap or appear visually detached from the connected port.

### Change

- Reuse a nearby same-net schematic label instead of inserting a duplicate.
- Re-anchor an existing same-text label to the current schematic port so the label remains visually connected.
- Preserve explicit display labels by not renaming an existing nearby label when its text differs from the fallback label text.

Fixed repro132: https://github.com/tscircuit/core/pull/2501/changes#diff-08a18e894b09a4fc83026c2efefe3ba77a7be63ac6dcf90a0ce6fdffc8668efd


imrishabh18 (changes requested):
Image

Still see duplicate ports

mohan-bee:
we need the two labels. i have fixed the anchor position for the label @imrishabh18

imrishabh18 (approved):

PR #2504, opened 2026-06-24
Add trace name/displayName support for schematic net labels

MustafaMulla29:


lib/components/primitive-components/Trace/Trace.ts
```diff
@@ -72,6 +72,14 @@ export class Trace
     return this._parsedProps.thickness ?? this._parsedProps.width
   }
 
+  _getSchematicNetLabelText(): string | undefined {
+    return (
+      this._parsedProps.name ??
+      this._parsedProps.displayName ??
+      this._parsedProps.schDisplayLabel
```
seveibar: Wrong order

seveibar (commented):
Unreadable test (tiny), also see important comment

lib/components/primitive-components/Trace/Trace.ts
MustafaMulla29: @seveibar whats the correct one? should it be reversed?

seveibar (approved):

PR #2513, opened 2026-06-27
Exposed Nets V2, New Render Phase, allows class management of traces created for exposedNets

seveibar:


seveibar:
@Abse2001 feel free to merge if it looks good? The vias are driving me crazy but i THINK they're right- we really need to have the short check somehow in this repo (maybe a circuit-to-png version? idk) I'm super confused about the vias but i don't think they're wrong(?)

seveibar:
we really need short testing here 😖

Abse2001 (approved):

PR #2515, opened 2026-06-28
exposeNets support, add net to net connection support to sequential trace autorouter

seveibar:


seveibar:
CC @imrishabh18 @Abse2001

PR #2516, opened 2026-06-28
Introduce schematicsheet

seveibar:


seveibar:
CC @Abse2001 @imrishabh18 merged

PR #2520, opened 2026-06-29
Emit a warning when schematic sheet is not found

MustafaMulla29:


lib/components/base-components/NormalComponent/NormalComponent.ts
```diff
@@ -706,6 +706,29 @@ export class NormalComponent<
...
+    // user gets feedback that their schSheetName was ignored.
+    const schSheetName =
+      this._parsedProps?.schSheetName ?? this.props.schSheetName
+    if (schSheetName && this.source_component_id) {
+      const referencedSheet = db.schematic_sheet
+        .list()
+        .find(
+          (sheet) =>
+            sheet.name === schSheetName ||
+            sheet.schematic_sheet_id === schSheetName,
```
seveibar: Wtf
```suggestion
```

seveibar (changes requested):
There is already a function _resolveSchematicSheetId which prevents a manually and potentially buggy search as features are added

that file is massive- maybe refactor the instance method out to a separate render function using the ClassName_instanceMethod.ts paradigm

seveibar (approved):
GOOD

PR #2522, opened 2026-06-29
feat: lay out and render schematic sheets independently

MustafaMulla29:
Partition a group's components by schematic_sheet_id and run matchpack once per sheet, so components on different sheets are never packed together or influence each other's placement. Resolve each schematic trace / net label's sheet from the components it connects (rather than the board), so wires and net labels render on their own sheet. Adds a toMatchStackedSchematicSnapshot matcher and multi-sheet repros.


MustafaMulla29:
For the cross sheet connection, do we emit a warning or core should just show fallback connection netlabels?

tests/fixtures/extend-expect-circuit-snapshot.ts
```diff
@@ -24,7 +25,7 @@ async function saveSvgSnapshotOfCircuitJson({
 }: {
   soup: AnyCircuitElement[]
   testPath: string
-  mode: "pcb" | "schematic" | "pinout"
+  mode: "pcb" | "schematic" | "schematic-stacked" | "pinout"
```
seveibar, line 28:
```suggestion
  mode: "pcb" | "schematic" | "schematic_sheets" | "pinout"
```

Please update AGENTS.md with our handbook rule that enums should use underscores

seveibar (changes requested):
Please do only repros for the first PR- the implementation here is highly suspect- particularly because it uses the forbidden ai slop term “fallback”


The architecture is critical for this, if the implementation for independent sheet rendering is bad then we might have many bugs. Make sure your repros have more significant subcircuits like the ones from ti

seveibar:
Note:

image

PR #2523, opened 2026-06-29
Auto-scale simulation graphs with independent axes

ShiboSoftwareDev:
Automatically scales voltage and current simulation graphs into independent display
  sections when graphIndependentAxes is enabled. Removes deprecated per-probe graph
  scaling props while preserving probe display names and colors.

  Refactors graph scaling and scope trace insertion into typed helpers, and adds
  deterministic SVG snapshot coverage for multi-channel, mixed-shape, flat-voltage,
  and tiny-current graph cases.


seveibar (approved):
I think we should probably still allow the manual override sometimes- just because i know for voltage it would be common to have everything have a 0V center and the same Y axis scaling. But i think this default is perfect

seveibar (approved):
many of the schematics seem a bit buggy such that the snapshots can't really be understood- there are overlapping elements that don't need to be overlapping- it's hard to evaluate

image

PR #2524, opened 2026-06-29
Repro multi schematic sheet subcircuits

MustafaMulla29:


MustafaMulla29:
@seveibar merged, let me know if something is wrong in the repros

PR #2525, opened 2026-06-29
add text aware bounding box for switch and push buttons

Sang-it:
- Adds text-aware bounds handling for switch and pushbutton schematic symbols.

  - Keeps traces and net labels aligned when expanded text bounds affect pin positions.

  - Deduplicates overlapping explicit/solver net labels without removing distant labels on the same net.


seveibar (commented):
Is there an example of a circuit this fixes? It feels like the premise of text expanded boxes may be wrong- we can move reference designator text without moving the symbol

seveibar:
I think you should NOT change unrelated snapshota- i spent 5m searching for what changed

Sang-it:
This is the bounding box before
Screenshot 2026-06-29 at 10 56 22 PM

I couldn't find any tests with switches but I found a old example with resistors. The trace can route through it as it has no idea that the text even exists. This pr avoids that from ever happening with switches.

Screenshot 2026-06-29 at 11 02 50 PM

Sang-it:
Another repro I just found - (not in current changes, found while trying to fix the gnd/vcc label collision)
Screenshot 2026-06-29 at 11 39 27 PM

lib/components/primitive-components/Group/Group_doInitialSchematicTraceRender/applyNetLabelPlacements.ts
```diff
@@ -31,6 +120,49 @@ export function applyNetLabelPlacements(args: {
     schematicPortIdsWithRoutedTraces,
   } = args
   const { db } = group.root!
+  const eligiblePortIds = getPortIdsInsideExpandedTextBounds(db)
+  const explicitNetLabelTextsByPortId = new Map<string, Set<string>>()
```
seveibar, line 124:
```suggestion
  const explicitNetLabelTextsByPortId = new Map<SchematicPortId, Set<string>>()
```

lib/components/primitive-components/Group/Group_doInitialSchematicTraceRender/applyNetLabelPlacements.ts
```diff
@@ -1,14 +1,103 @@
...
+  return {
+    minX: center.x - halfWidth,
+    maxX: center.x + halfWidth,
+    minY: center.y - halfHeight,
+    maxY: center.y + halfHeight,
+  }
+}
+
+const netLabelTextBoundsOverlap = (
+  a: ReturnType<typeof getNetLabelTextBounds>,
```
seveibar, line 43:
```suggestion
  a: Bounds,
```

lib/components/primitive-components/Group/Group_doInitialSchematicTraceRender/applyNetLabelPlacements.ts
```diff
@@ -1,14 +1,103 @@
...
+  b: ReturnType<typeof getNetLabelTextBounds>,
+) => a.minX < b.maxX && a.maxX > b.minX && a.minY < b.maxY && a.maxY > b.minY
```
seveibar, line 45: replace with math-utils bounds overlapping function

lib/components/primitive-components/Group/Group_doInitialSchematicTraceRender/applyNetLabelPlacements.ts
```diff
@@ -1,14 +1,103 @@
...
+
+const hasExplicitNetLabelForPlacementPort = ({
+  text,
+  schPortIds,
+  explicitNetLabelTextsByPortId,
```
seveibar, line 50:
```suggestion
  explicitNetLabelTextsBySchPortId,
```

lib/components/primitive-components/Group/Group_doInitialSchematicTraceRender/applyNetLabelPlacements.ts
```diff
@@ -1,14 +1,103 @@
...
+  text,
+  sourceNet,
+  solverLabelCenter,
+  explicitNetLabels,
+}: {
+  text: string
+  sourceNet: SourceNet
+  solverLabelCenter: Point
+  explicitNetLabels: Map<
+    string,
```
seveibar, line 70: idk what this map is?
```suggestion
    string,
```

lib/components/primitive-components/Group/Group_doInitialSchematicTraceRender/applyNetLabelPlacements.ts
```diff
@@ -1,14 +1,103 @@
...
+    {
+      text: string
+      source_net_id?: string | null
+      center?: Point
+    }
+  >
+}) =>
```
seveibar, line 77: define return type to improve clarity

seveibar:
```suggestion
}): Array<SchematicNetLabelId> =>
```

lib/components/primitive-components/Group/Group_doInitialSchematicTraceRender/applyNetLabelPlacements.ts
```diff
@@ -1,14 +1,103 @@
...
+}: {
+  text: string
+  schPortIds: string[]
+  explicitNetLabelTextsByPortId: Map<string, Set<string>>
+}) =>
+  schPortIds.some((schPortId) =>
+    explicitNetLabelTextsByPortId.get(schPortId)?.has(text),
+  )
+
+const getOverlappingExplicitNetLabelIdsForSameSourceNet = ({
```
seveibar, line 60: explicit concept isn't clear

lib/components/primitive-components/Group/Group_doInitialSchematicTraceRender/snap-to-pins-inside-expanded-bounding-box.ts
```diff
@@ -0,0 +1,119 @@
...
+    .filter((center): center is Point => Boolean(center))
+}
+
+export const snapPointToPinInsideExpandedBoundingBox = (
+  params: {
+    point: Point
+    schematicPortIds: string[]
+    eligiblePortIds: Set<string>
+  },
+  db: CircuitJsonUtilObjects,
```
seveibar, line 53: not proper context pattern

seveibar:
```suggestion
  ctx: { db: CircuitJsonUtilObjects } ,
```

PR #2527, opened 2026-06-30
Lay out schematic sheets independently

MustafaMulla29:


seveibar (changes requested):
Already mentioned this: compute a schematic tree, do not filter the tree. It will simplify the code to get a circuit json schematic tree instead

seveibar (approved):

seveibar:
to me this is much cleaner!

PR #2530, opened 2026-06-30
Fix: repro14 where exposingNets didn't create `pcb_trace` but the schematic showed connection

imrishabh18:
For these net-to-net traces, it created logical source_trace entries, but when building PCB routes it did not follow those nets far enough to discover the chip pins behind them. So it saw no physical points to route between, and produced no pcb_trace.


tests/features/subcircuit-circuit-json/subcircuit-circuit-json14.test.tsx
```diff
@@ -47,32 +47,29 @@ const SubcircuitExample2 = (props: any) => {
...
-        <trace from=".Sub2 > net.SDA" to="net.I2C_SDA" />
+      <trace from=".Sub1 > net.SDA" to="net.I2C_SDA" />
+      <trace from=".Sub2 > net.SDA" to="net.I2C_SDA" />
 
-        <trace from=".Sub1 > net.SCL" to="net.I2C_SCL" />
-        <trace from=".Sub2 > net.SCL" to="net.I2C_SCL" />
-      </board>,
-    )
+      <trace from=".Sub1 > net.SCL" to="net.I2C_SCL" />
+      <trace from=".Sub2 > net.SCL" to="net.I2C_SCL" />
```
seveibar, line 64: interesting syntax! I like it

lib/utils/autorouting/getSimpleRouteJsonFromCircuitJson.ts
```diff
@@ -445,10 +445,30 @@ export const getSimpleRouteJsonFromCircuitJson = ({
...
   for (const net of source_nets) {
-    const connectedSourceTraces = db.source_trace
+    const directlyConnectedSourceTraces = db.source_trace
       .list()
       .filter((st) => st.connected_source_net_ids?.includes(net.source_net_id))
 
+    if (
+      directlyConnectedSourceTraces.length > 0 &&
+      directlyConnectedSourceTraces.every((st) =>
+        st.name?.startsWith("exposed_net."),
```
seveibar: look for these regex hacks like a hawk!!!!! Remove!

seveibar (commented):
be very careful to avoid the regex /string matching, it is ALWAYS a big problem! An explicit boolean can be placed on the net etc. but honestly i don't understand WHY it matters if a net is exposed for SRJ generation! There is a better solution here!

seveibar:
SRJ should not need to know about exposed nets!

seveibar (approved):

PR #2534, opened 2026-06-30
Fix net labels overlapping pins

Sang-it:
Stacked PR 1/3.

Pure deletion in `insertNetLabelsForPortsMissingTrace.ts`: stop moving net label anchor for an already-near net label onto a port center.

The single-port path (`connectedPortCountForKey <= 1`) above it still re-anchors. For multi-pin/shared nets this dragged GND/VCC  labels on top of chip pins.

Before:
Screenshot 2026-06-30 at 11 42 57 AM

After (gnd is here a normal label cause we currently donot allow trace crossing for those)
Screenshot 2026-06-30 at 5 05 47 PM


tests/repros/__snapshots__/repro132-duplicate-multidrop-net-label-schematic.snap.svg
Sang-it, line 1: Fixed in #2535

tests/repros/__snapshots__/repro126-trace-overlap-box-resistor-schematic.snap.svg
Sang-it, line 1: Fixed in #2536

PR #2543, opened 2026-07-01
repro140 - missing net label for passives with long net labels

Sang-it:
Net labels do not show up when the net label text is too long.


imrishabh18 (commented):
@Sang-it Make sure you pull main and follow the repro number for file names, 140 was already there and this caused a duplicate

Sang-it:
Yeah, probably wasn't rebased on main when I added the test. There was no 140 at that time. Will be more careful next time.

PR #2547, opened 2026-07-01
Scope parent net autorouting source traces

0hmX:
## Summary

- Prevent parent-board autorouting from turning already-solved child subcircuit internals into new routing endpoints.
- Preserve explicit `exposedNets` behavior, where parent routing must still be able to reach exposed child nets.
- Add a regression test for the GameBoy/Pico failure mode.

## Cause

`getSimpleRouteJsonFromCircuitJson` builds a Simple Route JSON view for the subcircuit currently being routed.

For parent-board routing it intentionally includes descendant subcircuit elements. That is needed so the parent autorouter can see already-routed child copper as geometry and avoid colliding with it.

The bug was in the net-derived connection builder:

- `source_net`s were scoped to the current subcircuit.
- `source_trace`s were taken from the expanded DB that also contained descendants.
- Connectivity keys then collapsed shared nets such as `GND`, `V3V3`, and `VSYS` across parent and child scopes.

That allowed child-only traces and pads from an already-routed subcircuit to be added to the parent board's `pointsToConnect`.

In the GameBoy repro, the Pico subcircuit routed first, but the parent board route later saw Pico internal power/ground pins as endpoints and tried to route inside the Pico again.

## Fix

Scope the `source_trace`s used for net-derived parent connections.

The filter now includes:

- traces in the current `subcircuit_id`
- traces that directly reference a current-scope source net
- unrouted descendant pin traces only when they are connected through explicit `exposed_net.*` bridge traces

The filter excludes:

- descendant-only internals that merely share a collapsed connectivity key
- descendant traces that are already represented by child `pcb_trace`s

This keeps routed child copper available as geometry/obstacles while preventing it from becoming a new parent routing request.

## Why `exposedNets` Needs Special Handling

Some valid parent routes connect to nets exposed by child subcircuits.

In those cases, the parent-scope bridge traces are named `exposed_net.*` and connect the parent net to the child net. The child pin trace may be the only trace with a physical PCB port, so excluding every descendant trace would break existing exposed-net routing.

The fix therefore allows only the descendant source nets reached through those explicit exposed-net bridge traces, and still ignores them if the child route has already produced a `pcb_trace`.

## Validation

- `bun test tests/features/subcircuit-circuit-json/subcircuit-circuit-json14.test.tsx tests/subcircuits/parent-autorouting-net-scope.test.tsx`
  - `2 pass`
  - verifies existing exposed-net behavior still routes
  - verifies parent autorouting does not include child subcircuit internals
- Downstream GameBoy repro:
  - `bun run build`
  - build passed
  - generated `dist/index/circuit.json`
  - `pcb_trace`: `229`
  - `pcb_via`: `260`
  - `pcb_autorouting_error`: `0`

## Notes

The parent autorouter still receives descendant `pcb_trace` geometry for collision avoidance and subcircuit handoff behavior. The change only limits which descendant `source_trace`s are converted into parent net connection endpoints.

### DIFF
--- a/lib/utils/autorouting/getSimpleRouteJsonFromCircuitJson.ts
+++ b/lib/utils/autorouting/getSimpleRouteJsonFromCircuitJson.ts
@@ -450,8 +450,7 @@ export const getSimpleRouteJsonFromCircuitJson = ({
   const currentSubcircuitSourceTraces = db.source_trace
     .list()
     .filter(
-      (st) =>
-        !subcircuit_id || (st as any).subcircuit_id === subcircuit_id,
+      (st) => !subcircuit_id || (st as any).subcircuit_id === subcircuit_id,
     )
   const exposedDescendantSourceNetIds = new Set<string>()
   for (const st of currentSubcircuitSourceTraces) {
--- a/lib/utils/autorouting/getSimpleRouteJsonFromCircuitJson.ts
+++ b/lib/utils/autorouting/getSimpleRouteJsonFromCircuitJson.ts
@@ -446,7 +446,50 @@ export const getSimpleRouteJsonFromCircuitJson = ({
   const connectionsFromNets: SimpleRouteConnection[] = []
   const connectionFromNetId = new Map<string, SimpleRouteConnection>()
   const handledNetConnectivityKeys = new Set<string>()
-  const sourceTraces = db.source_trace.list()
+  const sourceNetIds = new Set(source_nets.map((net) => net.source_net_id))
+  const currentSubcircuitSourceTraces = db.source_trace
+    .list()
+    .filter(
+      (st) =>
+        !subcircuit_id || (st as any).subcircuit_id === subcircuit_id,
+    )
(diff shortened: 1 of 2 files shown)


lib/utils/autorouting/getSimpleRouteJsonFromCircuitJson.ts
```diff
@@ -446,7 +446,49 @@ export const getSimpleRouteJsonFromCircuitJson = ({
...
-  const sourceTraces = db.source_trace.list()
+  const sourceNetIds = new Set(source_nets.map((net) => net.source_net_id))
+  const currentSubcircuitSourceTraces = db.source_trace
+    .list()
+    .filter(
+      (st) => !subcircuit_id || (st as any).subcircuit_id === subcircuit_id,
+    )
+  const exposedDescendantSourceNetIds = new Set<string>()
+  for (const st of currentSubcircuitSourceTraces) {
+    if (!st.name?.startsWith("exposed_net.")) continue
```
seveibar, line 457: this hack isn't allowed!!!

PR #2550, opened 2026-07-01
Update autorouter, support trace having connected obstacles, remove trace trimming logic, fix subcircuit handling

seveibar:
## What Changed

This is a code-clarity follow-up built on top of #2549.

- Documented the SRJ contract in `getSimpleRouteJsonFromCircuitJson`: `connections` are routing work for the current autorouter, while `traces` are existing copper to preserve.
- Removed the breakout-point route trimming helper for preserved child subcircuit traces.
- Preserved child subcircuit traces now keep their original route geometry and carry `connectedTo` metadata derived from the connectivity map, source trace, attached nets/ports, and route endpoint PCB ports.
- Updated the local SRJ trace type so core can pass `traces[*].connectedTo` through to the updated `@tscircuit/capacity-autorouter`.
- Renamed the local routed-trace bookkeeping so it is clear which `source_trace`s are already represented by preserved SRJ traces.
- Renamed `_createTracesForExposedNets` to `_createTracesForExposedConnections`.
- Added `Trace._exposesSubcircuitConnection` and set it on generated exposed-connection helper traces, so re-render protection only matches traces created by this helper.
- Removed generated `name` values from exposed-connection helper traces. They now keep only `displayName`, because these traces are generated routing intent, not user-named children.
- Updated subcircuit duplicate-name DRC so immediate non-subcircuit children generally need unique names, but same-named traces are allowed when every duplicate has the same non-empty `subcircuit_connectivity_map_key`.
- Added focused unit coverage for duplicate trace-name connectivity keys and preserved SRJ trace connectivity metadata.

## Why

The subcircuit/SRJ interaction is easy to misread because the same circuit-json data can represent two different things:

- a child subcircuit route that already exists and should be preserved as SRJ `traces`
- parent-board routing intent that still needs to be emitted as SRJ `connections`

With `traces[*].connectedTo` support in the autorouter, core no longer needs to trim preserved trace geometry around breakout points to make parent routing work. The preserved trace stays geometrically accurate, and the autorouter gets the same-net relationship explicitly through SRJ connectivity metadata.

For naming, the invariant is:

- generated exposed-connection helper traces should not set `name`
- user-named immediate children should not collide across different things
- same-named traces may coexist only when the connectivity map proves they are the same underlying connection
- nested subcircuits remain separately scoped

## Validation

- `./node_modules/.bin/biome format --write lib/utils/autorouting/getSimpleRouteJsonFromCircuitJson.ts lib/utils/autorouting/SimpleRouteJson.ts tests/utils/autorouting/simple-route-json-ignores-routed-subcircuit-traces.test.ts`
- `bun test tests/utils/autorouting/simple-route-json-ignores-routed-subcircuit-traces.test.ts tests/components/primitive-components/group-duplicate-trace-name-connectivity-key.test.ts`
- `bunx tsc --noEmit` still fails on existing unrelated baseline errors around graph display props, `source_unnamed_trace_warning`, and the `circuit-to-svg` `convertCircuitJsonToStackedSchematicSheetsSvg` export.

Previously attempted:

- `bun test tests/subcircuits/subcircuit-exposed-nets-copper-pour01.test.tsx`

The fixture-backed exposed-net test is still blocked at module load by the existing `circuit-to-svg` export issue: `convertCircuitJsonToStackedSchematicSheetsSvg` is not exported.


seveibar:
@Abse2001 this is a derivative of your PR- the main thing i did was updated the autorouter to support prerouted traces being connected to pointsToConnect, i believe this will fix the failure you ran into (i believe that you were getting static reachability failures because it was marking mutually connected traces as obstacles)- the downside is the autorouter isn't properly simplifying traces, which we have repro'd and @0hmX is aware of it

i want to merge this with the minor regression in trace simplification so that we have the subcircuit fix, you can try it out(?) and see if the traces aren't too messed up, and hopefully @0hmX can follow up with the fix for the traces not simplifying properly when approximating rects are being used

seveibar:
Also CC @imrishabh18 & @MustafaMulla29 

@imrishabh18 this should fix the subcircuit issue you ran into

@MustafaMulla29 this removes the trace trimming you introduced by fixing a bug in the autorouter allowing pointsToConnect to be within obstacle (preexisting) traces

I'm eager to merge this because the trace simplification issue seems very fixable but the subcircuit issue i want to make sure gets done properly. I BELIEVE this PR fixes the subcircuit trace-sharing issue in a pretty good way- there is an explicit carve-out in the SRJ handling for bringing over pre-existing traces. That seems fine to me!

PR #2551, opened 2026-07-02
bump schematic-trace-solver to 0.0.75

Sang-it:


seveibar (approved):

seveibar:
- RP2040 SDA label
- V5 moving further than necessary
- junction bug

PR #2555, opened 2026-07-02
fix(sch): create cross-sheet net labels per schematic sheet

MustafaMulla29:
For a connection spanning multiple schematic sheets (e.g. a shared I2C bus wired with direct pin-to-pin traces), the auto-generated cross-subcircuit net label collapsed to a single label for the whole net instead of one per sheet — and that
label was stamped with the rendering group's sheet while positioned at a pin from a different sheet. So a controller bus label like `U1_PA0` showed up on the *sensor* sheet, and the other sheets on the bus got no label at all.

## Fix
Made net-label creation/de-duplication in `insertNetLabelsForPortsMissingTrace` sheet-aware: each port's sheet is resolved from its own `schematic_component`, and that sheet id is used for the dedup lookups, the orphan-cleanup, and the inserted label's `schematic_sheet_id`. Every sheet a net reaches now gets its own label anchored to its own pin. Single-sheet circuits are unchanged.


seveibar (approved):

seveibar:
Note schematic net labels that span multiple sheets are especially marked in traditional schematic software, will want to adjust the circuit Json  specification to do something similar

PR #2557, opened 2026-07-02
update auto router to 0.0.624

0hmX:
not changing any default just an update


imrishabh18 (changes requested):
Image

But this is making it worse, right? Rather than updating to the latest, update to the version which is not breaking this

Abse2001 (approved):

Abse2001 (commented):
nvm we can't merge this something is wrong

tests/examples/__snapshots__/example41-solar-battery-charger-pcb.snap.svg
0hmX, line 1: Okay, so this is bad, but not as bad as it looks. The behavior is actually more correct now.

Previously, because of how the BGA was generated, single-layer obstacles were being converted into multi-layer obstacles. So a top-only obstacle on a 4-layer board was getting converted into an all-layer obstacle.

That caused the via to be pushed farther outside. The regions are also getting a high penalty on the inside side, though I’m not sure why.

before: the new bga solver
Screenshot from 2026-07-03 01-07-47
Screenshot from 2026-07-03 01-07-27
Screenshot from 2026-07-03 01-06-50

after the new bga solver that merge:
Screenshot from 2026-07-03 01-08-36
Screenshot from 2026-07-03 01-09-00
more search state on the inside is possible which auto router is using 
Screenshot from 2026-07-03 01-09-26

Also, this test is only for Pipeline 7. The best fix is to avoid creating multilayer regions when the region size is too small compared to the via diameter.

I’ll implement that, but can we merge this so @seveibar’s fix for the autorouter can go through to the CLI and @Abse2001 can use it?

imrishabh18: A merge is kind a okay, but long will you take to fix this? If that could be done in a day, maybe we wait it out?

0hmX: Yes, my clanker is working. I’m just rushing because of the Game Boy ordering and stuff idk

0hmX: the cluster gone i will fix the reset iteratively too much

imrishabh18 (approved):

PR #2559, opened 2026-07-02
fix(schematic): resolve netlabel net connections to the net's ports (floating power net label)

### LINKED ISSUE #3759: Floating power net label
[Package code to reproduce](https://tscircuit.com/editor?#data:application/gzip;base64,H4sIAH3kPWoAA7WSMW/CQAyF/4rlCaQKEio6IILUItYOqOp0y5GY5tTEF11cNVKU/16ngaZbKwGb/d6z/Q1u0XFGzUzqBldITeWDQEZH+1EITKaQbGBiGGB98DZk8OkyyRODy6gsDUJO7i2Xn37TJzUbqHa1+DC0AGxL0tA+NniWhojltDfi99E4ei9VcNxvjR6i+9FJPTOl4jzXSdtC5ThegUEmmT0avOuFxVnY7p5fdnsl7LphfP4X2+JWbE+Xs/06dF227f/Z1C3sgQrQQk+8LjU0blbltO80sZ5/P4zWU8PYfQH25fKJZwIAAA==)

I added a V5 net label connect to R3 via a common net, why is there a V5 port over R1?

Image

It seems to be only with power ports because it works fine with other net labels:

Image

chayan-1906:
Fixes tscircuit/tscircuit.com#3759

## Problem

A power net label connected through a net renders a phantom rail on top of an unrelated component:

```tsx
<resistor name="R3" ... connections={{ pin1: "net.C", pin2: "net.CENTER" }} />
<netlabel net="V5" connection="net.C" />
```

The V5 rail correctly appears at R3.pin1, but a second, floating V5 rail is drawn on top of R1 (whichever component sits at schematic origin).

## Root cause

`NetLabel._getConnectedPorts()` resolves the `connection` selector with `selectOne(connection)` and casts the result to `Port`. For a net selector like `"net.C"` this returns the **Net** instance, which has no schematic position, so the user's label anchors at the fallback `(0, 0)`.

For power/ground nets the schematic trace solver then places its own rail at the real pin (`getSchematicPortIdsWithAssignedNetLabels` can't see that the pin already has a user-defined label, for the same reason), routes a stub trace to it, and marks the port connected — so `insertNetLabelsForPortsMissingTrace` never rescues the stranded user label. Result: a correct rail at R3.pin1 plus a floating duplicate at `(0, 0)`.

Non-power labels were silently rescued by the missing-trace pass, which is why only power/ground rails visibly floated — though they still produced two identical labels stacked at the pin.

## Fix

Resolve net selectors to the net's connected ports (`Net.getAllConnectedPorts()`) in `_getConnectedPorts()`. The user label then anchors at the real pin, and the existing solver logic skips both the duplicate placement (`applyNetLabelPlacements`) and the stub trace (`applyTracesFromSolverOutput`).

Before (issue repro): two `V5` labels — `(0.00, 0.00)` on top of R1 + `(-0.65, 4.36)` near R3.
After: one `V5` label at `(-0.55, 3.96)` = R3.pin1, `rail_up` symbol kept.
Regular labels (e.g. `net="MYNET"`): same position as before, no longer duplicated.

## Testing

- New repro test `tests/repros/repro142-floating-power-net-label.test.tsx` — fails on main (phantom anchors 3.99 units away from R3.pin1), passes with the fix
- All netlabel tests pass (24 tests across 23 files)
- Full suite: 1131 pass / 28 skip; the only failure (`autorouter-preset-laser-prefab`) also fails on unmodified main (pre-existing)
- `bunx tsc --noEmit` clean, `biome format` clean, `bun run build` succeeds


chayan-1906:
Keeping this active — ready for review whenever someone gets a chance. It's a small root-cause fix (+8 lines in `NetLabel._getConnectedPorts()`) with a repro test for the floating power net label bug (tscircuit/tscircuit.com#3759). All 10 checks green, with no conflicts. Happy to adjust anything.

chayan-1906:
@seveibar Could you or someone on the team take a quick look when you get a chance? The stale bot keeps flagging this while it sits with all checks green. It's a small root-cause fix (+8 lines) for tscircuit/tscircuit.com#3759 — floating power net label — with a repro test included. Happy to adjust anything.

PR #2568, opened 2026-07-03
Update capacity-autorouter version to 0.0.625

0hmX:


tests/repros/__snapshots__/repro56-rotate-group-pcb.snap.svg
0hmX, line 1: Image

0hmX: i think it improve but could be wrong

imrishabh18 (approved):

seveibar (approved):

PR #2569, opened 2026-07-03
improve diode aliasing for cathode and anode, support pin1 != anode

seveibar:


tests/components/normal-components/__snapshots__/diode-sod123-regular-vs-kicad.snap.svg
```diff
@@ -0,0 +1,32 @@
+<svg width="800" height="1292" viewBox="0 0 800 1292" xmlns="http://www.w3.org/2000/svg">
```
seveibar, line 1: CC @Abse2001

PR #2574, opened 2026-07-04
Propagate board solderMaskColor/silkscreenColor to pcb_board

### LINKED ISSUE #3277: <board> prop `solderMaskColor` silently ignored — not propagated to circuit-json or 3D renderer
## Summary

The `solderMaskColor` prop on `<board>` (and the related per-face / silkscreen color props) is defined in `@tscircuit/props` `boardProps` but never reaches `circuit-json`, so the 3D viewer always renders the default green board regardless of the value.

The prop is accepted at the JSX layer (no error), giving the impression it works.

## Reproduction

```tsx
export default () => (
  <board
    width="40mm"
    height="40mm"
    outline={/* 40mm circle */}
    thickness="0.8mm"
    solderMaskColor="black"
  />
)
```

Build with `tsci build entry.circuit.tsx --3d`, then inspect:

1. The generated `circuit.json` `pcb_board` element:
   ```json
   {
     "type": "pcb_board",
     "thickness": 0.8,
     "material": "fr4",
     "num_layers": 2,
     ...
   }
   ```
   No `solder_mask_color` (or equivalent) field. `thickness` and `material` do propagate; the color does not.

2. The rendered `3d.png` is green even though `solderMaskColor="black"` was set.

## Expected

`solderMaskColor="black"` should result in either:

- A field on `pcb_board` (e.g. `solder_mask_color: "black"`) that the 3D renderer picks up, or
- An explicit "not yet supported" error at build time so users know the prop is a no-op.

## Affected packages

- `@tscircuit/props` — defines the prop in `boardProps` ✅
- `@tscircuit/core` — does NOT copy the prop into the `pcb_board` circuit-json element ❌
- `circuit-json` (`pcb_board` schema) — has no `solder_mask_color` field to receive it ❌
- `@tscircuit/3d-viewer` — has nothing to read, falls back to default green ❌

## Versions

- `tscircuit` CLI: 0.0.1786 (latest at time of filing)
- `@tscircuit/core`: latest pulled with above
- `@tscircuit/props`: latest pulled with above
- Bun 1.3.14 / Node 24, Windows 11

## Why this matters

For multi-board stacks or visual previews of color-coded designs (e.g. a poker-chip-form-factor module rendered in 3D for review), the color prop is the only mechanism users expect to control PCB appearance. Silent failure is hard to debug.

## Suggested fix

Either:

1. Add `solder_mask_color`, `top_solder_mask_color`, `bottom_solder_mask_color`, `silkscreen_color`, `top_silkscreen_color`, `bottom_silkscreen_color` fields to the `pcb_board` element in `circuit-json`, then wire `@tscircuit/core` to copy the corresponding `boardProps` values through, then have `@tscircuit/3d-viewer` honor them.
2. Or, if not planned, emit a build-time warning when these props are set.

Happy to PR if scope confirmed.

abdalraof-albarbar:
Closes tscircuit/tscircuit#3277. Repro: #2573.

The `<board>` `solderMaskColor`/`silkscreenColor` props typecheck and render fine, but `@tscircuit/core` never copied them into the `pcb_board` circuit-json element, so the value was a silent no-op (identical circuit-json regardless of the prop). `circuit-json`'s `pcb_board` already has `solder_mask_color`/`silkscreen_color` fields, so this just wires them through in `Board.ts` next to `material`. Flips the repro test to passing; sibling board tests unaffected.

### DIFF
--- a/lib/components/normal-components/Board.ts
+++ b/lib/components/normal-components/Board.ts
@@ -525,6 +525,12 @@ export class Board
         y: point.y + (props.outlineOffsetY ?? 0) + outlineTranslation.y,
       })),
       material: props.material,
+      ...(props.solderMaskColor !== undefined && {
+        solder_mask_color: props.solderMaskColor,
+      }),
+      ...(props.silkscreenColor !== undefined && {
+        silkscreen_color: props.silkscreenColor,
+      }),
 
       min_trace_width:
         subcircuitProps.minTraceWidth ?? jlcMinTolerances.min_trace_width,
--- a/tests/components/normal-components/board-solder-mask-color.test.tsx
+++ b/tests/components/normal-components/board-solder-mask-color.test.tsx
@@ -1,24 +1,21 @@
 import { test, expect } from "bun:test"
 import { getTestFixture } from "tests/fixtures/get-test-fixture"
 
-test.failing(
-  "board solderMaskColor and silkscreenColor reach the pcb_board",
-  async () => {
-    const { circuit } = getTestFixture()
+test("board solderMaskColor and silkscreenColor reach the pcb_board", async () => {
+  const { circuit } = getTestFixture()
 
-    circuit.add(
-      <board
-        width={20}
-        height={20}
-        solderMaskColor="black"
-        silkscreenColor="white"
-      />,
-    )
+  circuit.add(
+    <board
+      width={20}
+      height={20}
+      solderMaskColor="black"
+      silkscreenColor="white"
+    />,
+  )
 
-    await circuit.renderUntilSettled()
+  await circuit.renderUntilSettled()
 
-    const pcb_board = circuit.db.pcb_board.list()[0]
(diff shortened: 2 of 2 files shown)


abdalraof-albarbar:
Heads up on CI: `test (3)` failed on two ngspice simulation tests (`ngspice-spice-analysis07-opamp-half-wave-rectifier`, `ngspice-graph-independent-axes-auto`) — these are unrelated to this change and look like ngspice flakiness: they **pass locally on this branch**, pass on `main`, and this change is a no-op unless a board sets `solderMaskColor`/`silkscreenColor` (conditional spread, so no existing `pcb_board` snapshot changes). `test (1/2/4)`, type-check, format, and smoke-test are all green. A re-run of that shard should clear it.

seveibar (approved):

seveibar:
@copilot resolve the merge conflicts in this pull request

PR #2580, opened 2026-07-04
Add pinLabels support to switch components

Abse2001:
>


seveibar (approved):
How do you configure the symbol? The usage seems unrealistiv with symbol configuration

seveibar:
Maybe pinLabels isnt appropriate for this

seveibar:
pinLabels : { COM1: [“pin1”] }

makes more sense to me, and we disallow pin number based configuration

seveibar:
Althought, pinAttributes would work for this actually- or pin1: [“COM1”, “OTHER_LABEL”]

Then we have a symbol to label matching thing

seveibar:
So it might be ok but still bad test usage i think- the symbol mapping should be clear

PR #2585, opened 2026-07-06
fix(autorouter): wire autorouter traceClearance into the capacity router

maci0:
## Problem

Setting `autorouter={{ traceClearance: ... }}` on a board/subcircuit has **no effect** — traces are routed at the same spacing regardless. The value is written to the pcb_group's `autorouter_configuration.trace_clearance` (Group.ts) but is never passed to the autorouter, so the capacity mesh router always falls back to its default obstacle margin.

## Fix

`SimpleRouteJson` already has a `defaultObstacleMargin` field — the trace-to-obstacle spacing the capacity solver reads (`obstacleMargin ?? srj.defaultObstacleMargin`). It just wasn't being set. This threads it through:

- `getSimpleRouteJsonFromCircuitJson` gains a `defaultObstacleMargin?` param and emits it on the returned `SimpleRouteJson`.
- `Group` passes the autorouter's `traceClearance` (parsed from `Distance` to mm) as `defaultObstacleMargin`, for both the local capacity autorouter and the legacy solve-endpoint path.

## Test

Adds `simple-route-json-default-obstacle-margin.test.tsx` asserting the field is forwarded when set and `undefined` otherwise. Neighboring autorouting tests still pass; typecheck clean.

### DIFF
--- a/lib/components/primitive-components/Group/Group.ts
+++ b/lib/components/primitive-components/Group/Group.ts
@@ -760,6 +760,10 @@ export class Group<Props extends z.ZodType<any, any, any> = typeof groupProps>
                 db,
                 minTraceWidth: Number(props.minTraceWidth ?? 0.15),
                 nominalTraceWidth: this.props.nominalTraceWidth,
+                defaultObstacleMargin:
+                  this.props.autorouter?.traceClearance != null
+                    ? distance.parse(this.props.autorouter.traceClearance)
+                    : undefined,
                 subcircuit_id: this.subcircuit_id,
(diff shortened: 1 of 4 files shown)


maci0:
Reopen request: stale-bot closed, not on merit. Wiring `traceClearance` into the capacity router is needed for clearance fidelity between the declared design rule and what actually gets routed. CI-green; branch intact.

maci0:
Not rebasing. Same work is in open #3788 (forwards `traceClearance` into `defaultObstacleMargin`, including group config).

PR #2588, opened 2026-07-06
fix(schematic): only pass net label orientations for power and gnd

Sang-it:
Only pass net label orientation for power and gnd.

## Why?

To reduce redundant label orientation flips. 

Example:

Screenshot 2026-07-06 at 5 49 44 PM
Screenshot 2026-07-06 at 5 49 33 PM


imrishabh18 (commented):
If we are passing all the available direction, why not remove the `availableNetLabelOrientations` from the input problem and handle everything in the solver

Sang-it:
> If we are passing all the available direction, why not remove the `availableNetLabelOrientations` from the input problem and handle everything in the solver

we need ```availableNetLabelOrientations``` for power and gnd, we could maybe not pass it for other nets.

let me try that approach.

seveibar (commented):
there's more losses than wins from this

seveibar (changes requested):
not sure why we would do this, i don't see much gain, i see lots of harder to read labels. Is there a reference where this is common? Typically signal netlabels are horizontal

PR #2589, opened 2026-07-06
Preserve explicit narrow autoroute trace widths

ShiboSoftwareDev:
Set default widths on autorouter connections before lowering the SRJ minTraceWidth, so explicit narrow traces can route below the board default without causing unannotated traces to shrink. Adds coverage for a center-pad narrow trace repro and keeps default-width routing behavior intact.


lib/components/primitive-components/Group/get-autorouted-trace-connection.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export function getAutoroutedTraceConnection({
+  trace,
+  connectionByName,
+}: {
+  trace: SimplifiedPcbTrace
+  connectionByName: Map<string, SimpleRouteConnection>
+}): SimpleRouteConnection | undefined {
+  return (
+    connectionByName.get(trace.connection_name ?? "") ??
```
0hmX: Throw error if not found

0hmX (approved):

lib/components/primitive-components/Group/get-autorouted-trace-connection.ts
ShiboSoftwareDev: If we don’t find a connection, the current fallback to point.width seems safer than failing valid output

seveibar:
@0hmX @ShiboSoftwareDev i suspect this fix isn't correct

I think the trace width is supposed to "narrow" as it reaches the pads- that is how the autorouter outputs intentionally

seveibar (changes requested):
i surprised this was approved @0hmX ! 

There are DRC violations as a result of this change

image

lib/utils/autorouting/getSimpleRouteJsonFromCircuitJson.ts
```diff
@@ -627,7 +632,10 @@ export const getSimpleRouteJsonFromCircuitJson = ({
           ? preservedRoutedSubcircuitTraces
           : undefined,
       layerCount: board?.num_layers ?? 2,
-      minTraceWidth: minTraceWidth ?? board?.min_trace_width ?? 0.1,
+      minTraceWidth: Math.min(
+        defaultTraceWidth,
+        ...allConns.map((c) => c.width!),
```
seveibar, line 637: it feels like the autorouter should handle traces below minTraceWidth if it's explicitly passed in for the trace

seveibar (approved):
hesistant acceptance, see comment

PR #2590, opened 2026-07-06
Add new schematic repro (repro-144)

AnasSarkiz:
Screenshot 2026-07-06 at 7 45 44 PM


seveibar (commented):
this repro would be better in the schematic-trace-solver repo

PR #2591, opened 2026-07-06
Add repro for GND net label text too close with pins

techmannih:
image


MustafaMulla29 (approved):

imrishabh18 (commented):
@techmannih It's a duplicate repro 143 already has this

PR #2593, opened 2026-07-06
add repro for GND netlabel overlap on horizontal schematic trace

techmannih:


tests/repros/repro145-gnd-label-overlaps-horizontal-trace.test.tsx
MustafaMulla29, line 1: As said by seve no manual netlabels

PR #2599, opened 2026-07-07
update schematic symbols

mohan-bee:
i am trying to update the schematic-symbols one version by one, because updating to latest causing 100+ snapshots which will be hard to review !!1


tests/repros/__snapshots__/repro126-trace-overlap-box-resistor-schematic.snap.svg
mohan-bee, line 1: kind of funny how trace is draws here in inductor in small distance ,

mohan-bee, line 1: something is odd here 2 netlabels connected to the inductor ?? 🤔

PR #2600, opened 2026-07-07
Fix duplicate net label on unconnected pins with text-expanded bounds

MustafaMulla29:
A pin connected only through a net label (e.g. `R_FAULT_PULLUP.pin2` → `net.FAULT`) could render **two** overlapping net labels for the same port — one floating just off the pin, one anchored on it:

## Root cause

`createSchematicTraceSolverInputProblem` hands the schematic-trace-solver each pin at its real `schematic_port.center`, but hands it a chip box **expanded to fit the component's text** (`getSchematicComponentWithTextBounds`). For a small part with a long name, that box extends past the pin, so the solver anchors the pin's net label at the **box edge**, slightly off the pin.

- When the pin's connection is already drawn (routed trace / power symbol), the gap
  is bridged and it looks fine.
- When it isn't, the solver label **floats**, and the fallback pass
  `insertNetLabelsForPortsMissingTrace` adds a **second**, pin-anchored label for the
  same port. Its position-based dedupe (0.1mm) missed the floating one, so both
  survived.

Core already corrects this exact "pin ends up inside the text-expanded box" problem for **trace endpoints** (`extendTraceEndpointsToReachPinsInsideExpandedBoundingBox` in `applyTracesFromSolverOutput`); net-label anchors just never got the same treatment.

## Fix

In `applyNetLabelPlacements`, snap a **single-pin** net label's anchor onto the schematic port center when the pin is **not** `is_connected` — mirroring the existing trace-endpoint snapping. This makes the label connect to the pin and land exactly where `insertNetLabelsForPortsMissingTrace` anchors, so that pass's existing dedupe drops the duplicate. No dedupe thresholds were loosened.

The `!is_connected` guard is the same signal that pass uses to decide which ports it labels, so already-drawn connections (e.g. a power symbol placed deliberately off the pin) keep the solver's intentional edge placement — no regression there.

Before:
Screenshot 2026-07-07 at 7 52 53 PM

After:
Screenshot 2026-07-07 at 7 53 18 PM


seveibar:
@MustafaMulla29 again, i am SEARCHING snapshots-

tests/repros/__snapshots__/repro145-atmega328p-many-net-labels-schematic.snap.svg
MustafaMulla29, line 1: @seveibar this 
Screenshot 2026-07-07 at 7 44 08 PM

seveibar (approved):

Issue #1507, opened 2025-10-14
Fix repro62 packing issue - Packing over obstacles from "outside" the packing group


seveibar:
https://github.com/tscircuit/core/pull/1506

Probably requires passing obstacles to `calculate-packing`

<img width="1114" height="1198" alt="Image" src="https://github.com/user-attachments/assets/72aa764c-bf71-42c1-b95a-2464a6d84e39" />

/bounty $125

PR #2601, opened 2026-07-07
fix(autorouter): handle unbroken copper pour correctly (skip its nets + fill, not obstacle)

maci0:
An unbroken `<copperpour>` was being mishandled by the autorouter input builder in two ways. Both are needed to route a board that has a ground pour.

## 1. Don't route nets the pour already connects
`getSimpleRouteJsonFromCircuitJson` emitted a routing connection for every `source_net`, including nets whose connectivity is fully provided by the pour (typically GND). The autorouter then tried to route that net pad-to-pad; krt aborts (`KRT GridRouter found no route for connection source_net_2`), capacity-mesh wastes effort. Fix: collect the net keys the unbroken pours connect and skip those nets.

## 2. Realize the pour as fill, not a routing obstacle
`getUnbrokenCopperPourObstacles` was pushed straight into the routing obstacle set, so the pour became a hard keepout for **every other net**. But a pour conforms around traces (it is filled after routing, keeping its `traceMargin` clearance) - it must not block routing. On a 2-layer board with a GND pour on **both** layers this left no legal path for any signal: the static-reachability precheck correctly reported `N route(s) have no legal path` and the solver never converged. Fix: keep using these obstacles only to derive the poured-net keys (part 1); do not add them as keepouts.

## Result (breath-ketone board: nRF52840 module + BME690 + OLED + USB-C, dual GND pour)
- Before: precheck-abort / 0 traces / solver runs out of iterations.
- After: **clean route in ~7s, precheck passing, 505 traces, 0 clearance violations** (default effort, no other changes).

Related: the oversized SimpleRouteJson those pour obstacles produced is separately fixed by #2603 (connectedTo dedupe).

### DIFF
--- a/lib/utils/autorouting/getSimpleRouteJsonFromCircuitJson.ts
+++ b/lib/utils/autorouting/getSimpleRouteJsonFromCircuitJson.ts
@@ -121,14 +121,31 @@ export const getSimpleRouteJsonFromCircuitJson = ({
     ),
     sharedConnMap,
   )
-  obstacles.push(
-    ...getUnbrokenCopperPourObstacles({
-      connMap: sharedConnMap,
-      subcircuitComponent,
-      board,
-      group: pcbGroup,
-    }),
-  )
+  // Unbroken copper pours come in two routing roles, split by layer:
+  //
+  // - OUTER-layer pours (top/bottom) are conforming ground fills: the pour is
+  //   filled AFTER routing and flows around the routed traces (see CopperPour's
+  //   get-trace-obstacles / generate-and-insert-brep). They do not exist as
+  //   copper at routing time, so they must NOT be emitted as hard routing
+  //   obstacles - a 2-layer board with a full GND pour otherwise becomes
+  //   unroutable (every other-net trace over the pour area is blocked). The
+  //   pour itself provides the net's connectivity, so its nets are also
+  //   skipped as routing connections (see pouredNetConnectivityKeys below).
+  //
+  // - INNER-layer pours are solid planes reached by escape vias. They keep the
+  //   existing behavior: emitted as obstacles (the escape-via machinery uses
+  //   them as targets) and their nets still route, producing the vias.
+  const unbrokenCopperPourObstacles = getUnbrokenCopperPourObstacles({
+    connMap: sharedConnMap,
+    subcircuitComponent,
+    board,
+    group: pcbGroup,
+  })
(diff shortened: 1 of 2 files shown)


maci0:
Rebased onto v0.0.1401 (the file was reworked upstream, so this is a rewrite of the same two changes on the current code).

Validation, all on current main:

- **Motivating board** (2-layer, unbroken GND pour both layers, 30 components): main routes it with **53 unrouted signal pairs** (VBUS/V3V3/SDA/USB all dead). With only the pour-not-obstacle change: 0 unrouted, but **19 DRC violations** appear because GND is now trace-routed pad-to-pad (copper dragged across NPTH mounting holes and the board edge). With both changes: **0 unrouted, 0 DRC violations**.
- `simple-route-json-unbroken-copper-pour-obstacles` (the PR's test) and `repro-simple-route-json-45-degree` pass.
- Pour test suite (`example29`, `example37`, `board-copper-pour-covered-with-solder-mask`, `subcircuit-exposed-nets-copper-pour01`): same pass/fail profile as clean main (the 3 pre-existing local failures are unchanged), i.e. regression-neutral.
- `example39` (inner-layer escape vias) fails identically before and after this change in my local source run, so it is not affected by this PR.

maci0:
Reworked the fix to be layer-aware, which resolves the `example39-unbroken-copper-pour-escape-vias` failure the earlier version caused:

- **Outer-layer pours (top/bottom)** are conforming fills: filled after routing, flowing around traces. They are no longer emitted as routing obstacles, and their nets are skipped (the pour is their copper). This is what un-blocks the 2-layer full-GND-pour board (53 unrouted pairs → 0, and the 19 DRC violations from trace-routed GND → 0).
- **Inner-layer pours** keep the existing behavior unchanged: solid planes reached by escape vias, so they stay obstacles (escape-via targets) and their nets still route. example39 semantics are untouched.

The PR test now asserts both halves (outer pour absent from obstacles, inner pour present). Formatted with the repo biome config.

maci0:
Reopening request: this was auto-closed by the 1-day stale bot, not on review. The fix (skip poured nets as hard obstacles in the capacity router) is still valid and CI-green, and relevant to the same-layer-short problem tracked in #1507 / autorouter#1513. Could a maintainer reopen? Head branch (fix/skip-poured-nets-in-autorouter) is intact.

maci0:
Reopened against current main: https://github.com/tscircuit/core/pull/

PR #2603, opened 2026-07-07
fix(autorouter): dedupe obstacle.connectedTo to stop SimpleRouteJson bloat

maci0:
## Problem
In `getSimpleRouteJsonFromCircuitJson`, the connectivity-expansion pass pushes `getIdsConnectedToNet(id)` onto every obstacle's `connectedTo` with **no dedupe**:

```ts
for (const obstacle of obstacles) {
  const additionalIds = obstacle.connectedTo.flatMap((id) =>
    sharedConnMap.getIdsConnectedToNet(id),
  )
  obstacle.connectedTo.push(...additionalIds)   // <- duplicates accumulate
}
```

`getIdsConnectedToNet` returns heavily overlapping sets, so the array explodes. On a real 2-layer board with a GND `<copperpour>`, each pour rect's `connectedTo` held **21,171 entries with only 191 unique** (each id repeated ~110×). The pour tessellates into a few hundred obstacles, so the emitted `SimpleRouteJson` reached **~78 MB**, and the autorouter then stalled (`TinyHyperGraphSolver ran out of iterations`).

## Fix
Collect into a `Set` instead of pushing raw. Connectivity is identical; only the duplicates are removed.

## Impact (measured on the same board)
- `connectedTo` per pour obstacle: **21,171 -> 191**
- SimpleRouteJson: **~78 MB -> <1 MB**

No behavior change beyond the size reduction - the deduped IDs are the same set.

### DIFF
--- a/lib/utils/autorouting/getSimpleRouteJsonFromCircuitJson.ts
+++ b/lib/utils/autorouting/getSimpleRouteJsonFromCircuitJson.ts
@@ -149,11 +149,19 @@ export const getSimpleRouteJsonFromCircuitJson = ({
   })
 
   // Add every equivalent ID from the shared connectivity map to each obstacle.
+  // Dedupe the result: getIdsConnectedToNet returns heavily overlapping sets,
+  // so pushing them raw makes connectedTo explode. A GND copper pour on a real
+  // board produced 21,171 entries with only 191 unique (each id repeated ~110x)
+  // per pour rect; across a few hundred pour obstacles that ballooned the
+  // SimpleRouteJson to ~78 MB and stalled the autorouter. Deduping keeps it small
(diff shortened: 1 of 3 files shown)


maci0:
Reopen request: auto-closed by the 1-day stale bot, not on review. Deduping `obstacle.connectedTo` materially shrinks SimpleRouteJson on pour-heavy boards (fewer redundant same-net obstacle entries), which helps the capacity router terminate. CI-green; head branch intact. Could a maintainer reopen?

maci0:
Already on main: obstacle `connectedTo` is built with `new Set([...obstacle.connectedTo, ...additionalIds])`. No rebase.

PR #2607, opened 2026-07-07
Fix schematic trace endpoints stopping short of connected symbol ports

mohan-bee:
Fixes a schematic trace routing correctness  bug where solved traces could stop at a  component obstacle edge instead of reaching  the actual connected schematic port center,  causing visually disconnected pins even though  the netlist was correct. The fix makes endpoint snapping  apply to the actual ports associated with each  solved trace, so nearby axis-aligned endpoints  are extended to their connected pin centers  without touching unrelated pins. 

BEFORE: 

image

image

AFTER: 

image

image


tests/components/primitive-components/__snapshots__/schematic-rect-trace-using-connection-prop-schematic.snap.svg
imrishabh18, line 1: Why are these moving to the incorrect location?

seveibar (changes requested):
probably not the right fix, discuss with whoever wrote the original code but also i think the symbols probably have incorrect port positions and that's the real fix

MustafaMulla29 (changes requested):
This is a symbol issue afaik!

PR #2609, opened 2026-07-07
update schematic-symbols

mohan-bee:


tests/repros/__snapshots__/repro47-switch-matrix-diodes-schematic.snap.svg
mohan-bee, line 1: this is bad , idk is this coming  from trace solver or ??

imrishabh18: Image

Yes look into the traces solver, why there are different route pattern for both of these. It should have been like the bottom one

imrishabh18 (approved):

PR #2612, opened 2026-07-08
fix(autorouter): keep copper-to-edge clearance from cutouts

maci0:
## Problem

A `<cutout>` becomes a board edge (Edge.Cuts) in the fab output, so routed copper must keep the copper-to-edge clearance from it - not merely avoid overlapping it. `getObstaclesFromCircuitJson` emitted cutout obstacles at their exact physical size, and the autorouter routes flush against obstacle boundaries, so traces hug the slot and fail edge-clearance DRC at any fab.

Observed on a real 2-layer board with a 15 x 1.2mm thermal-isolation slot: 9 `copper_edge_clearance` violations (KiCad DRC with JLCPCB's rule set), traces as close as 0.05mm to the slot wall. The current workaround is pairing every cutout with a hand-sized keepout.

## Fix

Inflate cutout obstacles by 0.3mm on every side, for all three shapes (rect, circle, polygon). 0.3mm covers common fab edge-clearance rules (JLCPCB asks 0.3 for 2-layer).

## Tests

- New: `cutout-obstacle-edge-clearance` asserts the emitted obstacle is larger than the physical slot.
- `cutout-routing` PCB snapshot updated: traces now clear the slots (that is the behavior change).


maci0:
Reopen request: stale-bot closed after 1 day, not reviewed. Keeping copper-to-edge clearance from cutouts is needed whenever a board outline has interior cutouts or edge features. CI-green; branch intact.

maci0:
Reopened against current main as a fresh PR (this one was stale-bot closed).

maci0:
New PR: https://github.com/tscircuit/core/pull/3911

PR #2614, opened 2026-07-08
Pass schematic port facing directions to the trace solver

MustafaMulla29:
## What

When building the trace-solver input, forward each schematic port's true facing direction as `_facingDirection` (mapped from `schematic_port.facing_direction` via the existing `convertFacingDirectionToElbowDirection`).

Why

The chip box we hand the solver comes from getSchematicComponentWithTextBounds, i.e. it's grown to enclose the component's reference-designator/value text. For a small part with a large ref designator (a resistor labeled R_FAULT_PULLUP, a decoupling cap, …) that box is much bigger than the part, so the component's ports sit inside the box. The solver has to move an inside-pin to a box edge, and with no facing information it guessed from geometry — picking the nearest edge, which for a horizontal resistor whose wide ref text widened the box is the bottom. Both terminals were treated as facing y-, so the connection/net labels dropped below the part instead of leaving from its sides.

We already compute the correct facing from the schematic symbol (schematic_port.facing_direction, used elsewhere for matchpack and net-label insertion). We just weren't sending it to the solver, forcing it to re-derive (and mis-derive) it. This passes the ground truth instead of relying on the guess.

The roomy text-inclusive box is unchanged — it still gives parts breathing room, and the existing extendTraceEndpointsToReachPinsInsideExpandedBoundingBox still reconnects the routed trace back to the true port. Only the snap direction for these mis-detected pins changes.

Depends on

Requires the trace-solver change that snaps _facingDirection-carrying pins to
the edge along that direction instead of the nearest edge (and stops clearing it):
@tscircuit/schematic-trace-solver PR https://github.com/tscircuit/schematic-trace-solver/pull/645. Bump the dependency to that release. Without it, the solver's correctPinsInsideChips overwrites the facing we send here and nothing changes.

Screenshot 2026-07-08 at 4 07 45 PM


tests/components/primitive-components/__snapshots__/schematic-section-2-schematic.snap.svg
imrishabh18, line 1: this is getting worse

tests/repros/__snapshots__/repro84_schematic_pin_out-schematic.snap.svg
imrishabh18, line 1: Even for this the before trace was simpler

imrishabh18 (commented):
Overall looks good, but can you check why some of the simpler traces are choosing the longer trace

tests/repros/__snapshots__/repro84_schematic_pin_out-schematic.snap.svg
MustafaMulla29: Yes will definitely look into this.

tests/components/primitive-components/__snapshots__/schematic-section-2-schematic.snap.svg
MustafaMulla29: Yep will fix this.

MustafaMulla29:
There are more gains to this then losses, however will look onto the comments

lib/components/primitive-components/Group/Group_doInitialSchematicTraceRender/createSchematicTraceSolverInputProblem.ts
```diff
@@ -148,6 +149,15 @@ export function createSchematicTraceSolverInputProblem(
         pinId,
         x: schematicPort.center.x,
         y: schematicPort.center.y,
+        // Pass the port's true facing direction (known from the schematic
+        // symbol). The chip box handed to the solver is text-inclusive, so for
+        // small parts with a large reference designator the pins sit inside the
+        // box. The solver snaps such pins to the box edge along this facing
+        // direction (rather than guessing from geometry, which would pick the
+        // wrong edge for a resistor whose ref text widened the box).
+        _facingDirection: convertFacingDirectionToElbowDirection(
```
seveibar, line 158:
```suggestion
        facingDirection: convertFacingDirectionToElbowDirection(
```

seveibar: api issue

seveibar (commented):
@MustafaMulla29 this seems to have a lot of regressions- can you look into what's happening? Most of these are not improved no?

seveibar (approved):
i'm down to merge this but the regressions are _very very_ worrying, we need to create reproductions for the regressions (almost every schematic has a regression) then figure out what is going on

PR #2615, opened 2026-07-08
Fix: schematic pin labels which are not in the sequence

imrishabh18:
## BEFORE

image

## AFTER

image


lib/utils/filterPinLabels.ts
```diff
@@ -55,21 +53,9 @@ export function filterPinLabels(
...
-      pinLabels: { [pin]: label },
-    }
-
-    const result = chipProps.safeParse(testProps)
-    return result.success
-  } catch (error) {
-    return false
-  }
+function isValidPinLabel(_pin: string, label: string): boolean {
+  return /^[A-Za-z0-9_]+$/.test(label)
```
imrishabh18, line 60: is regex okay in this case?

seveibar (approved):

PR #2616, opened 2026-07-08
fix: Missing pin_number in the custom ports failed the ports to be unique

imrishabh18:


tests/repros/__snapshots__/repro129-host-custom-symbol-passives-schematic-schematic.snap.svg
imrishabh18, line 1: **BEFORE**

Both the R1 ports and C1 ports are connected to each other.

lib/components/primitive-components/Group/Group_doInitialSchematicTraceRender/getSchematicPortPinId.ts
```diff
@@ -0,0 +1,16 @@
+import type { SchematicPort, SourcePort } from "circuit-json"
+
+export const getSchematicPortPinId = ({
```
seveibar:
```suggestion
export const getSchematicPortSelector = ({
```

seveibar: not an ID!

seveibar (approved):

PR #2618, opened 2026-07-08
Update default autorouter pipeline to pipeline7

seveibar:


seveibar:
@copilot fix the types

PR #2624, opened 2026-07-09
update breakout test to use latest pipline 7

0hmX:


seveibar (approved):
cool, note that there is still some strange jaggedness that might mean we need a boundary fixing algorithm (i.e. a simplification that adjusts the inner and outer circuit traces near the boundary)

Issue #2620, opened 2026-07-09
board defaultTraceWidth (and nominalTraceWidth) have no effect on autorouted trace width


maci0:
Setting `defaultTraceWidth` on `<board>` does not change the width of autorouted traces. All segments come out 0.15mm. Same result with `nominalTraceWidth`.

```tsx
export default () => (
  <board width="20mm" height="20mm" defaultTraceWidth="0.3mm">
    <resistor name="R1" resistance="1k" footprint="0402" pcbX={-5} pcbY={0} />
    <resistor name="R2" resistance="1k" footprint="0402" pcbX={5} pcbY={0} />
    <trace from=".R1 > .pin2" to=".R2 > .pin1" />
  </board>
)
```

Exported KiCad PCB (`tsci export -f kicad_zip`):

```
(segment
  (start 95.51 100)
  (end 104.49 100)
  (width 0.15)
```

Expected `(width 0.3)`. The circuit json `pcb_trace` route points also all carry `"width": 0.15`, so the loss is in core, not the exporter.

Two causes (main at b1744d64):

1. `lib/components/primitive-components/Group/Group.ts:888-889` (`_runLocalAutorouting`) only reads `props.minTraceWidth` and `props.nominalTraceWidth`. `defaultTraceWidth` is accepted by the board/subcircuit props schema but never read anywhere in core, so it is silently dropped.
2. `lib/utils/autorouting/getSimpleRouteJsonFromCircuitJson.ts:567-570` fills each connection with `conn.nominalTraceWidth ??= minTraceWidth ?? board?.min_trace_width ?? 0.1`. The function's own `nominalTraceWidth` parameter is only stored as the top-level `SimpleRouteJson.nominalTraceWidth`, which the capacity autorouter ignores (its `TraceWidthSolver` consumes per-connection `nominalTraceWidth`). Since `Group.ts` always passes `minTraceWidth = 0.15`, every connection routes at 0.15mm no matter what the board requests.

Suggested fix: plumb `props.defaultTraceWidth ?? props.nominalTraceWidth` into the SRJ builder and use it for the per-connection `nominalTraceWidth`, keeping `minTraceWidth` as the lower bound only.

techmannih:
Hey @maci0 , thanks for reporting, we will take a look

PR #2625, opened 2026-07-09
add repro for board defaultTraceWidth being ignored by autorouted traces

techmannih:
ref #2620


seveibar (changes requested):
i don't understand this repro, you are defining a custom autorouter

seveibar (approved):

PR #2626, opened 2026-07-09
update auto router to 0.0.655

0hmX:


tests/breakout/__snapshots__/breakout-qfp16-with-header-and-passives-autorouting-srj.snap.svg
0hmX, line 1: The DRC has improved, so the jagged routing is better. It's a good repro. I'll debug in the next version.

AnasSarkiz (approved):

seveibar (approved):

PR #2631, opened 2026-07-10
Fix inherited default trace width handling in PCB routing

### LINKED ISSUE #2620: board defaultTraceWidth (and nominalTraceWidth) have no effect on autorouted trace width

techmannih:
fix #2620 

Fix PCB trace width resolution so inherited `defaultTraceWidth` is used for traces with no explicit thickness, instead of falling back too early to `minTraceWidth` / the manufacturing default.

## What changed

- add a shared helper to resolve inherited trace widths:
  - `minTraceWidth`
  - `nominalTraceWidth` / `defaultTraceWidth`
- update autorouter setup in `Group` to use the shared inherited width resolution
- update direct/manual trace rendering to use the same inherited fallback logic
- preserve the distinction between:
  - `minTraceWidth` as the DRC floor
  - `defaultTraceWidth` as the nominal routed width
- rename the local width variable in `Trace_doInitialPcbTraceRender` to make intent clearer


seveibar (changes requested):
too complex. you literally just need to add a backup `??` for nominalTraceWidth passed to SRJ

seveibar (changes requested):
mixing nominalTraceWidth with minTraceWidth

seveibar:
nominalTraceWidth should not be set to minTraceWidth

seveibar (changes requested):
ternary introduced makes code less clean and clear

PR #2633, opened 2026-07-10
Pass isCapacitor to matchpack

MustafaMulla29:


MustafaMulla29 (commented):
Almost every snapshot has regressions, will look into these!

seveibar (approved):

PR #2641, opened 2026-07-11
Fix SRJ obstacles for rotated and polygon SMT pads

ShiboSoftwareDev:
Preserve axis-aligned rotated pad orientation without applying rotation twice. Approximate non-rectangular polygon pads with scanline obstacles so curved pads retain their empty interior, while keeping rectangular polygons as single obstacles.
before:
image
after:
image

### DIFF
--- a/lib/utils/obstacles/getAxisAlignedRectFromPolygon.ts
+++ b/lib/utils/obstacles/getAxisAlignedRectFromPolygon.ts
@@ -1,0 +1,33 @@
+export const getAxisAlignedRectFromPolygon = (
+  points: Array<{ x: number; y: number }>,
+): {
+  center: { x: number; y: number }
+  width: number
+  height: number
+} | null => {
+  if (points.length !== 4) return null
+
+  const xs = [...new Set(points.map((point) => point.x))]
+  const ys = [...new Set(points.map((point) => point.y))]
+  if (xs.length !== 2 || ys.length !== 2) return null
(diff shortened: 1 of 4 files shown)


seveibar (approved):
cool

lib/utils/obstacles/getAxisAlignedRectFromPolygon.ts
```diff
@@ -0,0 +1,60 @@
+export const getAxisAlignedRectFromPolygon = (
+  points: Array<{ x: number; y: number }>,
+): {
+  center: { x: number; y: number }
+  width: number
+  height: number
+} | null => {
+  const tolerance = 1e-3
```
seveibar: is this determining the size of the boxes?

seveibar: for consistency we should use the same parameters that are used to approximate traces as boxes etc. if possible, more boxes make worse topologies, so there is a delicate balance- i just don't see the tuning parameter

ShiboSoftwareDev: this is for determining if the polygon is just a rect and we don't need to do approximation for it at all, distinct from the scanline approximation sizing logic, renamed it so the distinction is a bit more clear maybe

AnasSarkiz (approved):

PR #2646, opened 2026-07-12
bump: update circuit-to-svg to v0.0.381

anil08607:


tests/components/normal-components/__snapshots__/capacitor-polarized-schematic.snap.svg
techmannih, line 1: Can you please why the net label text overlapping?

techmannih (approved):

PR #2654, opened 2026-07-13
Update capacity-autorouter to 0.0.665

seveibar:
## Summary

- update `@tscircuit/capacity-autorouter` from `^0.0.655` to `^0.0.665`
- regenerate 35 PCB snapshots, four autorouting diagnostic snapshots, and two route-dependent 3D snapshots for the new solver output
- keep the existing peer dependency range unchanged

## Why

This brings `@tscircuit/core` onto the latest published capacity autorouter release and records the resulting local-routing output changes.

## Validation

- `bun install` (resolved `@tscircuit/capacity-autorouter@0.0.665`)
- `bunx tsc --noEmit`
- `bun run build`
- targeted rerun of regenerated routing and 3D snapshots
- all four GitHub Actions Bun Test shards
- GitHub Actions typecheck, smoke test, dependency check, and formatting checks

## Test fixture adjustments

- increase the QFP16 breakout padding from 1 mm to 2 mm so the updated solver has enough fanout space and produces no PCB trace DRC errors
- disable irrelevant PCB rendering in the schematic-only `repro131` test, avoiding an autorouter timeout without changing its schematic coverage


seveibar:
@ShiboSoftwareDev this brings your fixes live iiuc

PR #2656, opened 2026-07-13
fix schematic box explicit port connections

seveibar:
## Summary

- inherit `connections` targets for explicit ports on schematic-box groups
- keep group PCB ports aligned after layout and represent coincident port connections without autorouting
- include boxed child subcircuits in parent schematic routing while avoiding internal-only fallback labels
- add PCB and schematic regression snapshots for the reported RC filter

## Root cause

When an explicit `<port>` matched a `connections` key, core skipped creating the synthesized port but did not carry the connection target onto the explicit port. The group port therefore had no PCB position. Explicit child order and PCB packing could also leave a resolved group port stale. Separately, the parent schematic solver skipped subcircuit boundaries, so it never saw the boxed child port.

## Impact

`showAsSchematicBox` subcircuits can now use explicit directed ports with `connections`: internal coincident connections satisfy PCB validation, external traces route, and the boxed component connects correctly on the schematic.

## Validation

- `bun test` (1,155 passed, 31 skipped)
- `bunx tsc --noEmit`
- `bun run build`
- visual PCB and schematic snapshot review


seveibar:
Superseded by the reproduction-only PR #2662 and the adversarially reviewed replacement implementation in #2664.

PR #2659, opened 2026-07-13
Fix schematic auto-layout text overlap for rotated components with long names

MustafaMulla29:
In schematic auto-layout, a component whose name/value label is long enough (e.g. a voltage source named VBUS20_ECOK_IN_OUT, or a resistor RESIS1_ECOK_IN_OUT) overlaps the component next to it when auto-layout rotates it — which happens for any 2-pin part with a power/ground pin (the part gets stood upright so its ground pin faces down). Short-named parts (R2, C1) were unaffected.

Root cause

The text-measurement itself was never wrong — getSchematicComponentWithTextBounds correctly covers a part's label in any orientation (the bounding-box tests prove this at 0/90/180/270°). The bug was in how auto-layout used it:

@tscircuit/matchpack reserves each part's footprint while the part is still in its starting orientation, then rotates the whole box to the part's final position. But a symbol draws its label on a different side depending on orientation (a resistor's name sits above it when flat, but to the side when upright). So rotating the "name-above" box puts the reserved space on the wrong axis:

RESIS1_ECOK_IN_OUT measured flat : 2.16 wide × 0.68 tall   (name reserved horizontally, above)
matchpack rotates that box 90°   : 0.68 wide × 2.16 tall   (reservation now vertical)
but the upright name is horizontal, to the side → needs 2.16 wide → overlaps the neighbor

The error between "rotated starting box" and "correct final box" scales with label length. For short names it's smaller than the gap matchpack keeps between parts, so it was harmless; for long names it far exceeds the gap, so the name runs into the next part.

Fix

Make auto-layout measure the label footprint in the part's final (post-rotation) orientation — the same measurement the bounding-box tests use, but with the rotation applied — so the reserved space matches where the label will actually be drawn.

- getSchematicComponentWithTextBounds now accepts a ccwRotationDegrees and measures against the rotated symbol variant (which is where the symbol set actually places the label).
- Auto-layout determines each part's locked rotation up front (from its ports' nets) and sizes each chip once in its final orientation — a single pass, no re-measure.
- The rotated reservation is only applied when it grows past matchpack's inter-chip gap (chipGap). Anything smaller is provably absorbed by that gap and can't overlap a neighbor, so short-labeled parts keep their existing layout and unrelated components don't shift.

Why it's needed

Without it, any auto-laid-out board with power/ground-connected parts that have descriptive names produces schematics where the names collide with adjacent components — unreadable and misleading. The fix is scoped so it only affects parts whose label genuinely sticks out far enough to overlap; everything else is byte-identical.

Before: 
Screenshot 2026-07-13 at 5 22 21 PM

After:
Screenshot 2026-07-13 at 5 22 53 PM

REF: 
Screenshot 2026-07-13 at 5 29 43 PM


seveibar (changes requested):
Wrong repo for implementation or just wrong implementation- youre obfuscating the chip width/height.

seveibar (approved):
Sorry was confused

seveibar (changes requested):
Why are you getting the rotated symbol name-

seveibar (approved):
Your symbol rotation logic is a hack- i dont see another way atm

MustafaMulla29:
> Why are you getting the rotated symbol name-

Coz matchpack doesnt support it 

> reserves each part's footprint while the part is still in its starting orientation, then rotates the whole box to the part's final position. But a symbol draws its label on a different side depending on orientation (a resistor's name sits above it when flat, but to the side when upright). So rotating the "name-above" box puts the reserved space on the wrong axis

"Matchpack reserves each part's footprint while the part is still in its starting orientation, then rotates the whole box to the part's final position. But a symbol draws its label on a different side depending on orientation (a resistor's name sits above it when flat, but to the side when upright). So rotating the "name-above" box puts the reserved space on the wrong axis: 
RESIS1_ECOK_IN_OUT measured flat : 2.16 wide × 0.68 tall (name reserved horizontally, above)
matchpack rotates that box 90° : 0.68 wide × 2.16 tall (reservation now vertical)
but the upright name is horizontal, to the side → needs 2.16 wide → overlaps the neighbor"

PR #2675, opened 2026-07-14
fix: hide collapsed group internal connectivity labels

ShiboSoftwareDev:
## Summary

- skip raw connectivity-key fallback labels for direct ports on collapsed schematic groups when every connected source endpoint is inside that group
- preserve labels selected from named source nets, assigned display labels, cross-subcircuit connections, and visible component-port labels
- add a positive regression showing an external trace to `net.ROW1` still renders the named `ROW1` label
- update the internal-only snapshot so `ROW1` remains an open box pin

## Root cause

`insertNetLabelsForPortsMissingTrace` treats any port with a connectivity key but no routed schematic trace as needing a label. For a direct port on a `showAsSchematicBox` group, `connectsTo` creates real source connectivity to hidden child ports, but the connectivity island has no user-facing name. The label-selection chain therefore falls through to the generated `connKey`, leaking the implementation identifier outside the collapsed box.

The fix identifies the actual owning collapsed group through the runtime port tree and verifies that every source trace endpoint connected to the public port is contained inside that group before suppressing only the final connectivity-key fallback.

## Impact

Collapsed schematic boxes keep internal unnamed connectivity encapsulated. PCB routing and source connectivity are unchanged. Explicit external named nets and externally visible connections continue to render normally.

## Stack

- Base reproduction: #2674
- This PR contains the fix and targets the reproduction branch.

## Validation

- `bunx biome check ...`
- `bunx tsc --noEmit`
- targeted group-port and fallback-net-label tests
- full suite: **1168 passed, 31 skipped, 0 failed**


MustafaMulla29 (commented):
Not sure how will the user know the subcircuit is unnamed of we remove these cc @seveibar

ShiboSoftwareDev:
this is just the internal net being reflected outside, since it doesn't exist outside an unnamed netlabel is placed, now we don't show that netlabel since the outside port isn't connected to anything external

MustafaMulla29:
> this is just the internal net being reflected outside, since it doesn't exist outside an unnamed netlabel is placed, now we don't show that netlabel since the outside port isn't connected to anything external

Ok. Then it should also show the port dot right? Or am I wrong

MustafaMulla29 (approved):

ShiboSoftwareDev:
true I didn't notice that, it's likely an existing issue, but needs fixing nonetheless

ShiboSoftwareDev:
The missing open-terminal dot in this PR's collapsed-box snapshot is a separate `circuit-to-svg` issue: the renderer infers pin connectivity from any `source_trace`, including implementation-only traces hidden by the collapsed group, instead of honoring the public port's schematic connection state.

I opened a stacked renderer pair:

- Repro: tscircuit/circuit-to-svg#620
- Fix: tscircuit/circuit-to-svg#621

Once #621 is released, this PR can bump `circuit-to-svg` and refresh only the targeted snapshot so `ROW1` renders the standard open terminal dot.

seveibar (approved):

PR #2681, opened 2026-07-14
Fix subcircuit autorouting to ignore descendant routing intent

Abse2001:


seveibar:
CC @0hmX

lib/utils/autorouting/getSimpleRouteJsonFromCircuitJson.ts
```diff
@@ -435,12 +435,76 @@ export const getSimpleRouteJsonFromCircuitJson = ({
...
+      return candidate._exposesSubcircuitConnection && candidate.source_trace_id
+        ? [candidate.source_trace_id]
+        : []
+    }),
+  )
+  const exposedDescendantSourceNetIds = new Set<string>()
+  for (const trace of currentSubcircuitSourceTraces) {
+    if (
+      !exposedBridgeSourceTraceIds.has(trace.source_trace_id) &&
+      !trace.name?.startsWith("exposed_net.")
```
0hmX: string matching is not good i think

Abse2001: Ok looking at this, how should the fix be in your opinion?

seveibar: yes it should be a property if anything on the trace etc.

seveibar: nice catch @0hmX

0hmX (approved):

seveibar (approved):

PR #2696, opened 2026-07-15
Skip schematic phases for PCB-only Circuit JSON imports

rushabhcodes:
## Summary

Makes Circuit JSON inflation representation-aware so PCB-only imports do not synthesize and route an unrelated schematic.

- adds centralized render-phase gating to `Renderable`
- detects the PCB and schematic representations actually present in imported Circuit JSON
- skips schematic phases only within PCB-only imported subcircuits
- preserves normal schematic rendering for source-only imports, mixed/native designs, and imports containing schematic elements
- ignores error and warning records when detecting render domains
- keeps imported PCB traces and vias intact with or without board-level `schematicDisabled`

The PR also upgrades the KiCad import dependencies and refreshes the Arduino Uno snapshots for the current converter and router output:

- `kicad-to-circuit-json`: `^0.0.97` → `^0.0.113`
- `kicadts`: `^0.0.47` → `^0.0.51`

## Root cause

The KiCad converter emits a complete PCB representation and source connectivity, but no schematic representation. During inflation, core recreated normal components and traces from that source data and then treated them as a request to synthesize a new schematic.

For the Arduino Uno import, the synchronous schematic trace solver attempted to route that large synthesized schematic before PCB autorouting began. Bun's timeout could not promptly interrupt the synchronous work, so the repro appeared stuck.

Globally setting `schematicDisabled` avoided the symptom but incorrectly disabled schematics for the entire design and required callers to understand an inflation implementation detail.

## Implementation

`Renderable` now exposes a centralized phase-applicability hook that applies to both a component and its subtree. Skipped phases are marked complete so render state remains consistent.

Inflated subcircuits record whether their Circuit JSON contains renderable PCB or schematic elements. A subcircuit with renderable PCB elements and no renderable schematic elements skips its schematic phase block. Source and PCB phases continue normally.

Diagnostic records are excluded from domain detection through their canonical `error_type` and `warning_type` discriminants. Therefore:

- a `pcb_missing_footprint_error` cannot suppress schematic generation for a source-only import
- a stale `schematic_error` cannot enable schematic regeneration for a PCB-only import

The render-domain classifier is extracted into a small helper with direct unit coverage for PCB errors, schematic errors and warnings, and real PCB/schematic elements.

## Developer impact

Consumers can import PCB-only KiCad output without disabling schematics globally. Mixed designs can render native schematics alongside PCB-only imported subcircuits, while imported PCB geometry and rerouting behavior remain unchanged.

## Validation

- `bunx tsc --noEmit`
- `bun test tests/subcircuits/get-circuit-json-render-domains.test.ts` — 3 assertions
- `bun test tests/repros/repro116-arduino-uno.test.tsx tests/repros/repro117-schematic-disabled-inflation.test.tsx` — 122 assertions
- `bun test tests/repros/repro116-arduino-uno-reroute-center.test.tsx` — 124 assertions


seveibar (changes requested):
not the right pattern

seveibar:
there is an existing pattern for skipping phases, Renderable.ts is a very very important file that absolutely cannot contain this kind of hack

seveibar:
at the beginning of the render function you detect if schematic is disabled, very straightforward pattern! no hacks needed!!! scares me that this was PR'd!

PR #2704, opened 2026-07-16
fix148 schematicsheet auto-layout overlap

mohan-bee:
repro here: https://github.com/tscircuit/core/pull/2703 

**Before**: Auto-layout inside a schematic sheet placed components at (0, 0), while generated traces and net labels were missing because they did not inherit the sheet ID.

**After**: Components, traces, and net labels now determine their sheet from the connected components or ports, so automatic schematics render correctly inside the sheet.


seveibar:
Cc @MustafaMulla29

MustafaMulla29 (approved):

PR #2706, opened 2026-07-16
Fix bridged solderjumper pads falsely flagged as trace-clearance violations by resolving port ids before placement DRC

rushabhcodes:
## Summary

Bridged solderjumpers (e.g. `<solderjumper footprint="solderjumper2_bridged12" bridgedPins={[["1","2"]]} />`) were incorrectly producing `pcb_pad_trace_clearance_error`s on their own pads, and this could cascade into skipped/blocked autorouting for traces connected through the jumper.

## Root cause

A bridged solderjumper footprint includes a physical bridge trace between its shorted pins, generated by the footprint library. Since real `pcb_port_id`s don't exist yet at footprint-generation time, that trace is emitted with placeholder endpoint ids: `"{PIN1}"`, `"{PIN2}"`.

`SolderJumper` and `Jumper` resolve these placeholders into real port ids inside `doInitialPcbTraceRender()`, which runs during the `PcbTraceRender` render phase (see `orderedRenderPhases` in `lib/components/base-components/Renderable.ts`). The issue is phase ordering: `PcbPlacementDesignRuleChecks` (pad/trace clearance checking) runs **one phase earlier** than `PcbTraceRender`.

So when the clearance checker builds its connectivity map, the bridge trace's endpoints are still the literal strings `"{PIN1}"`/`"{PIN2}"` — they don't resolve to any pad. The checker concludes the trace is unconnected to the two pads it's physically touching (by design, at 0 gap) and raises a clearance violation for each one. Since the codebase treats the diagnostic as authored by the DRC phase rather than a routing decision, this is unrelated to the autorouter itself — the connectivity is fine once the ids resolve, the check just runs too early.

## Fix

Renamed the placeholder-resolution method in both `SolderJumper` and `Jumper` from `doInitialPcbTraceRender()` to `doInitialPcbPortAttachment()`. This moves the same logic to the `PcbPortAttachment` phase, which runs immediately after ports are created (`PcbPortRender`) and well before `PcbPlacementDesignRuleChecks`. By the time clearance checks run, the bridge trace already references real port ids, so the connectivity map correctly treats the bridged pads as connected.

No behavior changes for non-bridged jumpers/solderjumpers — the fixup logic itself is untouched, only its render phase.

## Reproduction

Added `tests/repros/repro123-bridged-solderjumper-skips-autorouting.test.tsx`: two resistors wired through a bridged 2-pin solderjumper. Before this fix it produced 2 `pcb_pad_trace_clearance_error`s; after the fix it produces 0, and autorouting between the resistors succeeds.

## Testing

- [x] New repro test passes with 0 `pcb_pad_trace_clearance_error` / `pcb_autorouting_error`, snapshot updated
- [x] Existing solderjumper/jumper test suite (`solderjumper*.test.tsx`, `jumper*.test.tsx`, `pinrows*.test.tsx`, `auto-jumper-autorouting.test.tsx`, jumper-related repros) passes
- [x] Full test suite run (`bun test`): 1173 pass / 9 fail / 31 skip — the 9 failures are pre-existing timeouts/autorouter issues (differential-pair routing, pcb-pack layout) reproducible on unmodified `main` and unrelated to this change


mohan-bee (approved):
lgtm.

ShiboSoftwareDev (approved):
nice

MustafaMulla29 (approved):

PR #2712, opened 2026-07-16
Remove obsolete core PCB DRC deduplication

ShiboSoftwareDev:
## Summary

- stop applying `dedupePcbDrcErrors` when `Board` aggregates DRC results
- insert all routing, placement, netlist, pin-specification, and custom DRC results directly
- update the integration test to verify distinct custom DRC records are preserved

## Why

`@tscircuit/checks@0.0.145` now classifies a trace/obstacle pair as either a physical overlap or a clearance violation. Core no longer needs to remove one result after aggregation, and doing so can incorrectly discard intentionally distinct results from custom checks.

The checks dependency update to `0.0.145` is already present on `main` from #2709; this PR completes the core integration.

## Validation

- `bun test tests/drc/preserve-pcb-drc-errors.test.tsx`
- `bunx tsc --noEmit`
- `bun run build`
- `bunx biome check lib/components/normal-components/Board.ts tests/drc/preserve-pcb-drc-errors.test.tsx`
- `bun test`: 1,183 pass, 31 skip, with one existing `differential-pair/port-selector` snapshot mismatch; the same mismatch reproduces when the old dedupe call is restored and contains routing geometry rather than DRC output

### DIFF
--- a/lib/components/normal-components/Board.ts
+++ b/lib/components/normal-components/Board.ts
@@ -1,5 +1,4 @@
 import {
-  dedupePcbDrcErrors,
   runAllNetlistChecks,
   runAllPinSpecificationChecks,
   runAllPlacementChecks,
@@ -691,7 +690,7 @@ export class Board
       }
 
       const checkResults = await Promise.all(checksToRun)
-      db.insertAll(dedupePcbDrcErrors(checkResults.flat()))
+      db.insertAll(checkResults.flat())
     }
 
     const subcircuit = db.subtree({ subcircuit_id: this.subcircuit_id })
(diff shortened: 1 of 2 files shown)


MustafaMulla29:
https://github.com/tscircuit/core/pull/2711/files
this should have failed the breakout tests right? if "dedupePcbDrcErrors" was the issue? maybe the issue is in checks repo?

ShiboSoftwareDev:
> https://github.com/tscircuit/core/pull/2711/files
> this should have failed the breakout tests right? if "dedupePcbDrcErrors" was the issue? maybe the issue is in checks repo?

Not sure what you mean, the checks change I did separates overlap errors and clearance errors, dedupe was a patch for a bug where an issue can count as both errors

MustafaMulla29:
https://github.com/tscircuit/core/pull/2711/files
you can see here that the error counts are increasing in the breakout tests.

PR #2714, opened 2026-07-16
Fix subcircuit autorouting to ignore descendant routing intent

Abse2001:


seveibar:
hard to review until tests pass! hoping to merge soon tho, i think it's ok

seveibar (approved):
I think this might expose every descendent net, not just exposed ones IIUC, we may need a test to confirm

PR #2717, opened 2026-07-17
Fix PCB autorouting around fixed footprint copper traces

rushabhcodes:
## Summary

Fixes a PCB autorouting bug where routes could cross fixed copper traces embedded in footprints, causing accidental electrical contact. This was reproducible with bridged solder jumpers, whose copper bridge was omitted from the autorouter obstacle model.

## Root cause

The Simple Route JSON obstacle input included pads, holes, vias, and keepouts, but omitted source-less `pcb_trace` elements used for fixed footprint copper. Because the bridge was invisible to the autorouter, a new route could pass directly through it and later produce an accidental-contact error.

## Changes

- Include source-less, footprint-owned PCB traces as fixed autorouting obstacles.
- Use `pcb_trace_id` as the obstacle connectivity identifier when a fixed trace has no `source_trace_id`.
- Update repro149 to assert that routing completes without PCB trace or autorouting errors.
- Update the PCB snapshot to verify that the autorouter avoids the solder-jumper bridge.

## Impact

This improves core autorouter obstacle modeling for footprint-defined copper. Routes now avoid solder-jumper bridges and other fixed copper primitives instead of creating accidental shorts.

## Validation

- `bun test tests/repros/repro149-autorouter-crosses-bridged-solderjumper.test.tsx`
- `bun test tests/repros/repro123-bridged-solderjumper-skips-autorouting.test.tsx tests/repros/repro122-solderjumper-bridged-props-pcb-footprint.test.tsx tests/components/normal-components/solderjumper.test.tsx`
- `bunx biome check lib/utils/autorouting/getSimpleRouteJsonFromCircuitJson.ts lib/utils/obstacles/getObstaclesFromCircuitJson.ts tests/repros/repro149-autorouter-crosses-bridged-solderjumper.test.tsx`

The full TypeScript check currently reports unrelated existing errors in ten-layer tests and tests using the `internalCircuit` chip prop.


seveibar (approved):
Cc @0hmX

0hmX (approved):

PR #2718, opened 2026-07-17
Support internal circuits in chip packages

MustafaMulla29:
## Summary

- add the `<internalcircuit>` primitive for use through `<chip internalCircuit={...}>`
- render internal components as package-scoped source and schematic components such as `Q1A` and `Q1B`
- resolve `pin.<label>` connections against the owning chip's package ports and scope internal component selectors to their package
- keep the owning chip as the only physical PCB/CAD/parts-engine component
- suppress the owning chip's outer schematic symbol and ports when its internal schematic representation is present
- replace the temporary repro-only JSX declaration with the merged `@tscircuit/props` intrinsic type

## Why

Real packages such as dual MOSFETs, resistor networks, and diode arrays contain multiple logical schematic elements but must produce one PCB component, footprint, and manufacturing part. This implements the behavior covered by the repros added in #2716.

For example, a dual MOSFET package produces `Q1A` and `Q1B` on the schematic while producing only `Q1` on the PCB.

## Validation

- `bun test` across the four internal-circuit cases and neighboring chip/render-subtree coverage: 12 passed, 0 failed
- `bunx tsc --noEmit`
- `bun run build` (ESM and DTS)
- `git diff --check`


lib/components/base-components/NormalComponent/NormalComponent.ts
```diff
@@ -160,6 +160,27 @@ export class NormalComponent<
...
   _impliedFootprintPinLabels?: Record<string, string | string[]>
 
+  override get name() {
+    const localName = super.name
+    const internalCircuit = this.getInternalCircuitAncestor()
+    const packageName = internalCircuit?.parent?.name
+
+    if (!localName || !packageName) return localName
+    return `${packageName}${localName}`
+  }
```
seveibar, line 170: Remove, the pattern is to use resolveName and anyway we never use inferred names (why is this here?)

seveibar (changes requested):
Repros arent good enough- i dont understand whats going on

seveibar (commented):
Make sure the visual snapshots are fully explanative, this is a critical feature that could have MASSIVE implications for tech debt if incorrectly implemented- the repros should be extremely clear before there is an attempt at implementation- none of the repros even resemble a user’s circuit

lib/components/base-components/NormalComponent/NormalComponent.ts
```diff
@@ -160,6 +160,27 @@ export class NormalComponent<
   _invalidPinLabelMessages: string[] = []
   _impliedFootprintPinLabels?: Record<string, string | string[]>
 
+  override get name() {
+    const localName = super.name
+    const internalCircuit = this.getInternalCircuitAncestor()
+    const packageName = internalCircuit?.parent?.name
```
```suggestion
    const parentName = internalCircuit?.parent?.name
```

lib/components/base-components/NormalComponent/NormalComponent.ts
seveibar: Again, i disagree with this pattern. The name is the name-

lib/components/primitive-components/InternalCircuit.ts
```diff
@@ -0,0 +1,36 @@
...
+    if (!(parent instanceof Chip)) {
+      throw new Error(
+        "<internalcircuit> must be provided through a <chip internalCircuit={...}> prop",
+      )
+    }
+    super.onAddToParent(parent)
+  }
+
+  override runRenderPhaseForChildren(phase: RenderPhase): void {
+    if (phase.startsWith("Pcb")) {
```
MustafaMulla29, line 28: same pattern is used in Ammeter.ts, CurrentSource.ts etc. I think we need a better way to just run schematic phases?

lib/components/base-components/NormalComponent/NormalComponent.ts
seveibar: Remove

lib/components/base-components/PrimitiveComponent/PrimitiveComponent.ts
```diff
@@ -142,10 +142,14 @@ export abstract class PrimitiveComponent<
     return this.lowercaseComponentName === "group"
   }
 
-  get name() {
+  resolveName() {
     return (this._parsedProps as any).name ?? this.fallbackUnassignedName
   }
 
+  get name() {
+    return this.resolveName()
```
seveibar, line 150: Remove

lib/components/primitive-components/InternalCircuit.ts
```diff
@@ -0,0 +1,36 @@
...
+  override onAddToParent(parent: PrimitiveComponent): void {
+    if (!(parent instanceof Chip)) {
+      throw new Error(
+        "<internalcircuit> must be provided through a <chip internalCircuit={...}> prop",
+      )
+    }
+    super.onAddToParent(parent)
+  }
+
+  override runRenderPhaseForChildren(phase: RenderPhase): void {
```
seveibar, line 27: Remove

PR #2732, opened 2026-07-20
Track physical trace connectivity for preserved routed subcircuits

Abse2001:
The child subcircuit routing itself was correct. The bug happened when Core prepared the parent-board SRJ.

Expected flow:

Route the Pico subcircuit.
Preserve its copper as static traces.
Mark which exact PCB ports each preserved trace physically connects.
Route only the remaining board connections.

Actual flow:

Core preserved the child copper geometry, so the autorouter saw it as static obstacles.
But Core dropped the autorouter’s connectsTo: [portA, portB] property.
It replaced it with broad connectedTo electrical-net metadata, which the autorouter does not use for physical route state.
Therefore, the board router saw the child copper but did not know that its endpoints were already connected.
It generated the Pico’s internal MST connections again, tried routing over the static Pico traces, and eventually ran out of iterations.

The fix:

Core now preserves the native connectsTo property with only the exact endpoints connected by each physical trace.
It no longer expands a child trace into every identifier on the electrical net.
This uses explicit properties—not trace-name or exposed_net.* string matching.
In the autorouter, overlapping physical pairs are now merged transitively. For example, [A,B] and [B,C] correctly mean that A, B, and C are already one connected component.

Verification:

Generated board SRJ: 51 remaining connections and 136 child traces.
All 136 child traces had physical connectsTo pairs.
The full SRJ routed successfully in the autorouter in about 84 seconds.
A clean rerun passed and matched the generated full-board SVG snapshot.
Core also completed the Gameboy routing using autorouter 0.0.696 with zero autorouting errors.
So this was a routing-state contract bug between Core and the autorouter—not a need for more iterations or a Gameboy-layout workaround.


seveibar (approved):
Nice

PR #2734, opened 2026-07-20
Add schematic-only internal MOSFET circuits

seveibar:
## What changed

- add `<internalcircuit>` as a semantic chip child without changing the rendering behavior of its children
- instantiate `<mosfet>` children as `InternalCircuitMosfet` while preserving normal MOSFET behavior elsewhere
- render internal MOSFETs and traces in the schematic only while keeping the owning chip as the single physical PCB package
- scope internal names and trace selectors to the owning chip so multi-unit devices can use names such as `Q1A` and `Q1B`
- enable the dual-MOSFET regression and update its schematic snapshot; the existing PCB snapshot remains unchanged

## Why

Chip internals need a schematic-only representation without duplicating physical components or making `<internalcircuit>` override how arbitrary children render. Selecting a MOSFET-specific internal class at React instance creation keeps that behavior explicit and contained.

## Impact

Users can provide a chip's `internalCircuit` prop to expose internal MOSFET units in the schematic. Ordinary MOSFETs and chips without an internal circuit retain their existing schematic and PCB behavior.

## Checks

- `bun test tests/repros/repro-internal-circuit-dual-mosfet.test.tsx tests/components/normal-components/mosfet.test.tsx tests/components/normal-components/chip-no-schematic-representation.test.tsx`
- `bunx tsc --noEmit`
- `bun run build`
- `git diff --check`


lib/components/base-components/NormalComponent/NormalComponent.ts
```diff
@@ -224,12 +224,13 @@ export class NormalComponent<
 
   doInitialSourceNameDuplicateComponentRemoval(): void {
     // Early return if component has no explicit name (auto-assigned names don't conflict)
-    if (!this.name) return
+    const nameForDuplicateCheck = this.getNameForDuplicateCheck()
```
seveibar, line 227: remove

PR #2737, opened 2026-07-21
Connect external traces to internal circuit ports

MustafaMulla29:
## Summary

- render a chip's schematic port at the connected internal-circuit symbol terminal while preserving separate chip and internal `schematic_port` identities
- treat coincident schematic ports as one geometry endpoint for layout and trace routing without changing names or render-phase ordering
- add focused feature tests for an external resistor connection and chip-pin fanout, including PCB snapshots that keep internal MOSFETs schematic-only

## Why

When `schShowInternalCircuit` replaces the chip body with its internal symbols, the chip port still needs schematic geometry so external traces such as `R1.pin2 -> U1.pin1` can render. The source trace existed, but the package port had no visible schematic endpoint.

Giving the chip port the connected internal terminal's geometry solves that while retaining both database ports. Geometry solvers now normalize coincident ports so the additional identity does not create zero-length routes or perturb schematic auto-layout.

## Impact

External components can connect through chip pins to displayed internal MOSFET terminals. Internal-circuit components and their internal traces remain schematic-only and do not gain PCB representations.

## Validation

- focused external-connection and fanout feature tests
- existing dual-MOSF internal-circuit test
- 12 surrounding matchpack and schematic-layout regression tests
- `bunx tsc --noEmit`
- `bun run build`

Before: 
Screenshot 2026-07-21 at 12 56 42 PM

After: 
Screenshot 2026-07-21 at 12 57 14 PM


seveibar (changes requested):
The schematic_port should be the same whether or not there is an internal circuit no? Looks like youre creating a dangerous new paradigm with a lot of corcuit json complexity by introducing anchors

seveibar:
The schematic port for “pin1” is essentially present whether or not there’s an internal circuit

lib/components/normal-components/Chip.ts
```diff
@@ -79,6 +79,46 @@ export class Chip<PinLabels extends string = never> extends NormalComponent<
...
+      )
+
+      // A package pin has one schematic location. Additional internal ports
+      // connected to it retain their own ports and fan out with normal traces.
+      chipPort.schematicPortGeometryAnchor =
+        internalCircuitPorts.find(
+          (internalPort) =>
+            internalPort.source_port_id &&
+            directlyConnectedSourcePortIds.has(internalPort.source_port_id),
+        ) ?? null
```
seveibar, line 120: It is a ports reaponsibility, not a chips, to decide where to be located. The only way this codebase scales is if its organized well

lib/components/primitive-components/Group/Group_doInitialSchematicTraceRender/Group_doInitialSchematicTraceRender.ts
```diff
@@ -27,8 +27,15 @@ export const Group_doInitialSchematicTraceRender = (group: Group<any>) => {
...
     userNetIdToConnKey,
     connKeysWithExplicitPortNetTraces,
+    schematicPortIdsConnectedAtSamePosition,
   } = createSchematicTraceSolverInputProblem(group)
 
+  for (const schematicPortId of schematicPortIdsConnectedAtSamePosition) {
+    group.root.db.schematic_port.update(schematicPortId, {
+      is_connected: true,
+    })
+  }
```
seveibar, line 37: Something like this is the worst possible tech debt you can introduce. I spoke about this on monday

DO NOT DO GEOMETRIC MATCHING FOR DATA MODELING ISSUES


DISTANCE CHECKS ARE GEOMETRIC MATCHING!!!!! IT IS AN EXPENSIVE, DANGEROUS AND DIFFICULT TO EVER FIX HACK!

lib/components/primitive-components/Port/Port.ts
```diff
@@ -25,6 +25,11 @@ export class Port extends PrimitiveComponent<typeof portProps> {
   schematic_stem_line_id: string | null = null
 
   schematicSymbolPortDef: SchSymbol["ports"][number] | null = null
+  /**
+   * A different port whose symbol terminal supplies this port's schematic
+   * geometry. This port retains its own source and schematic port identity.
+   */
+  schematicPortGeometryAnchor: Port | null = null
```
seveibar, line 32:
```suggestion
  internalCircuitPort: boolean = false
  exposedChipPort: Port | null = null
```

PR #2738, opened 2026-07-21
Hide internal connectivity labels on boxed subcircuit ports

MustafaMulla29:
## Summary

- hide implementation-only connectivity labels on dangling boxed subcircuit ports
- prevent internal `*_connectivity_net*` keys from becoming user-facing schematic text
- use Circuit JSON subcircuit membership when the parent schematic pass cannot access child `Port` instances
- update the merged repro snapshot

## Root cause

The existing fallback suppression depended on `group.selectAll("port")`. During the parent schematic routing pass, ports inside a child subcircuit are outside that selector scope, so the box port could not be recognized as internally connected and its connectivity-map key escaped as a schematic net label.

The fallback now also recognizes box pins from Circuit JSON and verifies whether their connections remain inside the collapsed subcircuit. Internal-only ports remain visually open. Explicit external named-net labels retain their existing behavior.

## User impact

Dangling boxed ports such as `STATUS` and `ENABLE` no longer show generated labels such as `power_connectivity_net0`. They render as open box pins until a parent-visible connection is added.

The separate missing GND routing issue remains outside this PR.

## Validation

- `bun test tests/features/group-ports tests/repros/repro154-subcircuit-exposed-port-schematic-connections.test.tsx` — 7 passed
- focused trace-label fallback tests — 4 passed
- repro snapshot contains no net-label elements for the dangling `STATUS` and `ENABLE` ports
- `bun run build` — ESM build succeeds; DTS generation remains blocked by the existing unrelated `schShowInternalCircuit` type error in `lib/components/normal-components/Chip.ts`


rushabhcodes (approved):

seveibar (commented):
The test net labels have changed- the pins are differrent?

Can you make a more clear repro? What ports are supposed to be labeled?

MustafaMulla29:
> The test net labels have changed- the pins are differrent?
> 
> Can you make a more clear repro? What ports are supposed to be labeled?

The labels were wrong with power_connectivity_net0 etc we should shown connectivity labels right? And it was not adding a fallback netlabel for missing trace. Am I doing something wrong? We should show fallback netlabels for trace path as well right? @seveibar

seveibar (approved):

seveibar:
Test needs explanatory text

PR #2739, opened 2026-07-21
fix: defaultTraceWidth/nominalTraceWidth board props now control autorouted trace width

### LINKED ISSUE #2620: board defaultTraceWidth (and nominalTraceWidth) have no effect on autorouted trace width

Hero988:
Fixes #2620

## What & why

`<board defaultTraceWidth="0.3mm">` (and `nominalTraceWidth`) had no effect on autorouted trace width — every autorouted segment came out at 0.15mm, exactly as diagnosed in the issue:

1. `Group._runLocalAutorouting` only read `props.minTraceWidth` / `props.nominalTraceWidth`; `defaultTraceWidth` was accepted by the props schema but never read anywhere in core.
2. `getSimpleRouteJsonFromCircuitJson` filled each connection's `nominalTraceWidth`/`width` from `minTraceWidth ?? board.min_trace_width ?? 0.1` and only stored its `nominalTraceWidth` parameter as the top-level `SimpleRouteJson.nominalTraceWidth`, which the capacity autorouter's `TraceWidthSolver` ignores (it consumes per-connection `nominalTraceWidth`).

## Changes

- **`Group._runLocalAutorouting`**: the nominal width passed to the SRJ builder is now `props.defaultTraceWidth ?? props.nominalTraceWidth ?? minTraceWidth`.
- **`getSimpleRouteJsonFromCircuitJson`**: connections without an explicit width (e.g. from a `<trace thickness>`) now default to the `nominalTraceWidth` parameter when provided; `minTraceWidth` remains the routing lower bound only (`SimpleRouteJson.minTraceWidth` is still computed from it).
- **Legacy remote solve-endpoint path**: passes the same `defaultTraceWidth ?? nominalTraceWidth` preference (previously passed raw `this.props.nominalTraceWidth` and also dropped `defaultTraceWidth`).
- Behavior is unchanged when neither prop is set (still 0.15mm), and an explicit `<trace thickness>` still takes precedence.

## Tests

- Updated `tests/components/primitive-components/trace-default-trace-width.test.tsx`, which previously asserted the buggy behavior (`width === 0.15` with a "Bug repro" note saying the fixed behavior should be 0.3) — it now asserts 0.3mm, and its PCB snapshot was regenerated (the routed trace is visibly wider).
- Added `tests/features/board-default-trace-width-autorouting.test.tsx` covering: `nominalTraceWidth` on the board, explicit `thickness` precedence over `defaultTraceWidth`, and the unchanged 0.15mm default.

## How I tested

Ran locally with bun 1.3.14 (full suite is too heavy for my machine, so I ran the directories surrounding the change):

- `bun test tests/features/board-default-trace-width-autorouting.test.tsx tests/components/primitive-components/trace-default-trace-width.test.tsx tests/utils/autorouting/` — all pass (the new tests fail before the fix, reproducing the issue exactly: 0.15 instead of 0.3/0.25)
- `bun test tests/utils/ tests/components/primitive-components/ tests/components/normal-components/ tests/features/ tests/subcircuits/ tests/groups/ tests/drc/ tests/breakout/` — no new failures (`auto-jumper-autorouting`, `schematic-section-rp2040`, and `differential-pair/routing-phase` time out identically on unmodified `main` on my machine)
- `bunx tsc --noEmit` — clean
- `bun run format` — no changes


lib/utils/autorouting/getSimpleRouteJsonFromCircuitJson.ts
```diff
@@ -730,7 +734,7 @@ export const getSimpleRouteJsonFromCircuitJson = ({
           : undefined,
       layerCount: board?.num_layers ?? 2,
       minTraceWidth: Math.min(
-        defaultTraceWidth,
+        fallbackMinTraceWidth,
```
seveibar: Undo changes here, afaik they do not change the implementation and renaming needs a justification

seveibar: (Entire file)

seveibar (commented):
The fix is probably fine, but remove the refactored details in the file i mentioned, try to keep changes minimal

lib/utils/autorouting/getSimpleRouteJsonFromCircuitJson.ts
Hero988: Done in 455bf46 — dropped the rename and the comment; the file is now a 2-line diff. Those two lines are the functional core of the fix rather than a refactor: the `nominalTraceWidth` param already existed on main but was never applied to connections, so every connection without an explicit width fell back to the min width (0.1/0.15) and the board's `defaultTraceWidth` never reached the autorouter. tsc clean, tests pass.

seveibar (approved):

PR #2740, opened 2026-07-21
fix: align transistor pin aliases/accessors with schematic-symbols (pin1=collector, pin2=base, pin3=emitter)

### LINKED ISSUE #3617: regression in the transistor symbol
after updating deps in of the PRs, the transistor symbol port locations changed, here is a repro in core:

Image

Hero988:
Fixes tscircuit/tscircuit#3617

## Root cause

The `schematic-symbols` package numbers the BJT ports as **pin1 = collector, pin2 = base, pin3 = emitter** — identically for `npn_bipolar_transistor` and `pnp_bipolar_transistor`, in both `_horz` and `_vert` orientations (verified against the currently-pinned `schematic-symbols@0.0.232`):

```
npn_bipolar_transistor_horz / pnp_bipolar_transistor_horz:
  ["1","collector"]  x=-0.55  y= 0.29
  ["2","base"]       x= 0.00  y=-0.40
  ["3","emitter"]    x= 0.55  y= 0.29
```

`Transistor.ts` disagreed with this in two places at once:

```ts
// aliases said pin2=emitter, pin3=base (symbol says pin2=base, pin3=emitter)
pin1: ["collector", "c"], pin2: ["emitter", "e"], pin3: ["base", "b"]

// and the accessors even disagreed with the aliases directly above them
emitter = this.portMap.pin1   // pin1 is aliased "collector"!
collector = this.portMap.pin2
base = this.portMap.pin3
```

Because schematic positions are assigned by pin number, ports ended up with **contradictory port_hints** — e.g. pin2 carried both `"base"` (from the symbol) and `"emitter"` (from the alias) — so name-based selectors and downstream consumers resolved pins arbitrarily.

## This was corrupting SPICE netlists, not just symbol positions

`circuit-json-to-spice` resolves BJT terminals by port hints. With the contradictory hints, the switch-circuit test (`ngspice-spice-analysis01-switch`) generated:

```
before:  SSW1 N2 0 ...      ← switch output shorted to ground
         QQ1  N3 0 0 NPN    ← base AND emitter both on node 0 (B-E short, transistor can never turn on)
after:   SSW1 N2 N3 ...
         QQ1  N4 N3 0 NPN   ← collector=N4, base=N3, emitter=GND — correct
```

The old simulation snapshot showed a flat collector voltage (transistor permanently off); the updated snapshot shows the correct 0↔5 V switching waveform. That's why the two snapshot updates in this PR are part of the fix, not incidental churn.

## Answers to the questions raised on the earlier attempts (props#698, props#714, core#2541 — all stale-closed)

- **@imrishabh18 "Wouldn't this mess up the PNP mapping?"** — No: the PNP symbol uses the *same* numbering as NPN (`pin1=collector, pin2=base, pin3=emitter`, verified above), so one mapping fixes both. The new test asserts positions for **both** npn and pnp.
- **@ShiboSoftwareDev "is this consistent within the repo?"** — Both BJT symbol variants and both orientations agree. MOSFET/JFET symbols use their own gate/source/drain labels and are untouched by this change (the `Mosfet` component has its own mapping; `mosfet-symbol-port-sides.test.tsx` still passes).
- Note the earlier core attempt (#2541) proposed `pin2=emitter`, which contradicts the symbols package — merging it would have swapped base and emitter.

## Changes

- `Transistor.ts`: aliases → `pin1: ["collector","c"], pin2: ["base","b"], pin3: ["emitter","e"]`; accessors → `collector=pin1, base=pin2, emitter=pin3`.
- `transistor-ports.test.tsx`: two new tests — (1) the schematic port named collector/base/emitter must sit at the symbol port with that label, for npn **and** pnp (red on `main`: 2 fail → green here); (2) accessors must match the aliases.
- Regenerated the two affected snapshots (schematic layout + the now-physically-correct ngspice switch simulation).

Ran `tsc --noEmit` (clean), `biome check` (clean), and every test file referencing transistors (12 pass / 0 fail), plus the spice-analysis suite.

A follow-up PR to `props` aligning `transistorPins` (`pin1=emitter, pin2=collector` there today) with the same order is ready; it's independent of this fix.

---


seveibar (commented):
This fix for this is likely in schematic-symbols by introducing aliases for collector, emitter and base such that there doesnt need to be a number mapping (can you look into that alternative solution?)

Hero988:
Looked into it — findings:

- schematic-symbols already ships named labels on every BJT variant: all 12 (npn/pnp × horz/vert/up/down/left/right) consistently expose `[["1","collector"],["2","base"],["3","emitter"]]`, so nothing needs adding there for BJTs (MOSFETs also ship `[["1","drain"],["2","source"],["3","gate"]]`).
- `NormalComponent` already concats the symbol port labels into the port's aliases when initializing from the schematic symbol (`NormalComponent.ts` ~404–425), so ports are resolvable by `collector`/`base`/`emitter` with no numbering involved.
- The actual contradiction is local to `Transistor.ts`: the hardcoded alias map (`pin1: ["collector","c"], pin2: ["emitter","e"], pin3: ["base","b"]`) and the accessors (`emitter = portMap.pin1, collector = portMap.pin2, base = portMap.pin3`) disagree with each other *and* with the symbol.

So I'll rework this PR to drop the hardcoded number mapping entirely and resolve collector/base/emitter from the symbol's named port labels. If you'd also like short-form aliases (`c`/`b`/`e`, `g`/`s`/`d`) added to schematic-symbols itself so every consumer gets them for free, happy to do a follow-up PR there.

Hero988:
Reworked as described above (d82d2fc4): `Transistor` no longer hardcodes any pin-number→role map. The schematic symbol's named port labels are now the single source of truth — `initPorts` only derives the short `c`/`b`/`e` aliases by finding which pin the symbol labels `collector`/`base`/`emitter`, and the accessors are now lazy getters via `portMap` (same pattern as `Diode`'s `anode`/`cathode`). Rebased on main; tsc clean; all 310 normal-components tests + both switch spice tests pass.

`Mosfet` needs no equivalent change — it has no hardcoded map and already resolves purely from symbol labels. Happy to add `drain`/`source`/`gate` accessors there in a follow-up if useful.

tests/components/normal-components/transistor-ports.test.tsx
imrishabh18, line 1: One file should only have one test

Hero988: Understood. I will split the added position and accessor scenarios so each test file contains one test.

Hero988: Done in f0905885 — the two added scenarios now live in their own files (`transistor-schematic-port-positions.test.tsx`, `transistor-pin-accessor-aliases.test.tsx`, one test each) and `transistor-ports.test.tsx` is back to its state on main. All three files pass locally; tsc + format clean.

seveibar (approved):

PR #2741, opened 2026-07-21
fix: autorouted traces ignore board defaultTraceWidth/nominalTraceWidth and fall back to minTraceWidth

rushabhcodes:
## Bug

Traces without an explicit `thickness` that go through local autorouting were
always routed at `minTraceWidth` (0.15mm), even when the board specified
`defaultTraceWidth` or `nominalTraceWidth`. The nominal width was accepted as
a prop but silently dropped before reaching the router.

Root cause: `getSimpleRouteJsonFromCircuitJson` accepted a `nominalTraceWidth`
parameter but never used it — it derived the width used to fill in
unset connection widths purely from `minTraceWidth`/`board.min_trace_width`.
The same (wrong) value was then reused to compute the returned `minTraceWidth`
bound, conflating two independent concerns: the minimum allowed trace width
and the default width to apply when a trace doesn't specify one.

## Fix

- `getSimpleRouteJsonFromCircuitJson.ts`: split the two concerns —
  `resolvedMinTraceWidth` (from `minTraceWidth`/`board.min_trace_width`) is
  now used only for the output `minTraceWidth` bound, while
  `defaultTraceWidth` (`nominalTraceWidth ?? resolvedMinTraceWidth`) is used
  to fill in connection widths that weren't explicitly set.
- `Group.ts`: both autorouting call sites now resolve
  `props.defaultTraceWidth ?? props.nominalTraceWidth ?? 0.15` and pass it
  through as `nominalTraceWidth`, so the board-level setting actually reaches
  the router.

## Test plan

- Added `trace-nominal-trace-width.test.tsx`, a fresh repro asserting that a
  board with `nominalTraceWidth="0.25mm"` produces autorouted traces at
  0.25mm (previously they came out at 0.15mm).
- Updated the existing `trace-default-trace-width.test.tsx` repro to assert
  the corrected behavior (0.3mm) instead of documenting the bug.
- Regenerated the affected PCB snapshots.
- Ran the full test suite; no new failures introduced (pre-existing,
  unrelated differential-pair timeout flakiness reproduces identically on
  `main`).


lib/components/primitive-components/Group/Group.ts
```diff
@@ -1935,6 +1935,13 @@ export class Group<Props extends z.ZodType<any, any, any> = typeof groupProps>
     }
   }
 
+  _resolveNominalTraceWidth(props: SubcircuitGroupProps): number | undefined {
+    const rawNominalTraceWidth =
+      props.defaultTraceWidth ?? props.nominalTraceWidth
```
imrishabh18: This seems wrong why `rawNominalTraceWidth` will have `defaultTraceWidth` as the prefered value

imrishabh18 (changes requested):

techmannih:
@rushabhcodes I think he is almost there. https://github.com/tscircuit/core/pull/2739

PR #2742, opened 2026-07-21
test: add clear boxed subcircuit net label repros

MustafaMulla29:
## What changed

Adds two separate visual repros so each boxed-subcircuit net-label bug is isolated and self-explanatory.

### Missing fallback label

The snapshot lists the exact cross-subcircuit traces:

- `power.3V3 -> mcu.3V3`
- `power.GND -> mcu.GND`

The 3V3 trace renders normally. The GND trace disappears; the expected fallback is a matching `power_GND/mcu_GND` label at both GND ports.

### Generated name leak

The second snapshot has only one internal connection: `sensor.STATUS -> sensor.U1.STATUS`. No trace leaves the `sensor` subcircuit.

The current output incorrectly renders `sensor_connectivity_net0`. The expected output is an open `STATUS` port with no net label.

Both snapshots render the declared connections, current bug, and expected output directly on the schematic. This PR intentionally contains no fix and follows the review feedback on #2738.

## Validation

- `bun test tests/repros/repro-boxed-subcircuit-missing-fallback-label.test.tsx`
- `bun test tests/repros/repro-boxed-subcircuit-internal-port-name-leak.test.tsx`


seveibar (changes requested):
NOT CLEAR!!!!!

seveibar:
These are subcircuits! What is the issue? What is supposed to happen? What are the connections!!!!!

MustafaMulla29:
> These are subcircuits! What is the issue? What is supposed to happen? What are the connections!!!!!

is it the correct behaviour to show "....connectivity_net0" netlabels for even not connected pins?

seveibar (approved):

PR #2744, opened 2026-07-21
chore: update calculate-packing version

rushabhcodes:
This pull request updates the `calculate-packing` dependency to a newer version in the `package.json` file. This is a minor maintenance change to ensure the project uses the latest features or bug fixes from the `calculate-packing` library.


MustafaMulla29 (approved):
If the dep update fixes something then it should be also updated in matchpack repo

seveibar (approved):

PR #2749, opened 2026-07-21
repro(pcb): add DIP-8 fabrication note pin label rendering repro

itisrohit:
DIP-8 fabrication note pin label rendering bug.

The snapshot shows unresolved `{pin1}`…`{pin8}` placeholders in fabrication notes.


seveibar (changes requested):
This should be in footprinter not core

itisrohit:
> This should be in footprinter not core

Thanks! I'll move the reproduction to footprinter.

itisrohit:
Closing this since the reproduction belongs in tscircuit/footprinter as suggested. I've moved the repro there.

PR #2760, opened 2026-07-22
Prefer readable pin names for anonymous schematic nets

MustafaMulla29:
## Summary

- replace generated slash-joined schematic net labels such as `A/B/C` with one readable component-pin label
- choose a pin from the connected component with the highest pin count, using the lower component name and lower pin number as deterministic tie-breakers
- apply the fallback consistently to explicit traces, component `connections`, and source-port fallbacks
- preserve explicit trace labels and named nets such as `GND` and `VCC`
- refresh affected schematic and simulation snapshots

## Why

Slash-joined anonymous labels can become extremely long and make larger schematics difficult to read. A stable label such as `U1_X1` keeps the schematic compact while still identifying the net by a useful endpoint.

## Validation

- `bun run build`
- 33 affected schematic and simulation snapshot tests pass
- focused explicit-label, named-net, lower-pin, and `connections` regressions pass
- formatting and diff checks pass


seveibar (commented):
can you try something, can you change it to the largest chip's pin name, preferring the lower pin name, e.g. U1_X1

i think this might be a better/readable default, i'm noticing these circuits are hard to read

seveibar (approved):

PR #2761, opened 2026-07-22
Render canonical USB-C ports when Parts Engine resolution fails

MustafaMulla29:
## Summary

- renders the canonical 16 USB-C source and schematic ports when standard connector part resolution fails
- keeps the existing Parts Engine failure warning and does not synthesize PCB pads or other physical footprint geometry
- shares one canonical USB-C pin definition between fetched-part normalization and the fallback
- updates the merged repro to snapshot and assert the corrected fallback behavior

## Root cause

`<connector standard="usb_c" />` previously derived all ports from fetched part Circuit JSON. A failed `findPart` call, failed `fetchPartCircuitJson` call, empty Circuit JSON result, or unavailable fetch method therefore left the connector with no ports and rendered an empty schematic box.

The failure handler now adds canonical logical ports only after resolution has failed. Successful resolution continues to use the fetched part's manufacturer-specific pin numbering and physical footprint.

## Validation

- USB-C Parts Engine failure paths: fetch rejection, part not found, Circuit JSON not found, and missing fetch method
- successful USB-C fetch, aliases, MPN fallback, CAD model priority, canonical pin ordering across two physical parts, and explicit-footprint bypass
- `bunx tsc --noEmit`
- `bun run build`
- full `bun test`: 1,214 passed and 31 skipped; one unrelated differential-pair PCB snapshot mismatch also reproduces on a clean `origin/main` worktree


lib/utils/connectors/convertCircuitJsonToUsbCStandardCircuitJson.ts
```diff
@@ -52,7 +29,7 @@ export const convertCircuitJsonToUsbCStandardCircuitJson = (
   }
 
   const canonicalHintsByPin: Record<`pin${number}`, string[]> = {}
-  for (const { label, aliases } of STANDARD_USB_C_PIN_LABELS) {
+  for (const { label, aliases } of USB_C_CANONICAL_PIN_DEFINITIONS) {
```
seveibar: i think the old name was better

seveibar (approved):

PR #2771, opened 2026-07-22
repro: auto-generated pin names leak into schematic net labels

itisrohit:
When a chip has no explicit `pinLabels`, ports get auto-generated names (`pin1`, `pin4`). Connections that the schematic router can't draw as traces fall back to labels from `getSourcePortNetLabelText`, which produces `U1_pin4` — an internal implementation detail visible on the schematic.


imrishabh18 (commented):
What is the issue?

itisrohit:
@imrishabh18 The reproduction connects `U1.pin4` to `R1.pin1` using the `connections` prop, without defining a named net or signal.

When the schematic is rendered, it displays `U1_pin4` as a visible label. From a user's perspective, this appears to be a signal or net name, even though they never defined `U1_pin4` anywhere in their circuit.

The issue I'm trying to clarify is whether displaying labels like `U1_pin4` is the intended behavior in this situation, or whether the schematic should avoid exposing names that the user never explicitly defined.

mohan-bee:
> @imrishabh18 The reproduction connects `U1.pin4` to `R1.pin1` using the `connections` prop, without defining a named net or signal.
> 
> When the schematic is rendered, it displays `U1_pin4` as a visible label. From a user's perspective, this appears to be a signal or net name, even though they never defined `U1_pin4` anywhere in their circuit.
> 
> The issue I'm trying to clarify is whether displaying labels like `U1_pin4` is the intended behavior in this situation, or whether the schematic should avoid exposing names that the user never explicitly defined.

i think the behaviour we currently have is intended

itisrohit:
> > @imrishabh18 The reproduction connects `U1.pin4` to `R1.pin1` using the `connections` prop, without defining a named net or signal.
> > When the schematic is rendered, it displays `U1_pin4` as a visible label. From a user's perspective, this appears to be a signal or net name, even though they never defined `U1_pin4` anywhere in their circuit.
> > The issue I'm trying to clarify is whether displaying labels like `U1_pin4` is the intended behavior in this situation, or whether the schematic should avoid exposing names that the user never explicitly defined.
> 
> i think the behaviour we currently have is intended

I dug into the history and found that PR #2760 changed the fallback selection from `R1_pin2/U1_pin2` to `U1_pin2`, but I couldn't find any discussion about whether autogenerated labels themselves were intended to be shown.

I also noticed that existing tests seem to prefer meaningful labels (for example `U1_X1`) over autogenerated ones (such as `R1_pin1`), but they don't appear to cover the case where every connected pin has an autogenerated name.

Was the intent that autogenerated labels like `U1_pin4` should also be user-visible, or is that simply the current behavior of the fallback label selection algorithm?

seveibar:
this isn't a bug, this is a new feature introduced today

seveibar:
reduces the size of net labels

PR #2776, opened 2026-07-22
repro: mosfet doesn't propagate manufacturerPartNumber and supplierPartNumbers

itisrohit:
`manufacturerPartNumber` and `supplierPartNumbers` passed to `<mosfet />` are
silently dropped — they never appear in the generated `source_component`.

The test asserts that
`source_component.manufacturer_part_number` and
`source_component.supplier_part_numbers` **should** contain the provided
values, but they are currently `undefined`.

Same pattern as the `<switch />` fix in PR #2765:
`Mosfet#doInitialSourceRender()` overrides the base class without calling
`super.doInitialSourceRender()` and omits these fields from its
`source_component.insert()` call.


imrishabh18 (commented):
You don't need a repro for this small test, just create the fix directly

seveibar (approved):

mohan-bee (approved):

PR #2782, opened 2026-07-23
Add differential pair post-routing phase

0hmX:
## Summary

- add a differential-pair post-processing render phase after PCB routing and before copper pours and DRC
- invoke one solver per routed subcircuit with all pair constraints and the complete routed Simple Route JSON
- validate solver output before applying any changes
- atomically replace pair routes and vias while preserving unrelated and manual routes
- preserve routed copper and emit actionable autorouting errors when solving or validation fails
- add integration, failure, ordering, event, and visual snapshot coverage

## Why

Core previously completed PCB routing without a dedicated boundary for coordinated differential-pair length matching. Applying solver output also needs to be all-or-nothing so malformed output cannot leave the circuit database partially updated.

## Impact

This wires core to the initial no-op solver contract without changing successful route geometry. It establishes the phase and safety boundary needed for the future length-matching algorithm.

## Related work

- Parent PRD: tscircuit/length-matching-post-process#1
- Closes tscircuit/length-matching-post-process#4
- Closes tscircuit/length-matching-post-process#5
- Closes tscircuit/length-matching-post-process#7
- Depends on solver scaffold: tscircuit/length-matching-post-process#9
- Merge/release dependency: publish `@tscircuit/length-matching-solver@0.0.1`, tracked by tscircuit/length-matching-post-process#8

## Validation

- 18 targeted tests passed
- `bunx tsc --noEmit`
- `bun run build`
- full suite: 1,231 passed; one existing routing test timed out under full-suite load and passed independently
- `git diff --check`


0hmX:
Superseded by #2818, which integrates the complete solver implementation. Merge order is tscircuit/length-matching-solver#28 first, then #2818.

PR #2785, opened 2026-07-23
Add TPS61222 schematic trace intersection repro

techmannih:


MustafaMulla29 (approved):
Screenshot 2026-07-23 at 12 04 02 PM

If i remove `schMaxTraceDistance` then its not reproducable, not sure why  `schMaxTraceDistance` in this particular circuit and that too a 160mm.

PR #2789, opened 2026-07-23
fix: propagate manufacturerPartNumber in resonator and 10 other components

itisrohit:
## Summary

Propagate `manufacturerPartNumber` (and `mfn` alias) to `source_component.manufacturer_part_number` in every component whose `doInitialSourceRender()` override was silently dropping it.

Each component already wrote `supplier_part_numbers` and `display_name`; `manufacturer_part_number` was simply forgotten — a copy-paste omission dating back to when the base class first added the field in PR #552.

## Components fixed (11 total)

Resonator, OpAmp, PowerSource, Ammeter, Inductor, CurrentSource, PushButton, VoltageSource, TestPoint, PinHeader, Battery

## Implementation

Each fix adds a single line to the `source_component.insert({...})` call:

```ts
manufacturer_part_number: props.manufacturerPartNumber ?? props.mfn,
```

Matching the pattern already used by the base class (`NormalComponent.ts:692`) and every properly-implemented peer (Resistor, Capacitor, Diode, Mosfet, Switch, Chip, Jumper, SolderJumper, Pinout, Connector, Led, etc.).

## Testing

- `manufacturer-part-number-propagation.test.tsx` — focused data-propagation checks for all 10 newly-fixed components
- `resonator-supplier-part-numbers.test.tsx` — existing Resonator check (with PCB + schematic snapshots)

Both pass. No existing tests are affected.

PR #2790, opened 2026-07-23
fix: propagate manufacturerPartNumber in inductor

itisrohit:
The `<inductor />` component accepts `manufacturerPartNumber` (and `mfn`)
via its Zod props but silently drops it — the field never reaches
`source_component.manufacturer_part_number` in the circuit JSON output.

Add `manufacturer_part_number` to `doInitialSourceRender()`, matching
the pattern already used by the base class and every other component
(Resistor, Capacitor, Diode, Mosfet, Switch, etc.).

Same pattern as the Resonator fix in #2789.


techmannih (commented):
@itisrohit Can you please add all missing manufacturerPartNumber

mohan-bee:
> @itisrohit Can you please add all missing manufacturerPartNumber

yes that works, add that in the previous PR itself !!

itisrohit:
Closing this PR in favor of #2789, which has been updated to include all remaining `manufacturer_part_number` propagation fixes requested during review. Thanks for the feedback!

PR #2792, opened 2026-07-23
repro155 explict netlabes draws extra traces

mohan-bee:


tests/repros/repro155-explicit-netlabels-draws-extra-traces.test.tsx
```diff
@@ -0,0 +1,63 @@
...
+        anchorSide="bottom"
+        connectsTo={[sel.C1.pin1, sel.C2.pin1, sel.C4.pin1]}
+        schY={4}
+      />
+      <netlabel
+        net="GND"
+        anchorSide="top"
+        connectsTo={[sel.C1.pin2, sel.C4.pin2]}
+        schY={2.1}
+      />
```
techmannih, line 56: I think we are not using manual netlabels

mohan-bee: yea me too confused about that .. because i see sometimes it creates with manual netlabels !! ..

MustafaMulla29: I think this is a valid issue, if the trace was going through the manual netlabels then that would have been not valid as schematic-trace-solver doesn't know about the manual netlabel obstacles.

MustafaMulla29:
@mohan-bee this is a valid repro

MustafaMulla29 (approved):

PR #2793, opened 2026-07-23
fix: propagate allowOffBoard to pcb_component in NormalComponent, Jumper, and SolderJumper

itisrohit:
`allowOffBoard` is defined on `CommonComponentProps` and accepted by every component, but was only serialized to `pcb_component.is_allowed_to_be_off_board` by `Chip`. The base class `NormalComponent.doInitialPcbComponentRender()` and the overrides in `Jumper` and `SolderJumper` silently dropped the prop, causing `allowOffBoard` to be ignored for all non-Chip components.

This adds the missing `is_allowed_to_be_off_board: props.allowOffBoard ?? false` to all three `db.pcb_component.insert()` call sites, matching the existing pattern in `Chip`.


tests/components/normal-components/allow-off-board-propagation.test.tsx
imrishabh18, line 1: One test in each file, you could add them in a single board and test it

itisrohit: Thanks! Done.

tests/components/normal-components/allow-off-board-propagation.test.tsx
```diff
@@ -0,0 +1,42 @@
...
+      <resistor
+        name="R1"
+        resistance="1k"
+        footprint="0402"
+        pcbX={20}
+        pcbY={0}
+        allowOffBoard
+      />
+      <jumper name="J1" pcbX={20} pcbY={0} allowOffBoard />
+      <solderjumper name="SJ1" pcbX={20} pcbY={0} allowOffBoard />
```
imrishabh18: I don't see these in the pcb view, add different position and footprint

tests/components/normal-components/allow-off-board-propagation.test.tsx
```diff
@@ -0,0 +1,42 @@
...
+    expect(pcb_component?.is_allowed_to_be_off_board).toBe(true)
+  }
+
+  const outsideErrors = circuit
+    .getCircuitJson()
+    .filter((el) => el.type === "pcb_component_outside_board_error")
+  expect(outsideErrors.length).toBe(0)
+
+  expect(circuit).toMatchPcbSnapshot(import.meta.path)
+  expect(circuit).toMatchSchematicSnapshot(import.meta.path)
```
imrishabh18:
```suggestion
```

itisrohit:
@imrishabh18 Done! Updated the positions and footprints so `J1` and `SJ1` are visible in the PCB snapshot, and removed the schematic snapshot.

imrishabh18 (approved):

seveibar (approved):

mohan-bee (approved):

PR #2796, opened 2026-07-23
Support schSectionName on schematic boxes

seveibar:
## Summary

- update `@tscircuit/props` to `0.0.590`, which adds `schSectionName` to `<schematicbox />`
- key schematic trace-solver section membership by rendered `schematic_component_id` instead of source component name
- add a regression test and schematic snapshot covering two sectioned schematic-box representations of the same source chip

Depends on and follows [tscircuit/props#750](https://github.com/tscircuit/props/pull/750), which is merged and published in `@tscircuit/props@0.0.590`.

## Why

A chip-backed `<schematicbox>` creates its own rendered schematic component while referencing another component's source identity. Core previously derived trace-solver sections through source component names, so schematic boxes named `U1 Power` and `U1 I/O` both reached the solver without a `sectionId` when they represented source chip `U1`.

Associating sections with the rendered schematic component preserves the existing behavior for normal components and also allows multiple schematic representations of one source chip to belong to different sections.

## Impact

`schSectionName` now reaches the schematic trace solver for chip-backed schematic boxes, keeping their routing scoped to the intended section. Existing `schSheetName` support remains unchanged.

## Validation

- regression test failed before the implementation with both section IDs `undefined`, then passed after the change
- focused section and trace-solver tests: 4 passing
- `bunx tsc --noEmit`
- `bun run build`
- `bun run format`
- `git diff --check`
- full `bun test`: 1,235 passing, 31 skipped, 1 unrelated snapshot mismatch in `differential-pair/port-selector`; the same mismatch reproduces on untouched `main` with `@tscircuit/props@0.0.589`
- visually inspected the new schematic snapshot


imrishabh18 (approved):

seveibar:
@copilot resolve the merge conflicts in this pull request

PR #2806, opened 2026-07-24
Use schematic port IDs in the schematic trace solver

imrishabh18:
## Summary

- use canonical `schematic_port_id` values as schematic trace-solver pin IDs end to end
- remove `getSchematicPortSelector` and the selector-to-port translation maps
- fix net-label routing when a `<chip>` and `<schematicbox chipRef="...">` render the same source pin
- strengthen the repro156 regression test and refresh the affected schematic snapshots

## Root cause

The trace solver used CSS-like selectors such as `U1.8` as its internal pin identity. That works while one source pin has one schematic representation.

In repro156, the chip and its schematic box both render the same source pin. Both ports therefore produced the same selector. Because the routing code stored ports in maps keyed by that selector, one rendered port overwrote the other and the net label could be routed to the wrong representation.

Each rendered schematic port already has a unique canonical `schematic_port_id`, so creating another identity from the component name and pin number was unnecessary.

## Fix

The solver now receives `schematic_port_id` directly and its output is consumed as schematic port IDs. There is no component-name suffix, port-ID suffix, or fallback selector.

The existing internally-connected-pin dedup behavior is preserved: its representative port ID is used only to detect duplicates, while the original schematic port ID is passed to the solver. The adjacent port lookup is keyed by `schematic_port_id` so separate schematic representations of one source port do not overwrite each other.

Removing `getSchematicPortSelector` only removes this internal routing identity layer. User-facing selectors such as `U1.GND` and `.pin1` still work normally; the public selector system is unchanged.

## Snapshot impact

The minimum-spanning-tree routing solver uses pin IDs to break equal-cost ties. Replacing derived selector strings with canonical port IDs can therefore select a different, but electrically equivalent, route in tied layouts. The affected schematic snapshots are refreshed accordingly.

The pushbutton internally-connected-pin snapshot is unchanged.

## Validation

- repro156 regression test passes
- `pushbutton-pin-configurations` and repro155 internally-connected-pin tests pass
- relevant schematic trace and section tests pass
- TypeScript typecheck passes
- package build passes


seveibar (commented):
can we improve the semantics

tests/repros/__snapshots__/repro-rp2040-gamepad-trace-alignment-schematic.snap.svg
imrishabh18, line 1: It used to look better before

tests/repros/__snapshots__/repro84_schematic_pin_out-schematic.snap.svg
imrishabh18, line 1: GND is not symbol anymore

tests/repros/__snapshots__/repro91-schematic-trace-junction-schematic.snap.svg
imrishabh18, line 1: traces very close

lib/components/primitive-components/Group/Group_doInitialSchematicTraceRender/createSchematicTraceSolverInputProblem.ts
```diff
@@ -440,9 +413,7 @@ export function createSchematicTraceSolverInputProblem(
 
       netConnections.push({
         netId: userNetId,
-        pinIds: uniqueSchematicPortIds.map(
-          (portId) => schematicPortIdToPinId.get(portId)!,
-        ),
+        pinIds: uniqueSchematicPortIds,
```
seveibar, line 416:
```suggestion
        schematicPortIds: uniqueSchematicPortIds,
```

seveibar (approved):

PR #2809, opened 2026-07-24
Add standalone schematicsymbol element

MustafaMulla29:
## Summary

- add a standalone `<schematicsymbol />` backed by `schematic-symbols`
- implement it as a schematic-only `PrimitiveComponent`
- support explicit physical-port mapping when both `chipRef` and `connections` are provided
- support explicit schematic sheets/sections, placement, and rotation
- register the JSX intrinsic/export and bump `@tscircuit/props` to `^0.0.591`

## Connection API

`connections` maps labels exposed by `symbolName` (including `pinN` aliases) to explicit physical chip-port selectors:

```tsx
<schematicsymbol
  name="A"
  chipRef=".Q1"
  symbolName="n_channel_e_mosfet_transistor_horz"
  connections={{
    gate: "Q1.G1",
    source: "Q1.S1",
    drain: "Q1.D1",
  }}
/>

<trace from=".R1 > .pin2" to=".Q1 > .G1" />
```

The mapping does not create electrical traces. It projects the real chip ports into the symbol; normal `<trace>` elements continue to target the physical chip.

There is intentionally no automatic label matching. Supplying `chipRef` without `connections`, or `connections` without `chipRef`, emits a `source_property_ignored_warning` for the unmatched property and renders the symbol standalone. A symbol with neither remains a standalone schematic element.

## Architecture

- `SchematicSymbol` creates its own source component so the renderer displays `name` using the same `{REF}` mechanism as other components
- mapped schematic ports reuse the referenced physical chip's real `source_port_id` values
- qualified symbol names such as `_horz` and `_right` rotate through the shared `PrimitiveComponent` resolver, without a `SchematicSymbol` override
- connected components are not inferred onto a symbol's sheet; each visible component declares its own `schSheetName`
- no manual `schematic_text` insertion
- no shared `PrimitiveComponent`, `NormalComponent`, or `Port` lifecycle changes
- no PCB representation or missing-footprint behavior
- no `internalcircuit` dependency
- `displayName` rendering remains explicitly deferred until `schematic_component` has a `display_name` field

## Coverage

- standalone library symbol
- MOSFET and diode mappings with explicit connections
- two MOSFET representations (`A` and `B`) of one physical chip without a schematic sheet
- MOSFET and op-amp representations with separate traces across two explicitly assigned schematic sheets
- PCB snapshots confirming the referenced physical chips and their routed connections, without duplicate PCB components
- schematic-section participation
- warning-and-ignore behavior for incomplete `chipRef` + `connections` pairs, with no automatic port matching

## Validation

- 12 targeted schematic-symbol, symbol-rotation, and sheet regression tests pass
- `bun run build`
- `bunx tsc --noEmit`
- Biome checks for modified TypeScript files
- `git diff --check`
- full suite: 1,239 pass, 31 skip; one unrelated differential-pair PCB snapshot mismatch also reproduces in isolation


tests/components/primitive-components/schematic-symbol-chipref-without-connections-warning.test.tsx
```diff
@@ -0,0 +1,35 @@
...
+  circuit.add(
+    <board>
+      <chip
+        name="Q1"
+        footprint="sot23"
+        noSchematicRepresentation
+        pinLabels={{ pin1: "G", pin2: "S", pin3: "D" }}
+      />
+      <schematicsymbol
+        name="A"
```
seveibar, line 16:
```suggestion
        name="Q1A"
```

seveibar (approved):

PR #2813, opened 2026-07-24
fix: default keepout to all board copper layers (#2613)

### LINKED ISSUE #2613: keepout silently defaults to layers=["top"] - bottom-layer routing ignores it
## Behavior

`<keepout>` silently defaults to `layers={["top"]}` (Keepout.ts: `if (!layers) layers = ["top"]`). Nothing in the props or docs makes this obvious, and a keepout that only guards one copper layer is rarely what a board author wants.

## How it bit

2-layer board with a routed slot (`<cutout>`) paired with a keepout to hold the router off the slot. The keepout worked on `top` - and the autorouter promptly routed SDA along the slot on `bottom`, inside the copper-to-edge clearance (KiCad DRC `copper_edge_clearance` failures). Nothing warns that the guard was half-active.

Repro sketch:

```tsx
  <cutout shape="rect" width="8mm" height="1.2mm" />
  <keepout shape="rect" width="8.7mm" height="1.9mm" />  {/* top-only! */}
  <resistor name="R1" resistance="1k" footprint="0402" pcbX="-8mm" />
  <resistor name="R2" resistance="1k" footprint="0402" pcbX="8mm" />
  <trace from=".R1 .pin2" to=".R2 .pin1" />
</board>
```

Route can go on `bottom` straight through the keepout region.

## Suggestion

Default `<keepout>` to **all copper layers** (KiCad rule areas default this way too), or require `layers` explicitly. Top-only as a silent default is a footgun.

DPS0340:
Fixes #2613.

## Problem

`<keepout>` silently defaults to `layers={["top"]}` (`Keepout.ts`). As reported, a 2-layer board pairing a routed slot with a keepout only guarded the top copper — the autorouter routed SDA along the slot on `bottom`, inside the copper-to-edge clearance (KiCad DRC `copper_edge_clearance` failures), with no warning that the guard was half-active.

## Fix

Keepouts without an explicit `layer`/`layers` prop now default to **all of the board's copper layers**, via the same `_getBoard()?.allLayers` source PlatedHole already uses — so 2-layer boards get `["top","bottom"]` and multi-layer boards guard every copper layer. KiCad rule areas default the same way. Explicit `layer=` / `layers=` behavior is unchanged (covered by the existing "supports layer and layers props" test).

## Verification

- New test: a keepout without layer props emits `pcb_keepout.layers = ["bottom","top"]` (fails on `main` with `["top"]`)
- Existing explicit-layers test unchanged and passing
- Snapshot updates: `keepout-pcb` / `keepout-get-pcb-size-pcb` — the default keepouts now render on both layers
- `bunx tsc --noEmit` / `biome check` — clean


DPS0340:
@seveibar could I get a review on this when you have a moment?

It's a small one — 4 files, +36/-0 total — fixing #2613 (reported by @maci0): `<keepout>` silently defaults to `layers={["top"]}`, so on a 2-layer board a keepout only guards the top copper and the autorouter routes straight through it on the bottom.

The fix defaults keepout to all board copper layers instead of just `top`. Explicit `layers` props are untouched.

CI is green and it's mergeable against current `main`.

DPS0340:
@imrishabh18 @mohan-bee — tagging you as well since you've merged most recently here (@seveibar merged my #2814, so him too). Whoever picks it up first.

Small and self-contained: the logic change is **7 lines in `Keepout.ts`**; the rest is one test plus two snapshot refreshes.

Fixes #2613. `<keepout>` silently defaulted to `layers={["top"]}`, so on a 2-layer board a keepout guarding a routed slot only covered top copper — the autorouter then routed on `bottom` straight through it, producing KiCad `copper_edge_clearance` DRC failures with no warning that the guard was half-active.

Keepouts without an explicit `layer`/`layers` now default to all board copper layers, using the same `_getBoard()?.allLayers` source `PlatedHole` already uses. That matches how KiCad rule areas default. Explicit `layer=`/`layers=` behavior is untouched and still covered by the existing test.

```
main:  pcb_keepout.layers = ["top"]
here:  pcb_keepout.layers = ["bottom","top"]
```

New test fails on `main`. `tsc --noEmit` and `biome check` clean, CI green, `MERGEABLE / CLEAN`.

DPS0340:
Rebased onto current `main` (`b3c1714`, v0.0.1512) — this repo moves fast enough that the PR's base had drifted. No conflicts, and re-verified on the new base:

- `bun test tests/components/primitive-components/keepout.test.tsx` → 3 pass / 0 fail, including the new "Keepout without layers defaults to all board copper layers"
- `tsc --noEmit` clean
- `biome format` clean
- CI here: 10/10 checks green, `MERGEABLE / CLEAN`

Unchanged in substance: 7 lines in `Keepout.ts` plus a test and two snapshot refreshes.

DPS0340:
One piece of context that might matter for prioritising this: **a previous fix for #2613 was lost to stalebot, not rejected.**

`core#2726` by @holistis targeted the same issue with nearly the same shape (`Keepout.ts` + test + snapshot). It was opened 07-20, received no reviews, and `github-actions[bot]` closed it on 07-24 under this repo's `days-before-pr-stale: 2 / days-before-pr-close: 1` policy.

So #2613 has now had two independent fixes written for it, and the first one aged out before anyone looked. This PR is the second attempt at the same bug.

Not asking for special treatment — just flagging that the issue is real (a keepout silently guarding only top copper while the autorouter routes on bottom, producing KiCad `copper_edge_clearance` failures), the fix is 7 lines, and it's been written twice now.

Currently rebased onto `b3c1714`, 10/10 checks green, `MERGEABLE / CLEAN`. @imrishabh18 @mohan-bee @seveibar

Issue #2810, opened 2026-07-24
Zod schema allocation dominates circuit-JSON generation heap (Zod v3 eager-bind + non-memoized get config())


itsbalamurali:
### Summary

Circuit-JSON generation retains a very large number of Zod schema instances and
their eagerly-bound methods, dominating the heap on mid-size boards. Full
evidence, heap snapshot and retainer analysis are in
tscircuit/tscircuit#4077 — this issue tracks the fix, which is in
`@tscircuit/core`.

### Why this repo

`@tscircuit/core` is where both halves live:

- it depends on `zod ^3.25.67` (Zod v3 eager-binds ~15 methods per schema
  instance in the `ZodType` constructor), and
- it defines every component's `get config()`, where the base
  `PrimitiveComponent` returns a freshly-constructed `z.object({})` per access
  rather than a memoized/shared schema.

`@tscircuit/props` has no direct `zod` dependency and no `get config()`.

### The two code-level causes

1. **Zod v3 per-instance method binding.** Each schema instance carries ~15
   bound closures (`.parse`, `.safeParse`, `.parseAsync`, `.refine`,
   `.superRefine`, `.optional`, `.nullable`, `.array`, …). The snapshot shows
   ~714k `native_bind` nodes retained via `.bound_this` across ~47k instances.

2. **`get config()` rebuilds schemas.** `Net`/`Trace` return module-level
   consts (`zodProps: netProps`, `zodProps: traceProps`), but the base
   `get config()` returns a fresh `z.object({})`, and `config` is a plain
   getter with no memo cache. The same schema shape is reconstructed thousands
   of times (~9,952 identical `ZodOptional` instances in one build).

### Suggested fix (pure-TS)

1. **Migrate to Zod 4** (https://zod.dev/v4). v4 removes the eager per-instance
   method binding, which eliminates the ~714k `native_bind` closures wholesale,
   and documents large per-schema memory reductions. Biggest, lowest-risk win.

2. **Memoize `get config()` / hoist schemas.** Compute each component's
   `zodProps` once (static per class or module const) instead of returning a new
   `z.object({})` from a getter. The base `PrimitiveComponent` inline
   `z.object({})` is the clearest offender.

3. **Reuse parsed props.** `_parsedProps` is read widely; verify props are
   validated once and the parsed result reused rather than re-running
   `.parse`/`.optional()` per element.

### Verification

Measured transpiler-independent (same wall under tsx/SWC/oxc), and reproduced
[...]

PR #2814, opened 2026-07-24
perf: share the base PrimitiveComponent config instead of rebuilding per access (#2810)

DPS0340:
Progress on #2810 — the low-risk first slice of the suggested pure-TS fix.

## Change

The base `PrimitiveComponent.get config()` constructed a **fresh** `z.object({}).passthrough()` on every access:

```ts
get config(): BaseComponentConfig {
  return { componentName: "", zodProps: z.object({}).passthrough() }
}
```

`config` is a plain getter with no memo cache, accessed thousands of times per build, and each Zod v3 schema instance eagerly binds ~15 methods (`.parse`, `.safeParse`, `.refine`, …). The heap snapshot in tscircuit/tscircuit#4077 shows ~714k `native_bind` closures retained across ~47k schema instances, with the non-memoized base `get config()` called out as one of the two code-level causes.

The base config is now a module-level shared const — components that don't override `config` no longer allocate a schema (plus its ~15 bound closures) per access. Subclass configs already return module-level `zodProps` (`netProps`, `traceProps`, …) and are unchanged.

## Scope

This intentionally does **not** attempt the Zod v4 migration or a general per-class config memo (both larger, riskier follow-ups listed in the issue). It removes the one unambiguous per-access allocation.

## Verification

- Full `tests/components` suite: 562 pass, 1 fail — the failing `registers a differential pair using port selectors` test fails identically on `main` (pre-existing, unrelated)
- `bunx tsc --noEmit` / `biome check` — clean

### DIFF
--- a/lib/components/base-components/PrimitiveComponent/PrimitiveComponent.ts
+++ b/lib/components/base-components/PrimitiveComponent/PrimitiveComponent.ts
@@ -63,6 +63,17 @@ export interface BaseComponentConfig {
  * React subtrees or explicit handling of the "footprint" prop. But otherwise
  * has most of the features of a NormalComponent.
  */
+/**
+ * Shared base config: `get config()` is a plain getter accessed thousands of
+ * times per build, and constructing a fresh `z.object({}).passthrough()` per
+ * access retains ~15 eagerly-bound Zod v3 methods per instance — a major
+ * contributor to the heap growth in #2810.
+ */
+const basePrimitiveComponentConfig: BaseComponentConfig = {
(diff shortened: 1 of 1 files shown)


DPS0340:
Hi @seveibar — this is the small, independently reviewable config-allocation half of #2810. It shares the base PrimitiveComponent config instead of constructing a Zod schema per access; all CI checks are green. The Zod 4 package migration is deliberately separate, so this can land on its own whenever you have time to review.

seveibar (approved):

PR #2826, opened 2026-07-24
fix: USB-C connector with no partsEngine produces zero ports

itisrohit:
When `<connector standard="usb_c" />` is used without a `partsEngine` on the board, the guard at Connector.ts:375-376 exits early without calling the fallback handler, producing zero source/schematic ports.

The two-part guard (`partsEngine && !partsEngine.fetchPartCircuitJson` at line 368 + `if (!fetchPartCircuitJson) return` at line 375) is collapsed into a single `if (!partsEngine?.fetchPartCircuitJson)` that routes through `_handleStandardConnectorCircuitJsonFailure()`, producing the canonical 16 USB-C pin ports and emitting a `source_part_not_found_warning` containing "no partsEngine configured".


seveibar:
@itisrohit cmon

Issue #2621, opened 2026-07-09
copperpour stored as one pcb_copper_pour per fill fragment: KiCad export emits a zone per fragment and refills destroy the pour


maci0:
A single `<copperpour>` whose fill is split by traces produces one `pcb_copper_pour` record per disjoint fill fragment, and the KiCad exporter (correctly, given its input) emits one `(zone)` per record. On a real board we got 65+ GND zones from one pour. Minimal repro (a full-width trace splits the pour; GND anchors on both sides keep both fragments):

```tsx
export default () => (
  <board width="20mm" height="20mm">
    <resistor name="R1" resistance="1k" footprint="0402" pcbX={-9} pcbY={-5} />
    <resistor name="R2" resistance="1k" footprint="0402" pcbX={9} pcbY={-5} />
    <resistor name="R3" resistance="1k" footprint="0402" pcbX={0} pcbY={5} />
    <trace from=".R1 > .pin2" to=".R2 > .pin1" />
    <trace from=".R3 > .pin1" to="net.GND" />
    <via connectsTo="net.GND" holeDiameter="0.3mm" outerDiameter="0.6mm"
         fromLayer="top" toLayer="bottom" pcbX={0} pcbY={-8} />
    <copperpour connectsTo="net.GND" layer="top" clearance="0.3mm" />
  </board>
)
```

Circuit json: 2 `pcb_copper_pour` records (brep outer rings spanning y[-9.7,-5.38] and y[-4.62,9.7]). KiCad export: 2 zones, each `(net_name GND)`, each outline being one fragment's computed fill shape.

Expected: one zone per `<copperpour>` whose outline is the requested pour region, with the fill left for KiCad to compute.

Why it matters: in KiCad the zone outline is design intent and the fill is derived. With outline == previous fill fragment, any refill (edit + B, or a pcbnew script) reapplies clearances inside the already-shrunk outline, so copper shrinks every refill and fragments that split further are deleted as islands. Dozens of independent zones per pour also make the file uneditable, and a script that unions fragments into one zone cannot save the result (a KiCad zone keeps a single outline).

Cause: `CopperPour.doInitialPcbCopperPourRender` (`lib/components/primitive-components/CopperPour/CopperPour.ts:73-87`) inserts one record per solver `brep_shape`, drops the source region, and the records carry no id linking them to the pour element, so downstream cannot even union them per pour.

[...]

PR #2829, opened 2026-07-25
fix: set pcb_group_id on copper pour records

DPS0340:
Related to #2621 (does not close it — see "Scope" below).

`CopperPour` was the only PCB primitive that never populated `pcb_group_id` on the records it emits.

Every other primitive passes the enclosing group through — `PcbVia`, `CourtyardRect`, `SilkscreenPath`, `Hole`, `Port`, and `PcbNoteDimension` all use `this.getGroup()?.pcb_group_id` — so copper pours were inconsistently ungrouped in the emitted Circuit JSON.

### Evidence

With a `<via>` and a `<copperpour>` inside the same `<group name="G1">`, before this change:

```
via group:  [ "pcb_group_0" ]
pour group: [ undefined ]
```

Same group, same board, different answer. After the change both report `pcb_group_0`.

### Change

One line in `doInitialPcbCopperPourRender`, matching the existing convention, plus a regression test that asserts the pour's `pcb_group_id` equals the via's rather than just checking it's non-null.

### Scope

This is a consistency fix and is deliberately narrow. It is **not** a fix for #2621 (one `pcb_copper_pour` record per fill fragment) — a board-level pour still has no group, so that issue needs a circuit-json schema decision, which I've laid out with options in a comment there.

### Verification

- `bunx tsc --noEmit` — clean
- `biome format` — clean
- All 7 copper-pour test files: 15 pass / 0 fail

### DIFF
--- a/lib/components/primitive-components/CopperPour/CopperPour.ts
+++ b/lib/components/primitive-components/CopperPour/CopperPour.ts
@@ -76,6 +76,7 @@ export class CopperPour extends PrimitiveComponent<typeof copperPourProps> {
           layer: props.layer,
           brep_shape,
           source_net_id: net.source_net_id,
+          pcb_group_id: this.getGroup()?.pcb_group_id ?? undefined,
           subcircuit_id: subcircuit?.subcircuit_id ?? undefined,
           covered_with_solder_mask: coveredWithSolderMask,
         } as PcbCopperPour)
--- a/tests/repros/repro-copper-pour-group-id.test.tsx
+++ b/tests/repros/repro-copper-pour-group-id.test.tsx
@@ -1,0 +1,44 @@
+import { test, expect } from "bun:test"
+import { getTestFixture } from "tests/fixtures/get-test-fixture"
+
+test("copper pour records carry the enclosing pcb_group_id", async () => {
+  const { circuit } = getTestFixture()
+  circuit.add(
+    <board width="20mm" height="20mm">
+      <group name="G1">
+        <resistor
+          name="R1"
+          resistance="1k"
+          footprint="0402"
+          pcbX={0}
+          pcbY={5}
+        />
+        <trace from=".R1 > .pin1" to="net.GND" />
+        <via
+          connectsTo="net.GND"
+          holeDiameter="0.3mm"
+          outerDiameter="0.6mm"
+          fromLayer="top"
+          toLayer="bottom"
+          pcbX={0}
+          pcbY={-8}
+        />
+        <copperpour connectsTo="net.GND" layer="top" clearance="0.3mm" />
+      </group>
+    </board>,
+  )
+  await circuit.renderUntilSettled()
+
+  const pours = circuit.db.pcb_copper_pour.list()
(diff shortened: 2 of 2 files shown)


DPS0340:
@seveibar could you review this small one? 2 files, +45/-0.

Copper pour records weren't getting `pcb_group_id` set, so pours couldn't be associated back to their group the way other pcb elements can. This sets it from the owning group during pour generation.

Regression test included; CI green and mergeable.

DPS0340:
@imrishabh18 @mohan-bee — adding you two since you've handled most of the recent merges here; @seveibar merged my last one (#2814) so tagging him too. Whoever gets to it first.

This is 2 files, +45/-0, and it's a consistency fix rather than a behavior change:

`CopperPour` was the only PCB primitive that never set `pcb_group_id` on the records it emits. `PcbVia`, `CourtyardRect`, `SilkscreenPath`, `Hole`, `Port` and `PcbNoteDimension` all pass `this.getGroup()?.pcb_group_id`.

Concretely — a `<via>` and a `<copperpour>` inside the same `<group name="G1">`:

```
before:  via group: ["pcb_group_0"]   pour group: [undefined]
after:   via group: ["pcb_group_0"]   pour group: ["pcb_group_0"]
```

One line in `doInitialPcbCopperPourRender`, plus a regression test that asserts the pour's group id equals the via's (not merely non-null).

Note this is deliberately *not* a fix for #2621 (one `pcb_copper_pour` record per fill fragment) — a board-level pour still has no group, so that one needs a circuit-json schema decision, which I've laid out as options in a comment there.

Verification: `tsc --noEmit` clean, `biome format` clean, all 7 copper-pour test files 15 pass / 0 fail. CI green, `MERGEABLE / CLEAN`.

DPS0340:
Being straight about the value of this one, because I don't want it merged on a claim it doesn't support.

**What it fixes is consistency, not observable behavior today.** I went looking for a consumer that would visibly change, and there isn't one: nothing in `core` filters or groups by `pcb_group_id` on pours, so no render, export, or layout output differs. The concrete effect is exactly what the test asserts — a pour inside a `<group>` now reports the same `pcb_group_id` its sibling `<via>` already reported, instead of `undefined`.

The argument for merging it is that `pcb_group_id` is a schema-wide field (62 element types in circuit-json carry it) and `CopperPour` was the single PCB primitive not populating it — `PcbVia`, `CourtyardRect`, `SilkscreenPath`, `Hole`, `Port`, `PcbNoteDimension` all do. So it's a gap that will bite whenever something *does* start grouping by it, and it's one line.

The argument against is that it's speculative until a consumer needs it.

I'm raising this because I watched @Hero988 close `calculate-packing#111` after measuring that his change didn't actually improve the three real boards — that's the right bar, and I'd rather apply it to my own PR than have you spend review time discovering the same thing. If "no user-visible improvement" is disqualifying here, say so and I'll close it; if the consistency argument is enough, it's ready.

@imrishabh18 @mohan-bee @seveibar

itsbalamurali:
Dont close. this bug exists

Issue #2830, opened 2026-07-25
Transformed geometry points dominate generation heap: ~4.7M retained {x,y} objects (.center/.anchorPoint) via applyToPoint


itsbalamurali:
### Summary

The dominant retained allocation during circuit-JSON generation is not schema
validation — it is transformed geometry points. A heap snapshot near the OOM
point on a mid-size board retains **~4.7 million plain `{x, y}` objects** held
via `.center` and `.anchorPoint`, even though the finished Circuit JSON contains
only a couple hundred of them.

This is a separate root cause from the Zod schema retention in #2810 (that was
~1.24M nodes; this is ~4.7M). Both surface as "generation OOMs on mid-size
boards" (tscircuit/tscircuit#4077).

### Evidence

Heap snapshot retainer analysis (plain `Object` nodes, by retaining edge):

```
7,356,228  plain Object nodes total
2,352,503  retained via  .center
2,350,861  retained via  .anchorPoint
2,421,204  (object elements) / 2,144,526 Array   ← arrays holding them
```

Same board's finished `circuit.json`:

```
"center"       : 222 occurrences
"anchorPoint"  : 0 occurrences
```

So ~4.7M point objects are transient intermediates that stay reachable for the
whole generation rather than a few hundred that reach the output.

The CLI bundle has **597 `applyToPoint` call sites** (plus 91 `applyMatrix`, 29
`transformPoint`). Each `applyToPoint` returns a fresh `{ x, y }`. With deep
component nesting and many points per footprint, and transformed copies retained
by the intermediate representation, this multiplies into millions.

### Why it matters

- It is the single largest retained-heap contributor — larger than the Zod
  schemas — so it is the main reason generation needs multiples of GB and does
  not complete on mid-size boards under Node.
- `anchorPoint` in particular is 0 in the output but ~2.35M in the heap, i.e.
  purely intermediate that is never released until the end.

### Suggested directions

1. **Don't retain transformed intermediates.** If per-element transformed
   geometry is cached/accumulated during layout, ensure it is released once the
   element's output is produced rather than held on a long-lived structure.
2. **Reduce redundant transform passes.** 597 `applyToPoint` sites suggests
   points are re-transformed at multiple nesting levels; composing matrices once
   per element and applying a single transform would cut both allocations and
   time.
3. **Mutate/reuse point objects** in the hot transform path instead of
   allocating a new `{ x, y }` per call, where the source point is not shared.

### Repro

[...]

PR #2832, opened 2026-07-25
perf: back renderPhaseStates with typed arrays instead of 63 objects per component

DPS0340:
Follow-up to #2814, and it also carries a measurement that I think redirects #2830.

## What I measured

Node's sampling heap profiler (`HeapProfiler.startSampling`, 2048-byte interval) over `Benchmark1LedMatrix` with `routingDisabled` — 4,734 circuit-json elements, 2,114 components in the tree. Same script, same board, run against `origin/main` in a clean worktree and against this branch:

| | `origin/main` | this PR |
|---|---:|---:|
| total sampled allocation | 29.46 MB | **19.49 MB** |
| `Renderable` constructor (self) | 12.41 MB | **0.30 MB** |

**-33.8% total allocation**, and the `Renderable` constructor drops from **42% of all sampled allocation to 1.8%**.

## Cause

`Renderable`'s constructor eagerly built one `{ initialized, dirty }` object per render phase:

```ts
this.renderPhaseStates = {} as RenderPhaseStates
for (const phase of orderedRenderPhases) {
  this.renderPhaseStates[phase] = { initialized: false, dirty: false }
}
```

There are 63 phases, so every component in the tree allocated 63 objects plus the containing record. On this board that is **133,182 state objects** for 2,114 components — and the count scales with `orderedRenderPhases.length`, so it grows every time a phase is added.

## Change

The flags move into two `Uint8Array`s (one byte per phase each). The hot paths — `runRenderPhase` and `_markDirty` — index those arrays directly, and `_markDirty`'s "mark this phase and everything after it" loop becomes a single `dirtyFlags.fill(1, phaseIndex)`.

`renderPhaseStates` itself is still there and still behaves identically; it is now a lazily-created view that is only built if something reads it.

## Compatibility

`renderPhaseStates` is read by `lib` (`Trace_doInitialPcbTraceRender`, `NormalComponent`), by tests, and is serialized through `getRenderGraph()`, so I pinned the observable shape rather than assuming it:

- `Object.keys(states)` is still all 63 phases **in order**
- `"SourceRender" in states`, spreading, and enumeration all work
- writes through `states[phase].dirty = true` are visible on a fresh read
- `JSON.stringify(getRenderGraph())` still emits `{"initialized":...,"dirty":...}` per phase

Those four properties are the new test file.

## About #2830

While verifying this I could not reproduce the premise of #2830 on this board. Walking the live object graph from the circuit root after render finds **1,492** retained `{x, y}` objects (1,488 of them behind `.route`), not millions, and `anchorPoint` does not appear in `core`'s retained graph at all — `core` has exactly one `anchorPoint` reference, in `applyNetLabelPlacements`, reading solver output.

By allocation, transform helpers (`applyToPoint`/`applyMatrix`/`transformPoint`/`compose`/`decomposeTSR`) account for **0.13 MB of 29.46 MB — 0.46%**. So on this board the 597 `applyToPoint` call sites are not the bottleneck; the per-phase state objects were, by two orders of magnitude. If the reporter's board behaves differently I'd like the input JSON, since that would mean the retention is coming from a path this benchmark doesn't exercise. I'll follow up on #2830 with these numbers.

## Verification

Full suite **1259 pass / 0 fail / 31 skip**, `tsc --noEmit` clean, `biome format .` clean.

**Diff: 1 source file + 1 test file.** No snapshots touched.

cc @seveibar @imrishabh18 @mohan-bee

### DIFF
--- a/lib/components/base-components/Renderable.ts
+++ b/lib/components/base-components/Renderable.ts
@@ -167,6 +167,113 @@ export type RenderPhaseStates = Record<
   }
 >
 
+/**
+ * Backing store for `renderPhaseStates`.
+ *
+ * Every Renderable used to allocate one `{ initialized, dirty }` object per
+ * render phase in its constructor. With 63 phases that is 63 objects plus the
+ * containing record for every component in the tree, which made the Renderable
+ * constructor the single largest allocation site during generation.
+ *
+ * The flags now live in two `Uint8Array`s (one byte per phase each), and the
+ * per-phase objects are materialized lazily and cached only for the phases a
+ * caller actually touches. The public shape of `renderPhaseStates` is
+ * unchanged: it still reads and writes as
+ * `renderPhaseStates[phase].initialized` / `.dirty`, still enumerates every
+ * phase, and still serializes identically in `getRenderGraph()`.
+ */
+class RenderPhaseStateStore {
+  readonly initializedFlags: Uint8Array
+  readonly dirtyFlags: Uint8Array
+  private _views: Array<RenderPhaseStateView | undefined>
+
+  constructor() {
+    this.initializedFlags = new Uint8Array(orderedRenderPhases.length)
+    this.dirtyFlags = new Uint8Array(orderedRenderPhases.length)
+    this._views = new Array(orderedRenderPhases.length)
+  }
+
+  getView(phaseIndex: number): RenderPhaseStateView {
+    let view = this._views[phaseIndex]
+    if (view === undefined) {
+      view = new RenderPhaseStateView(this, phaseIndex)
+      this._views[phaseIndex] = view
+    }
+    return view
+  }
+}
+
+/**
+ * A live view onto one phase's flags. Reads and writes go straight to the
+ * shared typed arrays, so this stays in sync with `_markDirty` and with any
+ * other view of the same phase.
+ */
+class RenderPhaseStateView {
+  constructor(
+    private readonly _store: RenderPhaseStateStore,
+    private readonly _phaseIndex: number,
+  ) {}
+
+  get initialized(): boolean {
+    return this._store.initializedFlags[this._phaseIndex] === 1
+  }
+
+  set initialized(value: boolean) {
+    this._store.initializedFlags[this._phaseIndex] = value ? 1 : 0
+  }
+
+  get dirty(): boolean {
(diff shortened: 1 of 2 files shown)


DPS0340:
@mohan-bee — one focused ask, since you're active in this repo today. Not asking you to look at everything I have open.

**#2832 is the one worth your time**: it makes circuit-JSON generation allocate **33.8% less**, and it's 2 files.

`Renderable`'s constructor eagerly allocated one `{ initialized, dirty }` object *per render phase*, for every component:

```ts
this.renderPhaseStates = {} as RenderPhaseStates
for (const phase of orderedRenderPhases) {
  this.renderPhaseStates[phase] = { initialized: false, dirty: false }
}
```

63 phases × 2,114 components on `Benchmark1LedMatrix` = **133,182 objects**, and it scales with `orderedRenderPhases.length` — it gets worse every time a phase is added.

Measured with Node's sampling heap profiler, same script against `origin/main` in a clean worktree vs the branch:

| | main | #2832 |
|---|---:|---:|
| total sampled allocation | 29.46 MB | **19.49 MB** |
| `Renderable` constructor | 12.41 MB | **0.30 MB** |

The constructor drops from **42% of all sampled allocation to 1.8%**.

This is directly relevant to #2810 and tscircuit/tscircuit#4077 (generation OOMing past 4.4 GB on mid-size boards). While measuring it I also checked the premise of #2830 — transform helpers are **0.13 MB of 29.46 MB, 0.46%**, and stay flat when I quadruple the inline geometry that hypothesis points at. So this and the Zod work are the two real contributors; the transform path isn't one.

`renderPhaseStates` is read by `lib`, by tests, and serialized through `getRenderGraph()`, so I pinned the observable shape rather than assuming it: key order, `in`/spread, write-through, and JSON output are all covered by the new tests.

Full suite **1259 pass / 0 fail**, `tsc --noEmit` clean, format clean.

If you'd rather not take a perf change from an outside contributor, that's fair — but #2831 (mfn shorthand, 4 one-line fixes + test) and #2813 (keepout layers, closes #2613) are both small and independent if either is easier to say yes to. And if none of them are wanted, tell me and I'll close them; I'd rather that than leave a queue sitting.

DPS0340:

Sorry for the noise.

itsbalamurali:
@DPS0340 please consider reopening this.

PR #2834, opened 2026-07-25
fix: omit the voltage half of the fuse label when voltageRating is absent (#2833)

### LINKED ISSUE #2833: fuse without voltageRating renders a bare unit: "1A / V" on the schematic, "V" in circuit JSON
### What happens

A `<fuse />` without `voltageRating` renders a **bare unit with no number** — on the schematic symbol and in circuit JSON:

```tsx
<fuse name="F1" currentRating="1A" footprint="0402" />
```

```
source_component.display_current_rating  "1A"
source_component.display_voltage_rating  "V"      ← no value
schematic_component.symbol_display_value "1A / V" ← rendered on the schematic
```

`voltageRating` is optional in the props schema (`voltageRating?: number | string`), while `currentRating` is required — a fuse without a `currentRating` fails validation outright (`Invalid props for fuse "F4": currentRating (Required)`). So "current only" is a fully supported, and probably common, way to declare a fuse, and it's the case that renders wrong.

### Cause

`lib/components/normal-components/Fuse.ts` interpolates the formatted value unconditionally:

```ts
display_voltage_rating: `${formatSiUnit(voltageRating)}V`,
...
return `${formatSiUnit(current)}A / ${formatSiUnit(voltage)}V`
```

`formatSiUnit(undefined)` returns `""`, so the template collapses to just the unit suffix.

### Prior art in this repo

`Resonator._getSchematicSymbolDisplayValue()` already handles exactly this shape correctly — it drops the optional half rather than emitting an empty unit:

```ts
const freqDisplay = `${formatSiUnit(this._parsedProps.frequency)}Hz`
if (this._parsedProps.loadCapacitance) {
  return `${freqDisplay} / ${formatSiUnit(this._parsedProps.loadCapacitance)}F`
}
return freqDisplay
```

### Why no test catches it

`tests/components/normal-components/fuse.test.tsx` is the only fuse component test and it passes **both** ratings, so the optional-field path is never exercised.

Interestingly the repo already contains schematic snapshots showing the bug, they just weren't recognised as wrong:

- `tests/features/schematic-text-bounding-box/schematic-text-bounding-box-fuse.test.tsx` — four fuses, all `currentRating="2A"` with no voltage, all rendering `2A / V`.
- `schematic-section-rp2040.test.tsx` and `schematic-section-rp2040-autolayout.test.tsx` — `<fuse name="F1" currentRating="500mA" />`, rendering `500mA / V`.

Happy to submit a PR (already have one ready).

DPS0340:
Closes #2833.

A fuse declared with only `currentRating` — which is the only *required* rating, so this is a fully supported way to write one — rendered a bare unit with no number:

```
before                          after
display_voltage_rating  "V"     undefined
symbol_display_value    "1A / V"  "1A"
```

### The fix

Both call sites interpolated the formatted value unconditionally, and `formatSiUnit(undefined)` returns `""`, so the template collapsed to the unit suffix. Now the optional half is dropped instead:

```ts
if (voltage === undefined || Number.isNaN(voltage)) return currentDisplay
```

This mirrors `Resonator._getSchematicSymbolDisplayValue()`, which already handles the same required/optional shape this way — I followed the existing convention rather than inventing one.

`NaN` is covered as well as `undefined` because the props accept strings: `voltageRating="abc"` goes through `parseFloat` and would otherwise have produced `"NaNpV"`.

### Verification

Fuses that *do* declare a voltage are untouched:

```
F1  currentRating="1A"                     → "1A"           (was "1A / V")
F2  currentRating="2A"  voltageRating="32V" → "2A / 32V"     unchanged
F3  currentRating={0.5} voltageRating={250} → "500mA / 250V" unchanged
```

**The new test bites.** Reverting only `Fuse.ts`:

```
expect(sourceByName.F1.display_voltage_rating).toBeUndefined()
Received: "V"
(fail) <fuse /> omits the voltage half of the label when voltageRating is absent
```

It asserts both the source fields and the schematic symbol values, and includes a two-rating fuse in the same board so a regression in either direction is caught.

**Suite:** `1257 pass / 0 fail` (3 skipped suites unrelated). `biome format` and `tsc --noEmit` clean for the touched files.

### About the 3 updated snapshots

They aren't collateral — **they're the bug, captured**. All three contain fuses with no `voltageRating` and were storing the broken label:

- `schematic-text-bounding-box-fuse` — four fuses at `currentRating="2A"`, all showing `2A / V`
- `schematic-section-rp2040` and `-autolayout` — `<fuse name="F1" currentRating="500mA" />`, showing `500mA / V`

I rendered the bounding-box snapshot to confirm the labels now read `2A` and that the text bounding boxes shrink to match, rather than the label merely being blanked.

Worth noting the existing fuse component test passes *both* ratings, which is why the optional path was never exercised.


DPS0340:

| PR | file |
|---|---|
| #2838 | `Inductor.ts` |

   ```ts
   ```

DPS0340:

Sorry for the noise.

PR #2838, opened 2026-07-25
fix: carry inductor maxCurrentRating into source_component (#2837)

### LINKED ISSUE #2837: inductor drops maxCurrentRating: prop and schema field both exist, value never written
### What happens

`maxCurrentRating` is declared on `InductorProps` and `max_current_rating` exists on the `simple_inductor` circuit-JSON schema, but `Inductor` never writes it — the value is silently dropped:

```tsx
<inductor name="L2" inductance="10uH" maxCurrentRating="2A" />
```

```json
{
  "type": "source_component",
  "ftype": "simple_inductor",
  "name": "L2",
  "inductance": "10uH",
  "display_inductance": "10µH"
}
```

No `max_current_rating` anywhere in the output.

### Both ends already exist

```ts
// @tscircuit/props — InductorProps
maxCurrentRating?: number | string;
```

```ts
// circuit-json — source_component, ftype "simple_inductor"
max_current_rating: z.ZodOptional<z.ZodNumber>;
```

So this is purely a missing hand-off in `Inductor.doInitialSourceRender()`, which inserts `inductance`, `display_inductance`, `supplier_part_numbers`, `manufacturer_part_number` and `display_name`, but not the rating. Anything downstream reading `max_current_rating` (BOM export, part selection, DRC on current limits) sees nothing, with no warning that the prop was ignored.

### A wrinkle worth flagging for whoever fixes it

`maxCurrentRating` has **no zod transform**, unlike capacitor's `maxVoltageRating`:

```
capacitor: maxVoltageRating: z.ZodOptional<z.ZodEffects<...>, number, string | number>>  ← converts to number
inductor:  maxCurrentRating: z.ZodOptional<z.ZodUnion<[z.ZodString, z.ZodNumber]>>        ← raw
```

So the raw string reaches the component and a naive `parseFloat("500mA")` yields **500**, not 0.5 — off by 1000×. `parseSiUnit` doesn't work either (it returns `NaN` for `"2A"`, since it doesn't expect a unit suffix); `parseAndConvertSiUnit(...).value` handles both:

```
"2A"    → 2
"500mA" → 0.5
"1.5A"  → 1.5
"2"     → 2
"abc"   → NaN
```

### How I found it

Cross-referenced every `*Props` interface in `@tscircuit/props` against the props each component actually reads. Other hits from the same sweep that may or may not be intentional: `Capacitor.bypassFor`/`bypassTo`/`schSize`, `Resistor.schSize`/`tolerance`, `Board.topSolderMaskColor`/`bottomSolderMaskColor`/`doubleSidedAssembly`, `Chip.internalCircuit`/`pinCompatibleVariants`. I'm only filing this one because it's the one where the destination schema field already exists, so the fix is unambiguous.

PR ready.

DPS0340:
Closes #2837.

`maxCurrentRating` is declared on `InductorProps` and `max_current_rating` exists on the `simple_inductor` schema, but `doInitialSourceRender()` never wrote it — the value was dropped with no warning.

```
                                  before      after
L1  (no rating)                   undefined   undefined
L2  maxCurrentRating="2A"         undefined   2
L3  maxCurrentRating={0.5}        undefined   0.5
L4  maxCurrentRating="500mA"      undefined   0.5
```

### The part that isn't obvious

My first attempt used `parseFloat`, copying `Fuse`'s approach, and the checker caught it: **`parseFloat("500mA")` returns 500** — a 1000× error that would have written a plainly wrong value into circuit JSON rather than leaving it absent.

The reason `Fuse` gets away with `parseFloat` is that its ratings are conventionally written without SI prefixes. `maxCurrentRating` has no zod transform either, unlike capacitor's `maxVoltageRating`:

```
capacitor: maxVoltageRating: z.ZodOptional<z.ZodEffects<..., number, string | number>>  ← converted for you
inductor:  maxCurrentRating: z.ZodOptional<z.ZodUnion<[z.ZodString, z.ZodNumber]>>       ← raw string arrives
```

I then tried `parseSiUnit`, which returns **`NaN` for `"2A"`** — it doesn't expect a unit suffix. `parseAndConvertSiUnit(...).value` is the one that handles both, which I verified directly before using it:

```
"2A" → 2    "500mA" → 0.5    "1.5A" → 1.5    "2" → 2    "abc" → NaN
```

`NaN`/`null`/`undefined` all resolve to `undefined` so a malformed rating leaves the field absent rather than poisoning it.

### Verification

**The test bites.** Reverting only `Inductor.ts`:

```
expect(ratingByName.L2).toBe(2)
Received: undefined
(fail) <inductor /> carries maxCurrentRating into source_component
```

It covers all four cases in one board — omitted, string with unit, plain number, and prefixed unit — so the 1000× bug specifically is pinned by `expect(ratingByName.L4).toBe(0.5)`.

**Suite:** `1260 pass / 0 fail`, **no snapshots changed** (this only adds a source field). `biome format` and `tsc --noEmit` clean.

### Other props from the same sweep

I cross-referenced every `*Props` interface against the props each component reads. Several others appear unread — `Capacitor.bypassFor`/`bypassTo`/`schSize`, `Resistor.schSize`/`tolerance`, `Board.topSolderMaskColor`/`bottomSolderMaskColor`, `Chip.internalCircuit`/`pinCompatibleVariants`. I deliberately didn't touch those: unlike this one they have no obvious destination field, so "fixing" them would mean inventing schema semantics. Listed in #2837 in case they're worth triaging.

### DIFF
--- a/lib/components/normal-components/Inductor.ts
+++ b/lib/components/normal-components/Inductor.ts
@@ -6,7 +6,7 @@ import {
 } from "lib/utils/constants"
 import { NormalComponent } from "../base-components/NormalComponent/NormalComponent"
 import { Port } from "../primitive-components/Port"
-import { formatSiUnit } from "format-si-unit"
+import { formatSiUnit, parseAndConvertSiUnit } from "format-si-unit"
 import type { SourceSimpleInductor } from "circuit-json"
 
 export class Inductor extends NormalComponent<
@@ -41,11 +41,28 @@ export class Inductor extends NormalComponent<
   doInitialSourceRender() {
     const { db } = this.root!
     const { _parsedProps: props } = this
+    // `maxCurrentRating` is declared on InductorProps and `max_current_rating`
+    // exists on the simple_inductor schema, but the value was never carried
+    // across. Unlike capacitor's `maxVoltageRating`, this prop has no zod
+    // transform, so the raw string arrives here. `parseAndConvertSiUnit`
+    // handles the unit suffix, so "500mA" becomes 0.5 rather than 500.
+    const rawMaxCurrentRating = props.maxCurrentRating
+    const maxCurrentRating =
+      typeof rawMaxCurrentRating === "string"
+        ? parseAndConvertSiUnit(rawMaxCurrentRating).value
+        : rawMaxCurrentRating
+
     const source_component = db.source_component.insert({
       name: this.name,
       ftype: FTYPE.simple_inductor,
       inductance: this.props.inductance,
       display_inductance: this._getSchematicSymbolDisplayValue(),
+      max_current_rating:
+        maxCurrentRating === undefined ||
+        maxCurrentRating === null ||
+        Number.isNaN(maxCurrentRating as number)
+          ? undefined
+          : maxCurrentRating,
       supplier_part_numbers: props.supplierPartNumbers,
       manufacturer_part_number: props.manufacturerPartNumber ?? props.mfn,
       are_pins_interchangeable: true,
--- a/tests/components/normal-components/inductor.test.tsx
+++ b/tests/components/normal-components/inductor.test.tsx
@@ -21,3 +21,51 @@ test("<inductor /> component", async () => {
   expect(circuit).toMatchPcbSnapshot(import.meta.path)
   expect(circuit).toMatchSchematicSnapshot(import.meta.path)
 })
+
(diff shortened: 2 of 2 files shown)


DPS0340:

| PR | file |
|---|---|
| #2838 | `Inductor.ts` |

   ```ts
   ```

DPS0340:

Sorry for the noise.

PR #2840, opened 2026-07-25
fix: carry net highlightColor into pcb_trace.highlight_color (#2839)

### LINKED ISSUE #2839: net highlightColor is ignored: pcb_trace.highlight_color is never written
### What happens

`highlightColor` is documented on `NetProps` and `highlight_color` exists on the `pcb_trace` circuit-JSON schema, but nothing carries the value across — the prop has no effect:

```tsx
<net name="VCC" highlightColor="#ff0000" />
<resistor name="R1" resistance="1k" footprint="0402" pcbX={-6} />
<resistor name="R2" resistance="1k" footprint="0402" pcbX={6} />
<trace from=".R1 > .pin2" to="net.VCC" />
<trace from=".R2 > .pin2" to="net.VCC" />
```

Every resulting `pcb_trace` has `highlight_color: undefined`. `grep -rn "highlightColor\|highlight_color" lib/` returns **nothing** in core.

### Both ends already exist

```ts
// @tscircuit/props — interface NetProps
highlightColor?: string;
```

```ts
// circuit-json — pcb_trace
highlight_color: z.ZodOptional<z.ZodString>;
```

### The awkward part

`source_net` has **no** highlight field:

```
source_net: type, source_net_id, name, member_source_group_ids, is_power, is_ground,
            is_digital_signal, is_analog_signal, is_positive_voltage_source,
            trace_width, subcircuit_id, subcircuit_connectivity_map_key
```

so the colour can't simply be persisted on the net and read back later — it has to be resolved from the `Net` *component* at the point the `pcb_trace` is inserted.

There are also several insertion sites (`Trace_doInitialPcbTraceRender`, `Trace_doInitialPcbManualTraceRender`, `Group.ts`, `Net.ts`, `PcbTrace.ts`). I traced which one actually runs for a normal autorouted board by instrumenting each: it's the **segment insert in `Group.ts`** (around the `getSourceTraceIdForRoutedTrace` call), not the ones in `Trace_*`. That site only has a `source_trace_id`, so the lookup has to go `source_trace.connected_source_net_ids` → `Net` component → `highlightColor`.

### Related

Same class as #2837 (`inductor` dropping `maxCurrentRating`) — a documented prop with an existing destination schema field that was simply never written. I built the cross-reference by parsing every `interface *Props` and checking each name's snake_case form against the circuit-JSON schema; **these two were the only hits** where the destination field exists and core references neither name. Publishing that so the sweep isn't repeated.

PR ready.

DPS0340:
Closes #2839.

`NetProps.highlightColor` is documented and `pcb_trace.highlight_color` exists on the circuit-JSON schema, but the value was never written — `grep -rn "highlightColor\|highlight_color" lib/` returned nothing in core.

```
                                        before      after
trace on <net highlightColor="#ff0000"> undefined   "#ff0000"
trace on <net /> (no prop)              undefined   undefined
```

### Finding the right insertion site mattered more than the fix

There are five `db.pcb_trace.insert` sites. My first attempt patched the ones in `Trace_doInitialPcbTraceRender` / `Trace_doInitialPcbManualTraceRender` and added a `Trace._getHighlightColor()` helper. The helper resolved correctly — I verified it returns `#ff0000` — **and the output was still `undefined`**, because for a normally autorouted board none of those sites run.

I instrumented each site with a marker to find out which one actually executes: it's the **segment insert in `Group.ts`**, after `getSourceTraceIdForRoutedTrace`. That's where this PR writes the field.

That site has only a `source_trace_id`, and `source_net` has no highlight field to read back:

```
source_net: type, source_net_id, name, member_source_group_ids, is_power, is_ground,
            is_digital_signal, is_analog_signal, is_positive_voltage_source,
            trace_width, subcircuit_id, subcircuit_connectivity_map_key
```

So `getHighlightColorForSourceTrace` resolves `source_trace.connected_source_net_ids` → matching `Net` component → `highlightColor`. It's a small standalone helper rather than inline code so the other insertion sites can adopt it if they ever become reachable.

I kept the `Trace._getHighlightColor()` helper too, since the `Trace_*` sites are reachable for cached-route and manual-path boards, and left both paths consistent.

### Verification

**The test bites.** Reverting only `Group.ts`:

```
expect(highlightByNetName.VCC).toBe("#ff0000")
Received: undefined
(fail) carries net highlightColor into pcb_trace.highlight_color
```

It also asserts a second net **without** the prop stays `undefined`, so the fix can't pass by blanket-applying a colour.

**Suite:** `1260 pass / 0 fail`, no snapshot churn, `biome format` and `tsc --noEmit` clean.

One note on process: an intermediate run of mine reported 7 unrelated failures (3D/PCB snapshot tests). Those were an artifact of my machine hitting 100% disk during the run — they pass individually and the count returns to `0 fail` on a clean run of this same branch, and `main` is also `0 fail`. Flagging it so a stray number in my earlier notes isn't mistaken for a real regression.

### How this was found

Cross-referenced every `interface *Props` in `@tscircuit/props` against the circuit-JSON schema, keeping props whose snake_case name exists as a schema field but which core references under **neither** name. Exactly two hits: `InductorProps.maxCurrentRating` (#2837 / #2838) and this one. The sweep is now exhausted for that particular pattern.

### DIFF
--- a/lib/components/primitive-components/Group/Group.ts
+++ b/lib/components/primitive-components/Group/Group.ts
@@ -96,6 +96,7 @@ import {
 } from "./region-replacement"
 import { splitPcbTracesOnJumperSegments } from "./split-pcb-traces-on-jumper-segments"
 import { computeCenterFromAnchorPosition } from "./utils/computeCenterFromAnchorPosition"
+import { getHighlightColorForSourceTrace } from "lib/utils/get-highlight-color-for-source-trace"
 
 const getDistanceToPoint = (
   routePoint: PcbTrace["route"][number],
@@ -1509,6 +1510,11 @@ export class Group<Props extends z.ZodType<any, any, any> = typeof groupProps>
             ...pcb_trace,
             source_trace_id: sourceTraceId,
             route: segment,
+            highlight_color: getHighlightColorForSourceTrace({
+              db,
+              group: this,
+              sourceTraceId,
+            }),
           })
         }
       }
--- a/lib/components/primitive-components/Trace/Trace.ts
+++ b/lib/components/primitive-components/Trace/Trace.ts
@@ -73,6 +73,21 @@ export class Trace
     return this._parsedProps.thickness ?? this._parsedProps.width
   }
 
+  /**
+   * The `highlightColor` of a net this trace is connected to, if any.
+   *
+   * `pcb_trace.highlight_color` exists on the circuit-JSON schema and
+   * `NetProps.highlightColor` is documented, but nothing carried the value
+   * across, so the prop had no effect.
+   */
+  _getHighlightColor(): string | undefined {
+    for (const net of this._findConnectedNets().nets) {
+      const highlightColor = net?._parsedProps?.highlightColor
+      if (highlightColor) return highlightColor
+    }
+    return undefined
+  }
+
   _getSchematicNetLabelText(): string | undefined {
     return (
       this._parsedProps.schDisplayLabel ??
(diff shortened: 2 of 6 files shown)


DPS0340:

| PR | file |
|---|---|
| #2838 | `Inductor.ts` |

   ```ts
   ```

DPS0340:

Sorry for the noise.

PR #2842, opened 2026-07-25
fix: write display_offset_x/y as display strings (#2841)

### LINKED ISSUE #2841: display_offset_x/y are written as numbers, so circuit-to-svg silently discards them
### What happens

`pcb_component.display_offset_x/y` and `pcb_group.display_offset_x/y` are written as **raw numbers**, but circuit-json types them as strings:

```ts
display_offset_x: z.string().optional().describe(
  "How to display the x offset for this part, usually corresponding with how the user specified it"
)
```

Validating core's own output against circuit-json:

```
pcb_component: display_offset_x: Expected string, received number
pcb_group:     display_offset_x: Expected string, received number
```

### It isn't only a validation nit — the value is silently discarded downstream

`circuit-to-svg` consumes this field for the anchor-offset dimension labels, and explicitly falls back when it isn't a string:

```js
function formatOffsetLabel(axis, offsetMm, displayOffset) {
  const valueStr = typeof displayOffset === "string" ? displayOffset : offsetMm.toFixed(2)
  ...
}
```

So today the field **never reaches the label**: because core writes a number, the renderer ignores it and recomputes the value itself. The whole point of the field — "corresponding with how the user specified it" — is unreachable.

It also loses user intent. Writing `<resistor pcbX="3mm" />` stores `3`, so `"3mm"` is gone by the time anything reads it.

### Cause

Three sites write the number directly, one of them with a cast that suppresses the type error:

```ts
lib/components/base-components/NormalComponent/NormalComponent.ts:2273
  display_offset_x: resolvedPcbX as any,
lib/components/primitive-components/Group/Group.ts:1805
  display_offset_x: props.pcbX,
lib/components/primitive-components/Group/Group_doInitialPcbComponentAnchorAlignment.ts:122
  display_offset_x: pcbX,
```

### Prior art in this repo

`Subpanel.ts` already writes the correct shape at three sites:

```ts
display_offset_x: `${child._panelPositionOffset.x}mm`,
```

so the intended format is unambiguous.

PR ready. It changes rendered labels from e.g. `X: -2.00mm` to `X: -2mm` (the user wrote `pcbX={-2}`), which is the intended behaviour of the field rather than a regression — 8 PCB snapshots update accordingly.

DPS0340:
Closes #2841.

`display_offset_x/y` are `z.string()` on circuit-json but core wrote raw numbers, so its own output failed schema validation:

```
pcb_component: display_offset_x: Expected string, received number
pcb_group:     display_offset_x: Expected string, received number
```

### Why this is a real defect, not a typing nit

I checked what actually consumes the field before writing anything. `circuit-to-svg` uses it for anchor-offset dimension labels and **falls back whenever it isn't a string**:

```js
const valueStr = typeof displayOffset === "string" ? displayOffset : offsetMm.toFixed(2)
```

So the field never reached the label — the renderer ignored core's value and recomputed its own. That makes the field's stated purpose ("corresponding with how the user specified it") unreachable, and it also drops user intent: `<resistor pcbX="3mm" />` stored `3`.

### The fix

One helper, applied at the three sites that wrote numbers (one of which used `as any` to silence the type error):

```ts
export const formatDisplayOffset = (value) => {
  if (value === undefined || value === null) return undefined
  if (typeof value === "string") return value          // "3mm" preserved verbatim
  if (!Number.isFinite(value)) return undefined
  return `${value}mm`
}
```

The `mm` shape matches `Subpanel.ts`, which already emits `` `${x}mm` `` at three sites — I followed the existing convention rather than inventing one. A user-supplied string passes through untouched, which is exactly what the field description asks for.

### About the 8 updated PCB snapshots

They're the fix taking effect, not collateral. Labels change from `X: -2.00mm` to `X: -2mm` for `pcbX={-2}` — the renderer now uses core's value instead of recomputing. I confirmed this by dumping the values before and after:

```
before  display_offset_x: -2      (number)  → label "X: -2.00mm" (recomputed)
after   display_offset_x: "-2mm"  (string)  → label "X: -2mm"    (core's value)
```

Two test files also needed updating, and I changed them deliberately rather than loosening them:

- `group-anchor-position.test.tsx` — inline snapshot `0` → `"0mm"`.
- `pcb-edge-positioning-with-calc-and-notes.test.tsx` — it did `toBeCloseTo` against the raw field; now parses the display string first. The numeric assertions are unchanged in strength.

### Verification

**The test bites.** Reverting the two component files:

```
Expected: "string"
Received: "number"
(fail) display_offset_x/y are display strings, not raw numbers
```

The new test asserts the type, that a user string (`"5mm"`) survives verbatim, **and** that every emitted `pcb_component`/`pcb_group` passes `circuit-json`'s own `safeParse` for these fields — so it pins the schema contract, not just the formatting.

### Scope note

The same schema sweep also flagged `pcb_group.anchor_alignment` being `null` (schema expects an enum or absent) and `schematic_group.subcircuit_id` being `null` (expects a string). Both are `null`-vs-`undefined` issues in different code paths, so I left them out of this PR rather than widening it — happy to file them separately if useful.


DPS0340:

| PR | file |
|---|---|
| #2838 | `Inductor.ts` |

   ```ts
   ```

DPS0340:

Sorry for the noise.

PR #2844, opened 2026-07-25
fix: make hole circuit json parse against circuit-json schemas (#2843)

### LINKED ISSUE #2843: board-level holes and DIP plated holes emit circuit JSON that fails any_circuit_element.parse()
### What happens

A board-level `<hole />` or `<platedhole />` produces circuit JSON that **fails circuit-json's own parser**:

```tsx
  <hole name="H1" pcbX={-5} pcbY={5} diameter="1mm" />
  <resistor name="R1" resistance="1k" footprint="0402" pcbX={8} />
</board>
```

```ts
any_circuit_element.safeParse(pcbHole).success  // false
```

The element carries `pcb_component_id: null`, and every id on these schemas is `z.string().optional()` — which accepts `undefined` and **rejects `null`**. Confirmed by isolating it:

```
as-is                             → false
{ ...hole, pcb_component_id: undefined } → true
omitted entirely                   → true
```

### Cause

`PrimitiveComponent` initialises its ids to `null`:

```ts
// lib/components/base-components/PrimitiveComponent/PrimitiveComponent.ts:175
pcb_component_id: string | null = null
```

and the holes resolve their owner with `??`, which only falls through on `undefined`:

```ts
const pcb_component_id =
  this.parent?.pcb_component_id ??
  this.getPrimitiveContainer()?.pcb_component_id
```

A hole inside a footprint has an owning component, so it's fine. A hole placed directly on the board has none, so `null` reaches circuit JSON.

This matters beyond validation: core itself calls `any_circuit_element.parse()` in `DrcCheck.ts`, and `@tscircuit/runframe` and `@tscircuit/capacity-autorouter` also reference `any_circuit_element`.

### A second, independent defect found alongside it

`pcb_plated_hole` with `shape: "circular_hole_with_rect_pad"` — which is what a plain `<chip footprint="dip8" />` generates — is **also invalid**, for a different reason. The schema variant requires `hole_shape` and `pad_shape`:

```ts
var pcb_circular_hole_with_rect_pad = z.object({
  shape: z.literal("circular_hole_with_rect_pad"),
  hole_shape: z.literal("circle"),
  pad_shape: z.literal("rect"),
  ...
})
```

but `PlatedHole.ts` omits both at that insertion site. The neighbouring pill variants in the same file *do* set them, so this looks like an oversight rather than intent. So today a DIP footprint alone is enough to produce unparseable circuit JSON.

### Related and deliberately not included

The same sweep also found `pcb_silkscreen_text` and `pcb_fabrication_note_text` carrying `pcb_component_id: null`. Those schemas require a **non-optional** string, so the fix there is a design question (what should a board-level silkscreen text point at?), not a null→undefined normalisation — I left them alone rather than guess.

PR ready for the two hole defects.

DPS0340:
Closes #2843.

Two independent defects that both make core emit circuit JSON which fails `any_circuit_element.parse()`.

```
                                          before   after
board-level <hole />                      invalid  valid
board-level <platedhole />                invalid  valid
<chip footprint="dip8" /> plated holes    invalid  valid
```

### 1. `pcb_component_id: null` on board-level holes

`PrimitiveComponent` initialises ids to `null` (`pcb_component_id: string | null = null`), and the holes resolve their owner with `??`, which only falls through on `undefined`. A hole inside a footprint has an owner; one placed directly on the board doesn't, so `null` reached circuit JSON — and every id on these schemas is `z.string().optional()`, which rejects `null`.

Isolated it before fixing, so the diagnosis isn't a guess:

```
as-is                                     → false
{ ...hole, pcb_component_id: undefined }  → true
omitted entirely                          → true
```

Fixed with a small `optionalId` helper at the two sites where the schema actually permits an absent id.

### 2. `circular_hole_with_rect_pad` missing `hole_shape` / `pad_shape`

Found while verifying the first fix — the test still failed, and the culprit was a *different* element. That schema variant requires both fields:

```ts
var pcb_circular_hole_with_rect_pad = z.object({
  shape: z.literal("circular_hole_with_rect_pad"),
  hole_shape: z.literal("circle"),
  pad_shape: z.literal("rect"),
  ...
})
```

`PlatedHole.ts` omitted them at that one insertion site while the neighbouring pill variants in the same file already set them, so I matched the existing pattern. A plain `<chip footprint="dip8" />` was enough to trigger this.

### Deliberately out of scope

The sweep also found `pcb_silkscreen_text` and `pcb_fabrication_note_text` carrying `pcb_component_id: null`. I tried the same normalisation there and `tsc` rejected it — those schemas require a **non-optional** string. That makes it a design question (what should a board-level silkscreen text reference?) rather than a null→undefined fix, so I reverted it and reported it in #2843 instead of guessing.

### Verification

**The test bites.** Reverting `Hole.ts` + `PlatedHole.ts`:

```
Expected: > 0
Received: 0
(fail) board-level holes produce valid circuit json
```

The test asserts both directions: board-level holes have **no** id, footprint holes **keep** theirs (`typeof === "string"`), and every hole passes `any_circuit_element.safeParse`. So it can't be satisfied by blanket-clearing the field.

One inline snapshot changed — `plated-hole-pill-rotated2` had `"pcb_component_id": null` recorded, i.e. the bug was captured in an expectation. It now reads `undefined`.

Worth noting how this was found: validating core's own output against `circuit-json`'s exported schemas across several board shapes. It's cheap and it also surfaced `pcb_group.anchor_alignment: null` and `schematic_group.subcircuit_id: null`, which I've left for separate triage.

### DIFF
--- a/lib/components/primitive-components/Hole.ts
+++ b/lib/components/primitive-components/Hole.ts
@@ -8,6 +8,7 @@ import type {
   PcbHoleRect,
   PcbHoleCircle,
 } from "circuit-json"
+import { optionalId } from "lib/utils/optional-id"
 
 export class Hole extends PrimitiveComponent<typeof holeProps> {
   pcb_hole_id: string | null = null
@@ -52,9 +53,13 @@ export class Hole extends PrimitiveComponent<typeof holeProps> {
     const position = this._getGlobalPcbPositionBeforeLayout()
     const soldermaskMargin = props.solderMaskMargin
     const isCoveredWithSolderMask = props.coveredWithSolderMask ?? false
-    const pcb_component_id =
+    // `pcb_component_id` is `string | null` on PrimitiveComponent, and `??`
+    // only falls through on `undefined` — a board-level hole has no owning
+    // component, so the `null` used to reach circuit JSON and fail validation.
+    const pcb_component_id = optionalId(
       this.parent?.pcb_component_id ??
-      this.getPrimitiveContainer()?.pcb_component_id
+        this.getPrimitiveContainer()?.pcb_component_id,
+    )
 
     this.emitSolderMaskMarginWarning(isCoveredWithSolderMask, soldermaskMargin)
 
--- a/lib/components/primitive-components/PlatedHole.ts
+++ b/lib/components/primitive-components/PlatedHole.ts
@@ -13,6 +13,7 @@ import type {
 } from "circuit-json"
 import { decomposeTSR } from "transformation-matrix"
 import { selectPortForPcbPrimitive } from "./Port/selectPortForPcbPrimitive"
+import { optionalId } from "lib/utils/optional-id"
 
 export class PlatedHole extends PrimitiveComponent<typeof platedHoleProps> {
   pcb_plated_hole_id: string | null = null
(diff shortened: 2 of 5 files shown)


DPS0340:

| PR | file |
|---|---|
| #2838 | `Inductor.ts` |

   ```ts
   ```

DPS0340:

Sorry for the noise.

PR #2846, opened 2026-07-25
fix: omit group anchor_alignment and subcircuit_id instead of writing null (#2845)

### LINKED ISSUE #2845: group emits null anchor_alignment and subcircuit_id, failing circuit-json validation
### What happens

`<group />` emits two fields as `null` where circuit-json accepts a string or nothing, so the elements fail validation:

```tsx
  <group name="G1" pcbX={4}>
  </group>
</board>
```

```
pcb_group:       anchor_alignment: Expected 'top_left' | ... , received null
schematic_group: subcircuit_id: Expected string, received null
```

### Cause

Both are explicit `null`s written at insert time:

```ts
// lib/components/primitive-components/Group/Group.ts:468

// lib/components/primitive-components/Group/Group.ts:1590
subcircuit_id: this.subcircuit_id!,   // the field is `string | null = null`
```

The schemas are:

```ts
pcb_group.anchor_alignment  →  ninePointAnchor.default("center")
schematic_group.subcircuit_id  →  z.string().optional()
```

So `undefined` is correct in both cases — and for `anchor_alignment` it's strictly better than `null`, because omitting it lets the schema's own `"center"` default apply instead of storing a value the schema rejects.

A group placed outside any subcircuit legitimately has no `subcircuit_id`; the `!` assertion just hid that it was `null` rather than a string.

### These were already baked into expectations

Three existing tests record the invalid values, which is why nothing caught it:

- `tests/groups/group-outline.test.tsx` — `"anchor_alignment": null`
- `tests/components/primitive-components/group-subcircuit-id.test.tsx` — `"anchor_alignment": null`
- `tests/groups/group-schematic-box.test.tsx` — `"subcircuit_id": null`

### Context

Third and final batch from validating core's output against circuit-json's own schemas (see #2841 for `display_offset_x/y`, #2843 for the hole elements). With this, `<group />` no longer contributes schema violations.

PR ready.

DPS0340:
Closes #2845.

`<group />` wrote `null` into two fields that circuit-json types as string-or-absent, so both elements failed validation:

```
                                    before                    after
pcb_group.anchor_alignment          null (rejected)           undefined → schema default "center"
schematic_group.subcircuit_id       null (rejected)           undefined
```

### Why `undefined` rather than a value

```ts
pcb_group.anchor_alignment     → ninePointAnchor.default("center")
schematic_group.subcircuit_id  → z.string().optional()
```

For `anchor_alignment`, omitting is strictly better than substituting `"center"` in core: the schema already supplies that default, so leaving it out keeps a single source of truth and stays correct if the default ever changes. The explicit `pcbAnchorAlignment` prop still passes through unchanged — the test asserts a `"top_left"` group survives, so this can't be satisfied by blanket-clearing the field.

For `subcircuit_id`, the `!` assertion was hiding that the field is `string | null = null` on the component. A group outside any subcircuit genuinely has none, and `undefined` is what the schema expects.

### Three existing tests recorded the invalid values

That's why nothing caught this — the bug was baked into expectations:

- `group-outline.test.tsx` — `"anchor_alignment": null`
- `group-subcircuit-id.test.tsx` — `"anchor_alignment": null`
- `group-schematic-box.test.tsx` — `"subcircuit_id": null`

All three updated to `undefined`. No assertion was weakened; they're inline snapshots of emitted values.

### Verification

**The test bites.** Reverting `Group.ts`:

```
Received: null
(fail) groups do not emit null anchor_alignment / subcircuit_id
```

It asserts neither field is `null`, that an explicitly requested alignment survives, **and** that neither field appears in `safeParse` issues — so it pins the schema contract rather than just the absence of `null`.

### Note

Locating one of the three failures took a moment because **five test files share the test name "group schematic box"** — a single-file rerun of the wrong one passes and looks like flakiness. I isolated it by running each candidate file separately.

This is the third and final batch from validating core's output against circuit-json's exported schemas (#2841 `display_offset_x/y`, #2843 hole elements). With this merged, `<group />` no longer contributes violations.

### DIFF
--- a/lib/components/primitive-components/Group/Group.ts
+++ b/lib/components/primitive-components/Group/Group.ts
@@ -465,7 +465,9 @@ export class Group<Props extends z.ZodType<any, any, any> = typeof groupProps>
             trace_clearance: props.autorouter.traceClearance,
           }
         : undefined,
-      anchor_alignment: props.pcbAnchorAlignment ?? null,
+      // circuit-json declares this as `ninePointAnchor.default("center")`, so a
+      // `null` fails validation while omitting it lets the schema default apply.
+      anchor_alignment: props.pcbAnchorAlignment ?? undefined,
     })
     this.pcb_group_id = pcb_group.pcb_group_id
 
@@ -1587,7 +1589,10 @@ export class Group<Props extends z.ZodType<any, any, any> = typeof groupProps>
     const { _parsedProps: props } = this
     const schematic_group = db.schematic_group.insert({
       is_subcircuit: this.isSubcircuit,
-      subcircuit_id: this.subcircuit_id!,
+      // `subcircuit_id` is `string | null` on the component but
+      // `z.string().optional()` on the schema, so `null` fails validation.
+      // A group outside any subcircuit simply has none.
+      subcircuit_id: this.subcircuit_id ?? undefined,
       name: this.name,
       center: this._getGlobalSchematicPositionBeforeLayout(),
       width: 0,
--- a/tests/components/primitive-components/group-circuit-json-nulls.test.tsx
+++ b/tests/components/primitive-components/group-circuit-json-nulls.test.tsx
@@ -1,0 +1,63 @@
+import { expect, test } from "bun:test"
+import * as CJ from "circuit-json"
+import { getTestFixture } from "tests/fixtures/get-test-fixture"
+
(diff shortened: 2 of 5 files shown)


DPS0340:

| PR | file |
|---|---|
| #2838 | `Inductor.ts` |

   ```ts
   ```

DPS0340:

Sorry for the noise.

PR #2849, opened 2026-07-25
fix: make getString render ids stable per circuit (#2848)

### LINKED ISSUE #2848: warning/error messages embed a process-global counter, so identical circuits produce different text
### What happens

Rendering the **same board three times in one process** produces three different warning messages:

```
<trace#14(from:.R1 > .pin1 to:net.VCC) /> is missing a name. ...
<trace#36(from:.R1 > .pin1 to:net.VCC) /> is missing a name. ...
<trace#58(from:.R1 > .pin1 to:net.VCC) /> is missing a name. ...
```

Same input, same circuit JSON everywhere else — only the identifier drifts. I found it by diffing two renders of an identical board element-by-element: `source_unnamed_trace_warning` was the only element that differed.

### Cause

`getString()` embeds `_renderId`, which comes from a module-level counter that is never reset:

```ts
// lib/components/base-components/Renderable.ts
let globalRenderCounter = 0
...
this._renderId = `${globalRenderCounter++}`
```

So the number reflects **how many components have been constructed in the process so far**, not anything about the circuit. First circuit's trace is `#14`, the second is `#36`, the third `#58`.

### Why it matters beyond aesthetics

`getString()` feeds a lot of user-facing text, not just this one warning:

```
Trace ... routed outside the board boundaries.
... has chipRef "..." without connections. chipRef will be ignored.
... has both manual placement and prop coordinates. ...
... references schSheetName "...", which does not match any schematic sheet ...
... is ambiguous: references multiple non-overlapping pads: ...
Some ports did not have a matching PCB primitive ... Missing ports: ...
Failed to route net islands for "..."
```

Any consumer that snapshots, diffs, caches or deduplicates these strings sees spurious changes whenever an unrelated circuit was rendered earlier in the same process. It also makes the message useless as a stable handle: `trace#14` doesn't identify anything a user can look up.

### Note

`_renderId` itself needs to stay process-unique — `cssSelectPrimitiveComponentAdapter` uses it for identity comparison (`a._renderId === b._renderId`). So the fix belongs in the display path, not in the counter.

PR ready.

DPS0340:
Closes #2848.

Rendering the same board three times in one process produced three different warnings:

```
before                                    after
<trace#14(from:.R1 > .pin1 to:net.VCC) />  <trace#14(...) />
<trace#36(...) />                          <trace#14(...) />
<trace#58(...) />                          <trace#14(...) />
```

`getString()` embeds `_renderId`, which is a module-level counter that never resets — so the number reflects how many components the *process* has constructed, not anything about the circuit.

### Why the fix is in the display path, not the counter

`_renderId` has to stay process-unique: `cssSelectPrimitiveComponentAdapter` compares it for identity (`a._renderId === b._renderId`). Resetting the counter per circuit would break that. So `getString()` now offsets against the lowest id in the circuit, computed once and memoised on the root:

```ts
_getCircuitRelativeRenderId(): string {
  // walk the tree once, cache the minimum on root._renderIdOrigin
  const relative = Number(this._renderId) - root._renderIdOrigin
  if (!Number.isFinite(relative) || relative < 0) return this._renderId
  return `${relative}`
}
```

It falls back to the raw id whenever the offset can't be computed (no root yet, non-numeric id, negative result), so no call site can end up worse off than today.

### Verification

**The test bites.** Reverting `PrimitiveComponent.ts`:

```
expect(new Set(messages).size).toBe(1)
Expected: 1
Received: 3
```

A second test guards the obvious way to "fix" this wrongly: **ids must stay distinct within a circuit**. Two resistors on one board still produce two different strings, so collapsing everything to a constant can't pass.

I also checked that circuits of different shapes still get different numbers (a board with an extra capacitor reports `trace#26` vs `trace#14`) — the id remains meaningful, it just stops depending on process history.

**Suite:** `1261 pass / 0 fail`, `tsc --noEmit` 0 errors, `biome format` clean. No snapshots changed.

### How this was found

Diffing two renders of an identical board element by element. Of 105 elements, `source_unnamed_trace_warning` was the only one that differed — which is what pointed at `getString()` rather than at any single component.

Worth noting the blast radius: `getString()` also feeds the "routed outside the board boundaries", "chipRef without connections", "manual placement and prop coordinates", "schSheetName does not match", "is ambiguous: references multiple non-overlapping pads", "Missing ports" and "Failed to route net islands" messages. All of them inherit the same instability today.

### DIFF
--- a/lib/components/base-components/PrimitiveComponent/PrimitiveComponent.ts
+++ b/lib/components/base-components/PrimitiveComponent/PrimitiveComponent.ts
@@ -1384,22 +1384,56 @@ export abstract class PrimitiveComponent<
 
   getString(): string {
     const { lowercaseComponentName: cname, _parsedProps: props, parent } = this
+    const id = this._getCircuitRelativeRenderId()
     if (props?.pinNumber !== undefined && parent?.props?.name && props?.name) {
-      return `<${cname}#${this._renderId}(pin:${props.pinNumber} .${parent?.props.name}>.${props.name}) />`
+      return `<${cname}#${id}(pin:${props.pinNumber} .${parent?.props.name}>.${props.name}) />`
     }
     if (parent?.props?.name && props?.name) {
-      return `<${cname}#${this._renderId}(.${parent?.props.name}>.${props?.name}) />`
+      return `<${cname}#${id}(.${parent?.props.name}>.${props?.name}) />`
     }
     if (props?.from && props?.to) {
-      return `<${cname}#${this._renderId}(from:${props.from} to:${props?.to}) />`
+      return `<${cname}#${id}(from:${props.from} to:${props?.to}) />`
     }
     if (props?.name) {
-      return `<${cname}#${this._renderId} name=".${props?.name}" />`
+      return `<${cname}#${id} name=".${props?.name}" />`
     }
     if (props?.portHints) {
-      return `<${cname}#${this._renderId}(${props.portHints.map((ph: string) => `.${ph}`).join(", ")}) />`
+      return `<${cname}#${id}(${props.portHints.map((ph: string) => `.${ph}`).join(", ")}) />`
     }
-    return `<${cname}#${this._renderId} />`
+    return `<${cname}#${id} />`
+  }
+
+  /**
+   * A render id that is stable for a given circuit.
+   *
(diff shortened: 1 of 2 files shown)


DPS0340:

Sorry for the noise.

PR #2851, opened 2026-07-25
fix: strip the child combinator from trace error parent selectors (#2850)

### LINKED ISSUE #2850: trace pin-not-found error names the wrong element and falsely reports "It has no ports"
### What happens

When a trace references a pin that doesn't exist, the error message names the **wrong element** and reports **wrong facts about it**:

```tsx
<chip name="U1" footprint="soic8" pinLabels={{ pin1: ["VCC"], pin2: ["GND"] }} />
<trace from=".U1 > .VCC" to=".U1 > .NOPE" />
```

```
Could not find port for selector ".U1 > .NOPE".
Component ".U1 > " found, but does not have pin "NOPE". It has no ports
```

Two things are wrong: the component is printed as `".U1 > "`, and `U1` is a soic8 with **8 ports** — "It has no ports" is simply false.

Same for a missing component:

```
Could not find port for selector ".R9 > .pin1". Component ".R9 > " not found
```

### Cause

`Trace__findConnectedPorts.ts` splits the selector into a parent and a port token, and **both branches leave the child combinator attached**:

```ts
const dotIndex = selector.lastIndexOf(".")
if (dotIndex !== -1 && dotIndex > selector.lastIndexOf(" ")) {
  parentSelector = selector.slice(0, dotIndex)      // ".U1 > "
  ...
} else {
  const match = selector.match(/^(.*[ >])?([^ >]+)$/)
  parentSelector = match?.[1]?.trim() ?? ""         // ".U1 >"
  ...
}
```

`".U1 > "` isn't a selector for the chip — it's a descendant selector with an empty right-hand side. I checked what it actually resolves to:

```
selectOne(".U1")   → Chip  "U1", 8 ports
selectOne(".U1 >") → SmtPad (unnamed), 0 ports
```

So the code finds *an arbitrary descendant*, then reports that descendant's name and port count. That's where "It has no ports" comes from — and why the pin list a user needs (`[VCC, pin1, 1, GND, pin2, 2, ...]`) never appears.

The message is the whole point of this code path: it exists to tell the user which pins *are* available. Today it does the opposite.

### The bug is baked into an existing expectation

`tests/components/normal-components/chip-connections-invalid.test.tsx` records:

```
"Component \".unnamed_chip1 > \" not found"
```

which is why nothing caught it.

PR ready.

DPS0340:
Closes #2850.

The "pin not found" error existed to tell users which pins *are* available. It was doing the opposite:

```
before  Component ".U1 > " found, but does not have pin "NOPE". It has no ports
after   Component "U1" found, but does not have pin "NOPE".
        It has [VCC, pin1, 1, GND, pin2, 2, pin3, 3, pin4, 4, pin5, 5, pin6, 6, pin7, 7, pin8, 8]

before  Component ".R9 > " not found
after   Component ".R9" not found
```

`U1` is a soic8 with 8 ports, so "It has no ports" was flatly wrong.

### Root cause

Both selector-splitting branches leave the child combinator on `parentSelector` (`".U1 > "`). That isn't a selector for the chip — it's a descendant selector with an empty right side. I checked what it resolves to rather than assuming:

```
selectOne(".U1")   → Chip  "U1", 8 ports
selectOne(".U1 >") → SmtPad (unnamed), 0 ports
```

So the code was finding an arbitrary descendant and then reporting *that* element's name and port count. One line fixes both branches:

```ts
parentSelector = parentSelector.replace(/[\s>]+$/, "")
```

I trimmed only trailing combinator characters rather than rewriting the two branches, so nothing about which selector forms are accepted changes — only the string used for lookup and display.

### Verification

**Both tests bite.** Reverting the one-line change:

```
Expected to contain: "Component \"U1\" found"
Received: ... Component ".U1 > " found, but does not have pin "NOPE". It has no ports

Expected to contain: "Component \".R9\" not found"
Received: ... Component ".R9 > " not found
```

They assert the positive (correct name, real pin list) **and** the negative (no `".U1 > "`, no false "It has no ports"), so a cosmetic trim that still resolved the wrong element wouldn't pass.

**One existing expectation changed** — `chip-connections-invalid.test.tsx` recorded `"Component \".unnamed_chip1 > \" not found"`, i.e. the bug was captured in an inline snapshot. That's why nothing caught it. Updated to the correct string; no assertion weakened.

**Suite:** `1261 pass / 0 fail`, `tsc --noEmit` 0 errors, `biome format` clean.

### How this was found

Feeding deliberately broken inputs (a dangling component reference and a nonexistent pin) and reading the resulting messages, rather than only checking that *an* error is raised. The error type and count were both correct — only the text was wrong, which is exactly the part a user reads.

### DIFF
--- a/lib/components/primitive-components/Trace/Trace__findConnectedPorts.ts
+++ b/lib/components/primitive-components/Trace/Trace__findConnectedPorts.ts
@@ -57,6 +57,11 @@ export function Trace__findConnectedPorts(trace: Trace):
         parentSelector = match?.[1]?.trim() ?? ""
         portToken = match?.[2] ?? selector
       }
+      // Both branches can leave the child combinator behind: ".U1 > .pin1"
+      // yields ".U1 > ". That string doesn't select the component — it
+      // resolves to an arbitrary descendant — so the error message ends up
+      // naming the wrong element and reporting "It has no ports".
+      parentSelector = parentSelector.replace(/[\s>]+$/, "")
       let targetComponent = parentSelector
         ? trace.getSubcircuit().selectOne(parentSelector)
         : null
--- a/tests/components/normal-components/chip-connections-invalid.test.tsx
+++ b/tests/components/normal-components/chip-connections-invalid.test.tsx
@@ -38,7 +38,7 @@ test("Chip not having name messes up the connections, uses the pin of the first 
     [
       {
         "error_type": "source_trace_not_connected_error",
-        "message": "Could not find port for selector ".unnamed_chip1 > .pin1". Component ".unnamed_chip1 > " not found",
+        "message": "Could not find port for selector ".unnamed_chip1 > .pin1". Component ".unnamed_chip1" not found",
         "selectors_not_found": [
           ".unnamed_chip1 > .pin1",
         ],
--- a/tests/components/primitive-components/trace-missing-port-error-message.test.tsx
+++ b/tests/components/primitive-components/trace-missing-port-error-message.test.tsx
@@ -1,0 +1,57 @@
+import { expect, test } from "bun:test"
+import { getTestFixture } from "tests/fixtures/get-test-fixture"
+
+test("missing-pin error names the component and lists its pins", async () => {
+  const { circuit } = getTestFixture()
+
+  circuit.add(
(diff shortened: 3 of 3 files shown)


DPS0340:

Sorry for the noise.

PR #2853, opened 2026-07-25
fix: explain groups have no pins and suggest a child selector (#2852)

### LINKED ISSUE #2852: trace error on a group says "It has no ports" instead of pointing at the components inside it
### What happens

Pointing a trace at a `<group />` produces an error that is technically true but tells the user nothing:

```tsx
<group name="G1">
  <capacitor name="C1" capacitance="1uF" footprint="0402" pcbX={3} />
</group>
<trace from=".G1 > .pin1" to=".R2 > .pin1" />
```

```
Could not find port for selector ".G1 > .pin1".
Component "G1" found, but does not have pin "pin1". It has no ports
```

A group is a container — it never has pins. "It has no ports" reads like the group is broken, when the real answer is "select a component inside it". The user is left to guess which components exist and how to reference them.

Note the message above is what you get **with #2851 applied**; on current `main` it's worse, because the parent selector still carries the combinator and the code resolves `".G1 >"` to the group's first child:

```
selectOne(".G1")   → Group
selectOne(".G1 >") → Resistor "R1"
```

so `main` reports `Component "R1" found, but does not have pin "pin1"` — and `R1` **does** have `pin1`, making the message actively misleading. #2851 fixes the naming; this issue is about the remaining unhelpful detail.

### Suggested detail

The component's own children are right there, so the message can name them and give a working selector:

```
Component "G1" found, but does not have pin "pin1". It is a group, which has no
pins of its own. Select a component inside it, e.g. ".R1 > .pin1". It contains [.R1, .C1]
```

### Existing expectations

Two tests record the old wording, both with **empty** groups, so the phrasing there becomes "…and it contains no named components":

- `tests/components/base-components/trace-port-selector-errors.test.tsx` (`<group name="G1" />`)
- `tests/repros/repro99-trace-selector-pcbpath-no-throw.test.tsx` (`<group name="J1" />`)

PR ready (stacked on #2851).

DPS0340:
Closes #2852. **Stacked on #2851** — please merge that one first; this branch contains its commit, so GitHub will show the diff correctly once #2851 lands.

```
before  Component "G1" found, but does not have pin "pin1". It has no ports
after   Component "G1" found, but does not have pin "pin1". It is a group, which has
        no pins of its own. Select a component inside it, e.g. ".R1 > .pin1".
        It contains [.R1, .C1]
```

A group is a container, so "It has no ports" is true but reads like the group is broken. The children are already in hand at that point, so the message can name them and hand back a selector that actually works — reusing the pin the user asked for (`.pin1`), not a generic example.

### Scope

Only the `ports.length === 0` branch, and only when the component is a `Group`/`Subcircuit`. A real component with no ports still gets the original wording — there's a test for that, so this can't be satisfied by changing the string unconditionally.

### Verification

**The test bites.** Reverting the change:

```
Expected to not contain: "It has no ports"
Received: ... Component "G1" found, but does not have pin "pin1". It has no ports
```

The tests assert positive and negative on both paths: a group must say "It is a group", list `.R1`/`.C1`, and suggest `".R1 > .pin1"`; a chip must **not** say "It is a group" and must still report its pins.

**Two existing expectations updated.** Both use *empty* groups, so they land on the other phrasing:

- `trace-port-selector-errors.test.tsx` — `<group name="G1" />`
- `repro99-trace-selector-pcbpath-no-throw.test.tsx` — `<group name="J1" />`

I checked each was genuinely empty before changing it rather than editing to make tests pass; the new text ("…and it contains no named components") is accurate for both.

**Suite:** `1263 pass / 0 fail`, `tsc --noEmit` 0 errors, `biome format` clean.

### Why this needs #2851 first

On current `main` the parent selector still carries the combinator, and `".G1 >"` resolves to the group's **first child**:

```
selectOne(".G1")   → Group
selectOne(".G1 >") → Resistor "R1"
```

so `main` reports `Component "R1" found, but does not have pin "pin1"` — and `R1` does have `pin1`. #2851 fixes which element is named; this PR fixes what the message then says about it.

### DIFF
--- a/lib/components/primitive-components/Trace/Trace__findConnectedPorts.ts
+++ b/lib/components/primitive-components/Trace/Trace__findConnectedPorts.ts
@@ -100,7 +100,22 @@ export function Trace__findConnectedPorts(trace: Trace):
       const labelList = Array.from(new Set(portNames)).join(", ")
       let detail: string
       if (ports.length === 0) {
-        detail = "It has no ports"
+        // A group is a container, not a component with pins — telling the user
+        // it "has no ports" is true but useless. Point at what they can
+        // actually connect to instead.
+        const namedChildren = targetComponent.children
+          .filter((c) => (c as any).props?.name)
+          .map((c) => `.${(c as any).props.name}`)
+        if (
+          targetComponent.componentName === "Group" ||
+          targetComponent.componentName === "Subcircuit"
+        ) {
+          detail = namedChildren.length
+            ? `It is a group, which has no pins of its own. Select a component inside it, e.g. "${namedChildren[0]} > .${portLabel}". It contains [${Array.from(new Set(namedChildren)).join(", ")}]`
+            : "It is a group, which has no pins of its own, and it contains no named components"
+        } else {
+          detail = "It has no ports"
+        }
       } else if (!hasCustomLabels) {
         detail = `It has ${ports.length} pins and no pinLabels (consider adding pinLabels)`
       } else {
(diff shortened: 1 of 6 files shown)


DPS0340:

Sorry for the noise.

PR #2856, opened 2026-07-25
fix: report unparseable unit values instead of emitting NaN (#2855)

### LINKED ISSUE #2855: unparseable unit values (resistance="abc") raise no error and render as "NaNpΩ"
### What happens

A typo in a unit-bearing prop produces **no error at all** and renders garbage on the schematic:

```tsx
<resistor name="R1" resistance="abc" footprint="0402" />
<capacitor name="C1" capacitance="xyz" footprint="0402" />
<inductor name="L1" inductance="???" footprint="0402" />
<fuse name="F1" currentRating="bad" footprint="0402" />
```

```
source_component.resistance          NaN
source_component.display_resistance  "NaNpΩ"
schematic_component.symbol_display_value  "NaNpΩ"   ← drawn on the schematic
errors raised                        0
```

Same for the others: `NaNpH`, `NaNpA`, `"NaNpA / V"`.

### Cause

These props are declared as a zod **transform** that accepts `string | number`:

```ts
resistance: z.ZodEffects<z.ZodUnion<[z.ZodString, z.ZodNumber]>, number, string | number>
```

so validation succeeds and yields `NaN`:

```ts
resistorProps.safeParse({ name: "R1", resistance: "abc" })
// → { success: true, data: { resistance: NaN } }
```

Nothing downstream checks the parsed value, so the `NaN` flows into circuit JSON, then into `formatSiUnit`, which renders it as `"NaNpΩ"`.

The contrast is stark: **omitting** `resistance` gives a clear error (`resistance (Required)`), while **misspelling** it silently produces a component with no usable value. The failure mode most likely to happen in practice is the one with no diagnostic.

### Suggested behaviour

core already has `source_invalid_component_property_error` and uses it for bad `calc()` expressions. The same mechanism fits here:

```
Invalid resistance value for Resistor "R1": "abc" could not be parsed as a number.
```

I'd keep it as an error element rather than a throw, matching how the existing invalid-property path behaves — a single bad value shouldn't take down the whole render.

PR ready.

DPS0340:
Closes #2855.

A typo in a unit-bearing prop used to produce **zero errors** and draw garbage on the schematic:

```
before  resistance="abc"   → NaN, display "NaNpΩ", symbol "NaNpΩ", 0 errors
after   resistance="abc"   → same NaN in the value, plus
        Invalid resistance value for Resistor "R1": "abc" could not be parsed as a number.
```

These props are zod **transforms** accepting `string | number`, so validation succeeds and hands back `NaN`:

```ts
resistorProps.safeParse({ name: "R1", resistance: "abc" })
// → { success: true, data: { resistance: NaN } }
```

Worth noting the asymmetry that makes this bite: **omitting** `resistance` gives a clear `resistance (Required)` error, while **misspelling** it fails silently. The more likely mistake is the one with no diagnostic.

### Approach

Rather than adding a check to each component, this inspects the emitted `source_component` for `NaN` numbers once, in `doInitialSourceParentAttachment`. That covers every ftype — the sweep found `resistance`, `capacitance`, `inductance`, `currentRating` — and any future one, without touching the individual `doInitialSourceRender` implementations.

It reuses the existing `source_invalid_component_property_error` element and `_reportInvalidComponentPropertyError` helper (already used for bad `calc()` expressions), so this introduces no new error type and no new reporting path. The helper changed from `private` to `protected` for the subclass call; that's the only signature change.

The message names the **prop the user wrote**, not the circuit-JSON field — there's a small map for the cases where they differ (`current_rating_amps` → `currentRating`, `max_voltage_rating` → `maxVoltageRating`, …), with camelCase conversion as the fallback.

### Deliberately not a throw

I kept it as an error element rather than throwing, matching how the existing invalid-property path behaves. One bad value shouldn't abort the whole render — the user still gets a board plus a precise message.

### Verification

**The test bites.** Reverting `NormalComponent.ts` drops the expectation from 3 property errors to 0.

Both directions are asserted:

- three bad values → exactly three errors, keyed by `resistance` / `capacitance` / `currentRating`, each naming the offending component and the raw string;
- a valid `<resistor resistance="1k" />` in the *same* board raises nothing, and a second test with only valid values expects `[]`.

So a blanket "always warn" implementation can't pass.

**Suite:** `1261 pass / 0 fail`, `tsc --noEmit` 0 errors, `biome format` clean, no snapshots changed.

### DIFF
--- a/lib/components/base-components/NormalComponent/NormalComponent.ts
+++ b/lib/components/base-components/NormalComponent/NormalComponent.ts
@@ -125,6 +125,19 @@ export type PortMap<T extends string> = {
  * }
  */
 
+/**
+ * circuit-JSON field name -> the prop the user wrote, where the two differ by
+ * more than snake_case/camelCase.
+ */
+const NAN_CHECKED_SOURCE_FIELD_TO_PROP: Record<string, string> = {
+  current_rating_amps: "currentRating",
+  voltage_rating_volts: "voltageRating",
+  max_voltage_rating: "maxVoltageRating",
+  max_current_rating: "maxCurrentRating",
+  load_capacitance: "loadCapacitance",
+  max_resistance: "maxResistance",
+}
+
 export class NormalComponent<
     ZodProps extends z.ZodType = any,
     PortNames extends string = never,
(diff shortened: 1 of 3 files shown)


DPS0340:

Sorry for the noise.

PR #2858, opened 2026-07-25
fix: report non-positive physical dimensions (#2857)

### LINKED ISSUE #2857: negative and zero dimensions (hole diameter, via, board size) are accepted with no error
### What happens

Negative and zero physical dimensions are accepted silently — **no error, no warning** — and flow straight into circuit JSON:

```tsx
<hole name="H1" diameter="-2mm" />
<via name="V1" holeDiameter="-0.3mm" outerDiameter="0.6mm" />
<platedhole name="PH1" shape="circle" holeDiameter="-0.5mm" outerDiameter="0.9mm" />
<board width="0mm" height="0mm" />
```

```
pcb_hole.hole_diameter        -2     errors: 0
pcb_via.hole_diameter         -0.3   errors: 0
pcb_plated_hole.hole_diameter -0.5   errors: 0
pcb_board.width                0     errors: 0
```

A negative-width `<smtpad>` inside a custom footprint behaves the same way (`width: -1`, no error).

### Why it matters

These are unmanufacturable geometries. The user gets a board that renders (a hole with negative diameter simply doesn't appear, a zero-size board looks empty) and no indication that anything is wrong — so the mistake is only discovered downstream, or at fabrication.

It also produces circuit JSON that consumers can't reason about: a drill of `-2mm` has no meaning in any exporter.

### Suggested behaviour

core already has `pcb_placement_error` and a natural place to check — `Board_doInitialPcbPlacementDesignRuleChecks` already walks the whole subcircuit's circuit JSON for the other placement checks. A dimension check fits there:

```
pcb_hole "pcb_hole_0" has hole_diameter=-2mm, which must be greater than zero.
```

### Related but not the same

I also checked `holeDiameter` larger than `outerDiameter` (a hole bigger than its own pad) — also silently accepted. That one needs a judgement call about whether it's ever legitimate, so I've left it out; happy to file separately if it's worth catching.

PR ready.

DPS0340:
Closes #2857.

Negative and zero physical dimensions were accepted with **zero diagnostics**:

```
before                                          after
pcb_hole.hole_diameter        -2    errors: 0   pcb_hole "..." has hole_diameter=-2mm, which must be greater than zero.
pcb_via.hole_diameter         -0.3  errors: 0   (same)
pcb_plated_hole.hole_diameter -0.5  errors: 0   (same)
pcb_board.width                0    errors: 0   (same)
```

These are unmanufacturable geometries. A negative-diameter hole simply doesn't render, a zero-size board looks empty, and nothing tells the user why — the mistake surfaces downstream or at fabrication.

### Where the check lives

`Board_doInitialPcbPlacementDesignRuleChecks` already walks the whole subcircuit's circuit JSON for the other placement checks, so the dimension scan piggybacks on that single pass rather than being added to each primitive's insert site (there are many — `Hole.ts` alone has four).

It reuses the existing `pcb_placement_error` element, so no new error type is introduced.

The field list is explicit per element type (`pcb_board`, `pcb_hole`, `pcb_plated_hole`, `pcb_via`, `pcb_smtpad`) rather than "any numeric field", so coordinates, rotations and offsets — which are legitimately negative — can't be caught by accident.

### Verification

**The test bites.** Reverting the check:

```
Expected: 2   Received: 0    (non-positive dimensions reported)
Expected: true Received: false (zero-size board reported)
```

Three tests cover both directions:

1. a board with a bad hole and a bad via reports exactly 2 errors, and the valid resistor in the **same board** is not flagged (asserted explicitly — `not.toContain("pcb_smtpad")`);
2. a zero-size board is reported;
3. an all-valid board with a real hole, via and resistor produces **no** dimension errors.

So a check that fires on everything, or one that only handles the negative case, can't pass.

**Suite:** `1262 pass / 0 fail`, `tsc --noEmit` 0 errors, `biome format` clean, no snapshots changed. Notably the existing suite raised **no new failures**, which is the evidence that nothing in the repo legitimately relies on a non-positive dimension.

### Deliberately out of scope

`holeDiameter` larger than `outerDiameter` (a hole bigger than its own pad) is also accepted silently, but whether that's ever legitimate is a judgement call rather than a clear invariant, so I left it out and noted it in #2857.

### DIFF
--- a/lib/components/normal-components/Board_doInitialPcbPlacementDesignRuleChecks.ts
+++ b/lib/components/normal-components/Board_doInitialPcbPlacementDesignRuleChecks.ts
@@ -9,6 +9,60 @@ import {
 import type { AnyCircuitElement } from "circuit-json"
 import type { Board } from "./Board"
 
+/**
+ * Physical dimensions that must be greater than zero, per element type.
+ *
+ * A negative or zero size is silently accepted today: `<hole diameter="-2mm" />`
+ * produces `hole_diameter: -2` with no error, and a `<board width="0mm" />`
+ * renders an empty board. Nothing downstream rejects it, so the mistake only
+ * surfaces as a confusing render or an unmanufacturable export.
+ */
+const POSITIVE_DIMENSION_FIELDS: Record<string, string[]> = {
+  pcb_board: ["width", "height"],
+  pcb_hole: ["hole_diameter", "hole_width", "hole_height"],
+  pcb_plated_hole: [
+    "hole_diameter",
+    "outer_diameter",
+    "hole_width",
+    "hole_height",
+    "outer_width",
+    "outer_height",
+    "rect_pad_width",
+    "rect_pad_height",
+  ],
+  pcb_via: ["hole_diameter", "outer_diameter"],
+  pcb_smtpad: ["width", "height", "radius"],
+}
+
+const reportNonPositiveDimensions = ({
+  board,
+  elements,
+}: {
+  board: Board
+  elements: AnyCircuitElement[]
+}) => {
+  const { db } = board.root!
+
+  for (const element of elements as any[]) {
+    const fields = POSITIVE_DIMENSION_FIELDS[element.type]
+    if (!fields) continue
+
+    for (const field of fields) {
+      const value = element[field]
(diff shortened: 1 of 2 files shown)


DPS0340:

Sorry for the noise.

PR #2860, opened 2026-07-25
fix: report traces that connect a port to itself (#2859)

### LINKED ISSUE #2859: a trace connecting a port to itself is accepted silently and emits a duplicate-port source_trace
### What happens

A trace whose two ends resolve to the **same port** is accepted silently:

```tsx
<trace from=".R1 > .pin1" to=".R1 > .pin1" />
```

```
source_trace.connected_source_port_ids  ["source_port_0", "source_port_0"]
errors                                   0
```

The trace connects nothing, but it still produces a `source_trace` element listing the same port twice. Downstream that's a connection record that can never be routed and can't mean anything electrically.

The only feedback the user gets is unrelated: `source_pin_missing_trace_warning` for the *other* pin — which actively points away from the real mistake.

### Why it matters

`from=".R1 > .pin1" to=".R1 > .pin1"` is a copy-paste slip — you edit one selector and forget the other. It looks connected in the source, renders without complaint, and the missing connection only shows up much later.

Compare with the neighbouring cases, which are all diagnosed properly:

- unknown component → `Could not find port for selector ...`
- unknown pin → `... does not have pin "NOPE"`
- unnamed trace → `... is missing a name`

A self-connection is the one malformed trace with no message.

### Suggested behaviour

Reuse `source_trace_not_connected_error`, which already exists for traces that can't be connected:

```
<trace#27(from:.R1 > .pin1 to:.R1 > .pin1) /> connects a port to itself;
both ends resolve to the same port. Did you mean to connect two different pins?
```

### Cases deliberately NOT flagged

I checked several neighbours and left them alone because they can be legitimate:

- `<trace from=".R1 > .pin1" to="net.VCC" />` — resolves to one port plus a net; normal.
- both pins of a part joined to the same net — unusual but a real circuit (e.g. a jumper/link), not obviously an error.
- duplicate identical traces — already deduplicated via the existing connection-hash lookup.

PR ready.

DPS0340:
Closes #2859.

```
before  <trace from=".R1 > .pin1" to=".R1 > .pin1" />
        → source_trace with connected_source_port_ids ["source_port_0", "source_port_0"], 0 errors

after   → <trace#27(from:.R1 > .pin1 to:.R1 > .pin1) /> connects a port to itself;
          both ends resolve to the same port. Did you mean to connect two different pins?
```

The trace connects nothing but still emitted a `source_trace` listing the same port twice — a connection record that can never be routed. The only feedback was a `source_pin_missing_trace_warning` about the *other* pin, which points away from the actual mistake.

### Scope — what is deliberately still allowed

The guard requires **≥2 resolved ports, all identical, and no connected nets**. That last condition matters: `<trace from=".R1 > .pin1" to="net.VCC" />` also resolves to a single port, and flagging it would break ordinary net connections. There's a test for exactly that.

I checked the neighbouring cases and left them alone because they can be legitimate:

- a port joined to a net (above);
- both pins of a part on the same net — unusual but a real circuit (a link/jumper), not obviously wrong;
- duplicate identical traces — already deduplicated by the existing connection-hash lookup, so no new handling needed.

Reuses `source_trace_not_connected_error`, which already exists for traces that can't be connected, so no new error type.

### Verification

**The test bites.** Reverting `Trace.ts`:

```
Expected: 1   Received: 0
(fail) a trace whose ends resolve to the same port is reported
```

Both directions are asserted:

1. a board with one self-trace **and two valid traces** → exactly 1 error, the two real `source_trace` elements survive, and no trace lists a duplicated port;
2. a board where both pins join the same net, plus port→net traces → **zero** errors.

So a guard that fires on any single-port resolution can't pass.

**Suite:** `1261 pass / 0 fail`, `tsc --noEmit` 0 errors, `biome format` clean, no snapshots changed — evidence that no existing test relies on a self-connecting trace being accepted.

### DIFF
--- a/lib/components/primitive-components/Trace/Trace.ts
+++ b/lib/components/primitive-components/Trace/Trace.ts
@@ -268,6 +268,28 @@ export class Trace
 
     if (!allPortsFound) return
 
+    // A trace whose endpoints resolve to the same port connects nothing. It is
+    // almost always a copy-paste slip (`from=".R1 > .pin1" to=".R1 > .pin1"`),
+    // and today it silently produces a source_trace listing the same port
+    // twice with no diagnostic at all.
+    const resolvedPortIds = ports.map((p) => p.port.source_port_id)
+    if (
+      resolvedPortIds.length >= 2 &&
+      new Set(resolvedPortIds).size === 1 &&
+      this._findConnectedNets().nets.length === 0
+    ) {
+      const subcircuit = this.getSubcircuit()
+      db.source_trace_not_connected_error.insert({
+        error_type: "source_trace_not_connected_error",
+        message: `${this.getString()} connects a port to itself; both ends resolve to the same port. Did you mean to connect two different pins?`,
+        subcircuit_id: subcircuit.subcircuit_id ?? undefined,
+        source_group_id: subcircuit.getGroup()?.source_group_id ?? undefined,
+        selectors_not_found: [],
+      } as any)
+      this._couldNotFindPort = true
+      return
+    }
+
     this._traceConnectionHash = this._computeTraceConnectionHash()
 
     const existingTraces = db.source_trace.list()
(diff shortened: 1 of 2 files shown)


DPS0340:

Sorry for the noise.

PR #2862, opened 2026-07-25
fix: report non-positive schematic dimensions (#2861)

### LINKED ISSUE #2861: negative schematic dimensions (schWidth, schematicbox width, circle radius) are accepted with no error
### What happens

Negative schematic dimensions are accepted with **no error**, mirroring the PCB-side issue in #2857:

```tsx
<chip name="U1" footprint="soic8" schWidth={-4} schHeight={3} />
<schematicbox width={-2} height={2} />
<schematicrect width={-3} height={1} />
<schematiccircle center={{ x: 0, y: -8 }} radius={-1} />
```

```
schematic_component.size   { width: -4, height: 3 }   errors: 0
schematic_box.width        -2                          errors: 0
schematic_rect.width       -3                          errors: 0
schematic_circle.radius    -1                          errors: 0
```

All four also pass `any_circuit_element.safeParse` — the schemas don't constrain sign — so nothing downstream rejects them either.

### Why it matters

The symbol either renders inside-out or vanishes. There's no message pointing at the prop, so a typo like `schWidth={-4}` looks like a rendering bug rather than a bad input.

Note the contrast with the schematic props that *are* handled well: `schRotation={90}` correctly produces `boxresistor_up`, and `schWidth={5}` correctly resizes. Only the invalid case is silent.

### Suggested behaviour

Same shape as the PCB fix: reuse `source_invalid_component_property_error` (already used for bad `calc()` expressions and, in #2856, unparseable unit values) and report against the prop the user wrote:

```
Invalid schWidth for schematic_component "schematic_component_0": -4, which must be greater than zero.
```

### Implementation note for whoever picks this up

`schematic_box` does **not** carry a `subcircuit_id`, so a `db.subtree({ subcircuit_id })` query silently skips it — I hit exactly that while checking this and it looked like the box was fine when it wasn't. The scan has to walk the full element list.

PR ready.

DPS0340:
Closes #2861. Schematic-side counterpart to #2858 (PCB dimensions); the two touch different files and merge in any order.

```
before                                              after
schematic_component.size { width: -4 }  errors: 0   Invalid schWidth for schematic_component "...": -4, which must be greater than zero.
schematic_box.width      -2             errors: 0   Invalid width for schematic_box "..." ...
schematic_circle.radius  -1             errors: 0   Invalid radius for schematic_circle "..." ...
```

All four element types also pass `any_circuit_element.safeParse` — the schemas don't constrain sign — so nothing downstream catches them either. The symbol renders inside-out or vanishes, and a typo like `schWidth={-4}` looks like a rendering bug rather than a bad input.

### A trap worth flagging

**`schematic_box` does not carry a `subcircuit_id`.** My first implementation used `db.subtree({ subcircuit_id })`, matching the PCB check, and the box was silently skipped — the test showed 2 errors instead of 3 and the box looked fine when it wasn't. I traced it by dumping `subcircuit_id` per element:

```
schematic_box    → undefined
schematic_circle → "subcircuit_source_group_0"
```

The scan now walks the full element list. Worth knowing for any future schematic-wide check.

### Scope

Explicit per-type field list (`schematic_box`, `schematic_rect`, `schematic_circle`, `schematic_text`, plus the nested `size` on `schematic_component`) rather than "any numeric field" — schematic coordinates and rotations are legitimately negative and must not be flagged.

Reuses `source_invalid_component_property_error`, already used for bad `calc()` expressions and (in #2856) unparseable unit values, so no new error type. The message names the **prop the user wrote** (`schWidth`, not `size.width`).

### Verification

**The test bites.** Reverting the `Board.ts` hook:

```
Expected: 3   Received: 0
(fail) non-positive schematic dimensions are reported
```

Both directions asserted: three bad dimensions → exactly 3 errors naming `schWidth` / `schematic_box` / `schematic_circle`, with the valid resistor in the **same board** explicitly not flagged (`not.toContain("R1")`); and an all-valid board expects `[]`.

**Suite:** `1261 pass / 0 fail`, `tsc --noEmit` 0 errors, `biome format` clean, no snapshots changed — evidence that nothing in the repo relies on a non-positive schematic dimension.

### DIFF
--- a/lib/components/normal-components/Board.ts
+++ b/lib/components/normal-components/Board.ts
@@ -23,6 +23,7 @@ import { Subcircuit_doInitialRenderIsolatedSubcircuits } from "../primitive-comp
 import { Subcircuit_getSubcircuitPropHash } from "../primitive-components/Group/Subcircuit_getSubcircuitPropHash"
 import type { BoardI } from "./BoardI"
 import { Board_doInitialPcbPlacementDesignRuleChecks } from "./Board_doInitialPcbPlacementDesignRuleChecks"
+import { Board_doInitialSchematicDimensionChecks } from "./Board_doInitialSchematicDimensionChecks"
 
 const MIN_EFFECTIVE_BORDER_RADIUS_MM = 0.01
 
@@ -578,6 +579,11 @@ export class Board
 
   doInitialPcbPlacementDesignRuleChecks() {
     Board_doInitialPcbPlacementDesignRuleChecks(this)
+  }
+
+  doInitialSchematicTraceRender() {
+    super.doInitialSchematicTraceRender?.()
+    Board_doInitialSchematicDimensionChecks(this)
   }
 
   updatePcbDesignRuleChecks() {
--- a/lib/components/normal-components/Board_doInitialSchematicDimensionChecks.ts
+++ b/lib/components/normal-components/Board_doInitialSchematicDimensionChecks.ts
@@ -1,0 +1,65 @@
+import type { AnyCircuitElement } from "circuit-json"
+import type { Board } from "./Board"
+
+/**
+ * Schematic dimensions that must be greater than zero, per element type.
+ *
+ * A negative or zero size is accepted silently today — `<chip schWidth={-4} />`
+ * emits `size: { width: -4 }` and `<schematiccircle radius={-1} />` emits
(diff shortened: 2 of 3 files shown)


DPS0340:

Sorry for the noise.

PR #2864, opened 2026-07-25
fix: report ports that match no pad in their footprint (#2863)

### LINKED ISSUE #2863: pinLabels naming a pin the footprint lacks creates an unroutable port with no error
### What happens

`pinLabels` can name a pin the footprint doesn't have, and core accepts it silently:

```tsx
<chip name="U1" footprint="soic8" pinLabels={{ pin1: ["A"], pin99: ["Z"] }} />
```

```
source_port   9   ← "A", "Z", pin2..pin8
pcb_smtpad    8
pcb_port      8   ← nothing behind "Z"
errors        0
```

A 9th source port is created for a pad that doesn't exist. Nothing reports it.

### Why it matters

Connecting to that pin gives an error that points at the wrong thing:

```tsx
<trace from=".R1 > .pin1" to=".U1 > .Z" />
```

```
pcb_trace_missing_error: Trace [.R1 > .pin1 to .U1 > .Z] is not connected (it has no PCB trace)
```

The trace is fine — the pin label is wrong. The message sends you to inspect routing instead of the typo, and there's no indication that `.U1 > .Z` was never routable in the first place.

### Cause

`Port.doInitialPcbPortRender` returns early when a port matches no PCB primitive:

```ts
const pcbMatches = matchedComponents.filter((c) => c.isPcbPrimitive)
if (pcbMatches.length === 0) return
```

The neighbouring case — a port matching *multiple* non-overlapping pads — already reports `source_ambiguous_port_reference`. Only the zero-match case is silent.

### Important subtlety for whoever fixes this

A blanket check here is wrong. A component with **no footprint at all** (e.g. a bare `<solderjumper />`) legitimately has zero PCB primitives, and every one of its pins would be flagged. I hit exactly that: my first version broke 6 existing tests (`solderjumper`, `jumper`, panel and subcircuit repros).

The check needs to fire only when the parent component has *some* PCB primitives but this particular port matches none:

```
<solderjumper />        → 0 pads total → not a misconfiguration
<chip footprint="soic8" pinLabels={{pin99}}> → 8 pads, pin99 matches none → misconfiguration
```

PR ready.

DPS0340:
Closes #2863.

```tsx
<chip name="U1" footprint="soic8" pinLabels={{ pin1: ["A"], pin99: ["Z"] }} />
```

```
before  9 source_ports, 8 pads, 8 pcb_ports, 0 errors
after   same elements, plus
        Pin "Z" on U1 has no matching pad in its footprint, so it cannot be
        connected. Check that pinLabels only names pins the footprint provides.
```

Connecting to that pin previously produced `pcb_trace_missing_error: Trace ... is not connected`, which points at the trace when the actual mistake is the pin label.

`Port.doInitialPcbPortRender` returned early on `pcbMatches.length === 0`. The neighbouring case — a port matching *multiple* non-overlapping pads — already reports `source_ambiguous_port_reference`; only the zero-match case was silent. This reuses `source_component_misconfigured_error`, so no new error type.

### The guard condition is the whole point

My first version was a blanket check, and it **broke 6 existing tests** — `solderjumper` (×2), `jumper`, a KiCad panel repro, `subcircuit-circuit-json10`, and a subpanel placement test. The cause is legitimate:

```
<solderjumper />                              → 0 pads total  → not a misconfiguration
<chip footprint="soic8" pinLabels={{pin99}}>  → 8 pads, pin99 matches none → misconfiguration
```

A component with no footprint at all has no PCB primitives, so *every* pin matches nothing. The check now fires only when the parent has **some** PCB primitives but this particular port matches none. All 6 tests pass again without being modified — I fixed the check rather than the tests.

### Verification

**The test bites.** Reverting `Port.ts`:

```
Expected: 1   Received: 0
(fail) a pinLabels entry with no matching pad is reported
```

Three tests pin all three states:

1. `pin99` on a soic8 → exactly 1 error naming `"Z"` and `U1`, with the valid resistor not flagged;
2. a bare `<solderjumper />` → `[]` (the false-positive case that broke the 6 tests);
3. a fully valid soic8 with `pin1`/`pin8` labels → `[]`.

So neither a blanket check nor a no-op can pass.

**Suite:** `1262 pass / 0 fail`, `tsc --noEmit` 0 errors, `biome format` clean, no snapshots changed.

### DIFF
--- a/lib/components/primitive-components/Port/Port.ts
+++ b/lib/components/primitive-components/Port/Port.ts
@@ -444,7 +444,35 @@ export class Port extends PrimitiveComponent<typeof portProps> {
 
     const pcbMatches = matchedComponents.filter((c) => c.isPcbPrimitive)
 
-    if (pcbMatches.length === 0) return
+    if (pcbMatches.length === 0) {
+      // A port that matches no PCB primitive can never be routed. This happens
+      // when `pinLabels` names a pin the footprint doesn't have (e.g. `pin99`
+      // on a soic8): the source port is created, but there is no pad behind it.
+      // Without this the only feedback is a later "trace is not connected"
+      // error, which points at the trace rather than the bad pin label.
+      const parentComponent = this.getParentNormalComponent()
+      const sourceComponentId = parentComponent?.source_component_id
+      // Only meaningful when the component has *some* pads: a component with no
+      // footprint at all (e.g. a bare `<solderjumper />`) legitimately has no
+      // PCB primitives, and every one of its pins would otherwise be flagged.
+      const parentHasAnyPcbPrimitive = (parentComponent?.children ?? []).some(
+        (child: any) => child.isPcbPrimitive,
+      )
+      if (
+        sourceComponentId &&
+        parentHasAnyPcbPrimitive &&
+        this._parsedProps.pinNumber !== undefined
+      ) {
+        const componentName = parentComponent?.props?.name ?? "unknown"
+        const portLabel = this.props.name ?? `pin${this._parsedProps.pinNumber}`
+        db.source_component_misconfigured_error.insert({
+          error_type: "source_component_misconfigured_error",
+          source_component_ids: [sourceComponentId],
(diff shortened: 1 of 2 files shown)


DPS0340:

Sorry for the noise.

PR #2869, opened 2026-07-25
Pass fallback net label widths to schematic trace solver

imrishabh18:
## Summary

- resolve the same fallback net-label text that Core renders for long schematic connections
- measure that rendered label and pass its width through `InputDirectConnection.netLabelWidth`
- add a repro119 assertion for the `U1_pin4` width and update affected routing snapshots

## Why this belongs in Core

The schematic trace solver already accepts `netLabelWidth`, but Core only populated it for explicit trace labels. Generated fallback labels such as `U1_pin4` were therefore routed using the solver's placeholder width (0.45) even though Core rendered a 0.96-wide label. Passing the rendered width at the input boundary keeps the solver generic and makes routing match Core's actual output.

## Verification

- `bunx tsc --noEmit`
- `bun run format`
- focused affected snapshot tests (5 passing)
- full `bun test`: 1255 passing, with three expected snapshot updates subsequently accepted; the only other failure was an unrelated differential-pair test exceeding its 5s timeout, and it passed when rerun alone (3.94s)


tests/repros/__snapshots__/repro91-schematic-trace-junction-schematic.snap.svg
imrishabh18, line 1: This removes the GND symbol

seveibar (approved):

PR #2872, opened 2026-07-25
fix: honour the PinCount spelling in schPinArrangement (#2871)

### LINKED ISSUE #2871: schPinArrangement with leftPinCount/rightPinCount creates zero schematic ports
### Summary

`schPinArrangement` written with the **current** `...PinCount` field names produces a chip with **zero schematic ports**. The deprecated `...Size` names work.

```tsx
<chip name="U1" footprint="soic8" schPinArrangement={{ leftPinCount: 4, rightPinCount: 4 }} />
```

| arrangement | `schematic_port`s | sides |
|---|---|---|
| `{ leftSize: 4, rightSize: 4 }` | 8 | `{left: 4, right: 4}` |
| `{ leftPinCount: 4, rightPinCount: 4 }` | **0** | `{}` |
| no arrangement at all | 8 | `{left: 4, right: 4}` |

No error is raised. The chip renders with a schematic body and no pins, which looks like a symbol problem rather than a prop that was silently dropped. It's also worse than passing nothing — omitting `schPinArrangement` entirely gives you 8 ports.

`...PinCount` is not an invented spelling. `@tscircuit/props` declares both, and marks the `Size` fields as deprecated **in favour of** `PinCount`:

```ts
// schematicPortArrangement in @tscircuit/props
leftSize: z.number().optional().describe("@deprecated, use leftPinCount"),
...
leftPinCount: z.number().optional(),
```

So the documented, non-deprecated spelling is the one that doesn't work.

### Cause

Two places read `${side}Size` without a `${side}PinCount` fallback.

**1. Port creation** — `NormalComponent._addChildrenFromStringOrProps` (~line 335):

```ts
for (const side of sides) {
  const size = (schPortArrangement as any)[`${side}Size`]
  for (let i = 0; i < size; i++) { ... }
}
```

With `leftPinCount`, `size` is `undefined`, the loop body never runs, and no ports are queued.

**2. Schematic box sizing** — `getSizeOfSidesFromPortArrangement`:

```ts
const { leftSize = 0, rightSize = 0, topSize = 0, bottomSize = 0 } = pa as any
return { leftSize, rightSize, topSize, bottomSize }
```

Every side comes back `0`, so `getAllDimensionsForSchematicBox` lays out a box with no pins.

Both have to be fixed — patching only the first still yields `source_port` 8 / `schematic_port` 0, because the box has nowhere to put them. I confirmed that intermediate state while tracing this.

Notably, the same file **already knows about the `PinCount` names** elsewhere:

```ts
["leftSize","rightSize","topSize","bottomSize",
 "leftPinCount","rightPinCount","topPinCount","bottomPinCount"]
  .some((key) => key in schPortArrangement)
```

and `underscorifyPortArrangement` maps `leftPinCount → left_size` correctly. So the support is partial: the circuit-JSON serialisation handles it, port creation and box layout don't.

### Expected

`{ leftPinCount: 4, rightPinCount: 4 }` should behave identically to `{ leftSize: 4, rightSize: 4 }`, including for `topPinCount` / `bottomPinCount`.

PR follows.

DPS0340:
Fixes #2871.

### What was wrong

`schPinArrangement={{ leftPinCount: 4, rightPinCount: 4 }}` produced a chip with **zero schematic ports**, silently:

| arrangement | `schematic_port`s | sides |
|---|---|---|
| `{ leftSize: 4, rightSize: 4 }` | 8 | `{left: 4, right: 4}` |
| `{ leftPinCount: 4, rightPinCount: 4 }` | **0** | `{}` |
| no arrangement at all | 8 | `{left: 4, right: 4}` |

Worse than passing nothing, and with no error — it reads as a broken symbol rather than a dropped prop.

`...PinCount` is the **current** spelling. `@tscircuit/props` marks the `Size` fields `"@deprecated, use leftPinCount"`, so the documented name was the broken one.

### The fix — two sites, and both are required

**1. Port creation** (`NormalComponent`, ~line 335) read only `${side}Size`, so `size` was `undefined` and the loop never ran.

**2. Box layout** (`getSizeOfSidesFromPortArrangement`) destructured only the `Size` keys, returning `0` for every side.

Patching just the first is not enough, and I verified that rather than assuming: with only site 1 fixed the render produces **`source_port` 8 / `schematic_port` 0** — the ports exist but the box has no room allocated for them. Both sites needed the fallback.

Support was already partial elsewhere in the same area, which is what made this look like an oversight rather than a deliberate omission: `underscorifyPortArrangement` maps `leftPinCount → left_size` correctly, and `NormalComponent` already lists all eight key names when deciding whether an arrangement is explicit. Only these two spots missed it.

Implemented as `?? ` fallbacks (`PinCount` first, then the deprecated `Size`), so existing arrangements are untouched.

### Verification

After the fix:

```
{ leftPinCount: 4, rightPinCount: 4 }                          → 8 ports {left:4, right:4}
{ leftSize: 4, rightSize: 4 }                                  → 8 ports {left:4, right:4}
{ leftPinCount: 2, rightPinCount: 6 }                          → 8 ports {left:2, right:6}
{ leftPinCount:2, rightPinCount:2, topPinCount:2, bottomPinCount:2 } → 8 ports on all four sides
```

### Tests

`tests/components/normal-components/sch-pin-arrangement-pin-count.test.tsx`, four tests:

| case | expected |
|---|---|
| `leftPinCount`/`rightPinCount` 4/4 | 8 ports, `{left: 4, right: 4}` |
| `PinCount` 2/6 vs `Size` 2/6 | **results must be equal** — asserted by comparison, not against a constant |
| all four `PinCount` sides | ports on `left`, `right`, `top`, `bottom` |
| `leftSize`/`rightSize` alone | unchanged — the fix must not break the existing spelling |

The comparison test is the important one: it fails in either direction if the two spellings ever diverge again. The last row is the regression guard.

Bite-proofed: reverting `lib/` takes the file from **4 pass** to **1 pass / 3 fail**, with only the deprecated-spelling test still passing.

Full suite: **1259 pass / 0 fail**, `npx tsc --noEmit` clean, biome clean, no snapshot churn.

### DIFF
--- a/lib/components/base-components/NormalComponent/NormalComponent.ts
+++ b/lib/components/base-components/NormalComponent/NormalComponent.ts
@@ -332,7 +332,13 @@ export class NormalComponent<
       const sides = ["left", "right", "top", "bottom"]
       let pinNum = 1
       for (const side of sides) {
-        const size = (schPortArrangement as any)[`${side}Size`]
+        // `${side}PinCount` is the current spelling and `${side}Size` is the
+        // deprecated one (see schematicPortArrangement in @tscircuit/props).
+        // Only the deprecated form was read here, so an arrangement written with
+        // the current names created no ports at all.
+        const size =
+          (schPortArrangement as any)[`${side}PinCount`] ??
+          (schPortArrangement as any)[`${side}Size`]
         for (let i = 0; i < size; i++) {
           const nextPinNumber = pinNum++
           if (hasExistingOrQueuedPortWithPinNumber(nextPinNumber)) continue
--- a/lib/utils/schematic/getSizeOfSidesFromPortArrangement.ts
+++ b/lib/utils/schematic/getSizeOfSidesFromPortArrangement.ts
@@ -56,6 +56,16 @@ export const getSizeOfSidesFromPortArrangement = (
       bottomSize: getPinsFromSideDefinition(pa.bottomSide).length,
     }
   }
-  const { leftSize = 0, rightSize = 0, topSize = 0, bottomSize = 0 } = pa as any
-  return { leftSize, rightSize, topSize, bottomSize }
+  // `${side}PinCount` is the current spelling; `${side}Size` is the deprecated
+  // one (see schematicPortArrangement in @tscircuit/props, where the Size fields
+  // are marked "@deprecated, use ...PinCount"). Only the deprecated form was
+  // read here, so an arrangement written with the current names produced a box
+  // with zero pins on every side.
+  const arrangement = pa as any
+  return {
+    leftSize: arrangement.leftPinCount ?? arrangement.leftSize ?? 0,
+    rightSize: arrangement.rightPinCount ?? arrangement.rightSize ?? 0,
+    topSize: arrangement.topPinCount ?? arrangement.topSize ?? 0,
+    bottomSize: arrangement.bottomPinCount ?? arrangement.bottomSize ?? 0,
+  }
 }
--- a/tests/components/normal-components/sch-pin-arrangement-pin-count.test.tsx
+++ b/tests/components/normal-components/sch-pin-arrangement-pin-count.test.tsx
@@ -1,0 +1,69 @@
+import { expect, test } from "bun:test"
+import { getTestFixture } from "tests/fixtures/get-test-fixture"
+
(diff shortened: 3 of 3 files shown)


DPS0340:

Sorry for the noise.

PR #2876, opened 2026-07-25
fix: give inline footprints a silkscreen reference designator (#2875)

### LINKED ISSUE #2875: Inline <footprint> components get no silkscreen reference designator
### Summary

A component using an **inline `<footprint>`** gets no silkscreen reference designator, while the same component using a **string footprint** does.

```tsx
// string footprint
<chip name="U1" footprint="soic8" />
// → pcb_silkscreen_text: ["U1"]

// inline footprint
<chip name="J1" footprint={
  <footprint>
    <smtpad pcbX={-0.5} pcbY={0} width="0.55mm" height="1.1mm" shape="rect" portHints={["pin1"]} />
    <smtpad pcbX={0.5}  pcbY={0} width="0.55mm" height="1.1mm" shape="rect" portHints={["pin2"]} />
  </footprint>
} />
// → pcb_silkscreen_text: []
```

Verified directly against current `main`:

```
STRING: silkscreen_text= [ "U1" ]   cad_component= 1
INLINE: silkscreen_text= []         cad_component= 1
```

The board renders and exports without any error, so the missing designator only shows up when you look at the silkscreen — or when an exporter needs it. This is the root cause behind **tscircuit/circuit-json-to-kicad#51** ("Footprint files are missing things like reference designator"): the KiCad converter can't emit `fp_text reference` from data that isn't in the circuit JSON.

### Cause

`NormalComponent._addChildrenFromStringFootprint()` has a single `typeof footprint === "string"` branch:

```ts
if (typeof footprint === "string") {
  const fpCircuitJson = fp.string(footprint).soup()
  const fpComponents = createComponentsFromCircuitJson(
    { componentName: this.name ?? this.componentName, ... },
    fpCircuitJson,
  )
  this.addAll(fpComponents)
}
```

Footprinter emits a `pcb_silkscreen_text` placeholder, and `createComponentsFromCircuitJson` rewrites its text to the component name:

```ts
text: componentName || elm.text,
```

Inline JSX footprints never enter that function, so nothing ever creates the designator for them.

### Expected

An inline footprint should get the same reference designator a string footprint gets, placed clear of the footprint bounds, on the component's layer, and without duplicating one the footprint already declares.

PR follows.

DPS0340:
Fixes #2875. Also unblocks tscircuit/circuit-json-to-kicad#51.

### What was wrong

Inline `<footprint>` components got **no silkscreen reference designator**; string footprints got one. Verified on current `main`:

```
STRING (footprint="soic8"):  silkscreen_text = ["U1"]
INLINE (<footprint>...):     silkscreen_text = []
```

String footprints go through `createComponentsFromCircuitJson`, which rewrites footprinter's placeholder text to the component name. Inline JSX footprints never reach that function, so nothing creates the designator.

This is the root cause of **circuit-json-to-kicad#51** — that converter can't emit `fp_text reference` from data the circuit JSON doesn't contain. @dwiel diagnosed it to core in that thread; this confirms it and fixes it here.

### The fix

A new `_addSilkscreenReferenceDesignatorForInlineFootprint()`, called from `doInitialPcbComponentSizeCalculation()`.

That call site matters: it's the phase where the footprint's primitives have been laid out and real bounds are available, so the designator can be positioned against the actual footprint instead of a guess. Font size is derived from the footprint's larger dimension (clamped to 0.2–0.6mm) and the text sits one font-size above the top edge — the same relationship footprinter uses.

Guards, all covered by tests:

- Only runs for non-string footprints, so the working path is untouched.
- Skips if a text with the component name already exists, so a footprint that declares its own designator isn't duplicated.
- Uses the component's resolved layer, so bottom-layer parts get bottom silkscreen.

### Snapshot churn — 46 files, please read

**46 PCB snapshots change.** I verified this is the intended designator appearing rather than anything shifting:

- Every changed snapshot gains exactly one `<text ... class="pcb-silkscreen-text ...">` element containing the component name — 47 added silkscreen lines across the 46 files, and nothing else changed.
- Sampled `repro9-platedhole-pcbX-pcbY`: the single changed line is a new `<text>` reading `R1`. That test uses an inline `<footprint>`, so it's exactly the case this fixes.
- Every one of the 46 is an inline-footprint test. String-footprint snapshots are untouched.

All 46 failures were snapshot mismatches — **zero logic failures** before updating (I checked the failure text of every one).

If you'd rather not take this much snapshot churn in one go, I'm happy to split it, or to gate the behaviour behind a flag — say the word.

### Tests

`tests/components/normal-components/inline-footprint-reference-designator.test.tsx`, six tests:

| case | expected |
|---|---|
| inline footprint | silkscreen text contains `"J1"` |
| string footprint | exactly **one** `"U1"` — proves no duplicate on the working path |
| inline footprint | text's `pcb_component_id` matches the component, layer `top` |
| inline footprint | text sits **above the topmost pad edge**, font size finite and > 0 |
| `layer="bottom"` | designator on the **bottom** silkscreen |
| footprint that already has a `<silkscreentext text="J1">` | still exactly **one** `"J1"` |

Rows 2 and 6 are the guard rails against duplication; row 4 pins placement so a regression that dropped the text onto the pads would fail.

### Verification

`bun test`: **1266 pass / 0 fail** (1259 on `main`; +6 new, +1 from the added file count). `npx tsc --noEmit` clean, biome clean.


DPS0340:

Sorry for the noise.

PR #2880, opened 2026-07-26
Make Matchpack layout failures nonfatal

imrishabh18:
## What changed

- Stops throwing an execution error when the Matchpack solver reports failure.
- Records a `schematic_layout_error` with `is_fatal: false` and the original solver error in its message.
- Returns from the layout phase without applying any replacement placement, leaving existing schematic positions untouched.
- Converts repro159 from `test.failing` into a normal passing schematic snapshot regression.

## Why

Repro159 creates a 146-component strong-connection partition from the XIAO, series resistor, and 144-LED data chain. PackSolver2 is still making progress when it reaches its fixed 100,000-iteration ceiling, but core currently throws and aborts the entire circuit render.

Automatic schematic layout failure should be represented in Circuit JSON rather than preventing Circuit JSON from being produced.

## Impact

Circuits that exhaust Matchpack's iteration budget now finish rendering and expose a non-fatal `schematic_layout_error`. This change deliberately does not introduce a secondary layout algorithm or alter component positions after Matchpack fails.

## Validation

- `./node_modules/.bin/tsc --noEmit`
- `bun test tests/repros/repro159-matchpack-led-matrix-iterations.test.tsx` — passes in about 35 seconds
- `bun test tests/groups/group-schematic-matchpack-test.test.tsx tests/groups/group-schematic-matchpack-respects-margin.test.tsx tests/groups/group-schematic-matchpack-nested-off-origin.test.tsx`
- `git diff --check`


lib/components/primitive-components/Group/applySchematicMatchPackLayoutToTree.ts
```diff
@@ -854,14 +854,23 @@ export function applySchematicMatchPackLayoutToTree<
...
+      subcircuit_id: group.subcircuit_id ?? undefined,
+    })
+    return
   }
 
   const outputLayout = solver.getOutputLayout()
   debug("OutputLayout:", JSON.stringify(outputLayout, null, 2))
 
-  debug("Solver completed successfully:", !solver.failed)
+  debug("Using Matchpack layout")
```
imrishabh18:
```suggestion
  debug("Solver completed successfully:", !solver.failed)
```

lib/components/primitive-components/Group/applySchematicMatchPackLayoutToTree.ts
```diff
@@ -1128,5 +1137,5 @@ export function applySchematicMatchPackLayoutToTree<
     }
   }
 
-  debug("Matchpack layout completed successfully")
+  debug("Schematic layout completed successfully")
```
imrishabh18:
```suggestion
  debug("Matchpack layout completed successfully")
```

seveibar (approved):

PR #2883, opened 2026-07-27
Emit bus constraints in Simple Route JSON

seveibar:
## Summary

- add the `<bus />` primitive to the core catalogue and intrinsic JSX types
- resolve bus trace names or port selectors to routing-scope SRJ connection names
- emit bus membership through `SimpleRouteJson.buses`
- add visual regression tests for both supported authoring forms

## Why

The autorouter boundary currently carries individual connections and differential-pair constraints, but it cannot preserve higher-level bus membership. Without that metadata, routers cannot intentionally order or attract related BGA breakout traces.

Bus membership is routing intent, so this keeps it in Simple Route JSON rather than adding a persistent Circuit JSON element.

## Developer impact

Custom autorouters receive:

```ts
{
  buses: [
    {
      name: "DATA",
      connectionNames: ["source_trace_0", "source_trace_1"]
    }
  ]
}
```

Authors can identify members with trace names or port selectors such as `.U1 > .pin1`.

## Dependencies

- tscircuit/props#758 (merged)
- tscircuit/krt-wasm#17 (merged consumer implementation)

## Validation

- `bunx tsc --noEmit`
- `bun run build`
- `bun test tests/components/primitive-components/bus tests/repros/repro-simple-route-json-45-degree.test.tsx`
- visual inspection of the trace-name and port-selector PCB snapshots
- `git diff --check`


seveibar (commented):
CC @0hmX we are now passing buses to SRJ in core, this is to hopefully help with bus-level and pair-level repros. Note: Generally a bus contains a set of differential pairs, differential pair is generally the more important thing, but you don't want one pair to have a super long trace while the other pairs on the bus have a super short trace, so there is a bus skew and pair skew

PR #2886, opened 2026-07-27
Support selectable mapped ports on schematic symbols

MustafaMulla29:
## Summary

- model schematic-symbol library pins as real, selectable `Port` children
- initialize those ports through the established `InitializePortsFromChildren` lifecycle
- map symbol ports to physical chip ports through explicit `chipRef` and `connections`
- reuse the mapped physical port's source and PCB identity, so schematic traces and PCB traces describe the same electrical connection
- forward `displayName` through the schematic symbol's source component

## Selector behavior

A mapped schematic symbol is a sibling of the physical chip; `chipRef` establishes an electrical mapping, not component containment.

```tsx
<chip name="Q1" ... />

<schematicsymbol
  name="Q1A"
  chipRef=".Q1"
  symbolName="n_channel_e_mosfet_transistor_horz"
  connections={{
    gate: "Q1.G1",
    source: "Q1.S1",
    drain: "Q1.D1",
  }}
/>
```

Therefore:

- physical chip selectors remain valid, such as `Q1.pin1` and `Q1.G1`
- schematic-symbol selectors are rooted at its own name, such as `Q1A.pin3` and `Q1A.gate`
- selectors such as `.Q1 > .A` are not introduced because the schematic symbol is not contained by `Q1`

Pin numbers on the library symbol do not have to match physical package pin numbers. In the MOSFET test, `Q1A.pin1` (drain) maps to `Q1.pin7`, while `Q1A.pin3` (gate) maps to `Q1.pin1`, according to `connections`.

Both `chipRef` and `connections` are required for mapping. Supplying only one emits a `source_property_ignored_warning`. Port labels are not auto-matched.

The referenced chip remains the only physical PCB component. A trace may target either the schematic-symbol port or the mapped physical chip port and will use the same underlying electrical and PCB port identity.

## Testing

- `bun test tests/components/primitive-components/schematic-symbol-*.test.tsx`
- `bun run build`
- schematic and PCB snapshot coverage in the mapped MOSFET tests
- explicit assertions that `Q1A.pin1` maps to `Q1.pin7` and `Q1A.pin3` maps to `Q1.pin1`


seveibar (changes requested):
sorry @MustafaMulla29 i think you're messing up the spec quite a bit. `name="Q1A" chipRef="Q1"`, the schematicsymbol IS NOT CONTAINED WITHIN THE COMPONENT therefore the name resolution IS NOT ".Q1 > .A" because it is not contained

seveibar (approved):

PR #2888, opened 2026-07-27
Prevent duplicate nested schematic net labels

imrishabh18:
## What changed

- detects child schematic ports whose port-to-net trace is owned by another subcircuit routing pass
- suppresses the child connection only when the same source port has another schematic representation on the same sheet, such as the parent `schematicbox`
- preserves child-side net labels when no alternate visual representation exists
- updates the `repro160` schematic snapshot to remove the duplicate GND rail while preserving the explicit U1 connection and the U1A GND representation

## Root cause

In repro160, U1 and U1A are two schematic representations of the same source component. The GND port-to-net source trace belongs to the parent board, but the child `workaround` pass also treated its U1 representation as needing a fallback GND label. Both passes therefore visualized the same logical connection.

Trace ownership alone is not enough to suppress a child label: ordinary cross-subcircuit connections still need endpoint labels inside each child. The fix combines stable source-trace ownership with Circuit JSON schematic-port identity. A child port is skipped only when an externally owned port-to-net trace can be rendered through another schematic representation of that exact source port on the same sheet.

The implementation performs a targeted Circuit JSON lookup by `source_port_id` only for externally owned port-to-net traces. This avoids root component-tree selection and keeps the decision local and deterministic.

## Impact

Nested schematic boxes no longer generate duplicate fallback net labels for parent-owned connections. Ordinary child subcircuits keep the endpoint labels needed to show their connections to parent nets.

## Validation

- `bun test tests/components/normal-components/chip-with-subcircuit-net-label.test.tsx tests/repros/repro160.test.tsx tests/repros/repro126-trace-overlap-box-resistor.test.tsx tests/repros/repro156-schematicbox-chipref-netlabel.test.tsx tests/repros/repro154-subcircuit-exposed-port-schematic-connections.test.tsx tests/repros/repro-boxed-subcircuit-internal-port-name-leak.test.tsx tests/repros/repro-boxed-subcircuit-missing-fallback-label.test.tsx tests/repros/repro-subcircuit-cross-boundary-trace-name-label.test.tsx tests/groups/group-schematic-box-connected.test.tsx`
- `./node_modules/.bin/tsc --noEmit`
- Biome checks on all modified TypeScript files


lib/components/primitive-components/Group/Group_doInitialSchematicTraceRender/getSchematicPortIdsWithAssignedNetLabels.ts
```diff
@@ -5,10 +5,22 @@ import { type SchematicPortId, asSchematicPortId } from "./port-id-types"
...
-  const netLabels = group.selectAll("netlabel") as NetLabel[]
+  const netLabelsInGroup = group.selectAll("netlabel") as NetLabel[]
+  // A child subcircuit's selector scope cannot see a net label owned by an
+  // ancestor routing pass, even when that label connects to a port in the
+  // child. Include both scopes so the child solver does not add a fallback.
+  const netLabels = opts.includeNetLabelsOutsideGroup
+    ? [
+        ...new Set<NetLabel>([
+          ...netLabelsInGroup,
+          ...((group.root?.selectAll("netlabel") ?? []) as NetLabel[]),
```
seveibar: this is super expensive?

seveibar: non-deterministic

seveibar (commented):
see comments

seveibar (approved):

PR #2899, opened 2026-07-28
Auto-size chips for opposing pin labels

imrishabh18:
## Summary

- automatically size schematic chip boxes using the longest displayed pin label on each opposing side
- preserve explicit `schWidth` values and existing one-sided sizing behavior
- support readonly pin-label alias arrays in schematic dimension helpers
- add visual regressions for the reported RP2040 case and symmetric long opposing labels

## Root cause

Automatic schematic width used a single global label estimate. For array-form `pinLabels`, `label.length` measured the number of aliases instead of the displayed label text, and the calculation did not reserve space for the longest labels on both the left and right sides. This allowed the two label columns to overlap.

## Before / after

| Before | After |
| --- | --- |
| RP2040 schematic before automatic opposing-label sizing | RP2040 schematic after automatic opposing-label sizing |

## Validation

- `bun test tests/components/normal-components/chip-auto-width-long-pin-labels.test.tsx tests/components/normal-components/chip-auto-width-opposing-pin-labels.test.tsx`
- `bun test tests/utils/schematic/getAllDimensionsForSchematicBox.test.ts tests/repros/repro123-schematic-box-inner-label-collision.test.tsx`
- `bunx tsc --noEmit`
- `bun run build`


lib/utils/schematic/getAllDimensionsForSchematicBox.ts
```diff
@@ -2,6 +2,17 @@ import { getSizeOfSidesFromPortArrangement } from "./getSizeOfSidesFromPortArran
 import { parsePinNumberFromLabelsOrThrow } from "./parsePinNumberFromLabelsOrThrow"
 
 const DEFAULT_SCHEMATIC_BOX_PADDING_MM = 0.4
+const ESTIMATED_PIN_LABEL_CHARACTER_WIDTH = 0.1
```
0hmX, line 5: Can’t we get the raw, true value for this? It might be difficult, but I think this needs its own investigation.

imrishabh18: I am not sure if we can get the actual size, cause it varies the fallback font could be different for each browser and OS

PR #2904, opened 2026-07-28
Fix fuse sch show ratings

GokulPandi-M:
Motivation: 

The <fuse /> component already accepted the schShowRatings prop, but setting it to false was broken and had no effect. The component would always render the 1A / 24V text regardless of what the user passed. This PR fixes the internal rendering logic so that schShowRatings={false} correctly hides the ratings on the schematic symbol as expected.

Before:
Screenshot 2026-07-29 at 9 53 57 PM

After:
Screenshot 2026-07-29 at 9 47 48 PM

### DIFF
--- a/lib/components/normal-components/Fuse.ts
+++ b/lib/components/normal-components/Fuse.ts
@@ -19,7 +19,7 @@ export class Fuse extends NormalComponent<typeof fuseProps, PassivePorts> {
   }
 
   _getSchematicSymbolDisplayValue(): string | undefined {
-    if (!this._parsedProps.schShowRatings) {
+    if (this._parsedProps.schShowRatings === false) {
       return undefined
     }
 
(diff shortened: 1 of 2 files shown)


mohan-bee (commented):
dont use ternaries

mohan-bee (approved):
LGTM

seveibar (commented):
needs visual snapshot

lib/components/normal-components/Fuse.ts
```diff
@@ -19,6 +19,10 @@ export class Fuse extends NormalComponent<typeof fuseProps, PassivePorts> {
   }
 
   _getSchematicSymbolDisplayValue(): string | undefined {
+    if (this._parsedProps.schShowRatings === false) {
+      return undefined
```
imrishabh18: Don't return `undefined`

seveibar (approved):

PR #2906, opened 2026-07-28
repro: custom symbol name label missing for <chip symbol={...}> components

itisrohit:
When a component uses a custom `<symbol>` (e.g. `<chip symbol={<symbol>...</symbol>}>`), renders it via `_doInitialSchematicComponentRenderWithReactSymbol` which never inserts a `schematic_text` element for the component's name/refdes.

The user sets `name=\L1\` but no label appears on the schematic. This is inconsistent with box-type components which auto-generate the name label through `_doInitialSchematicComponentRenderWithSchematicBoxDimensions` (`db.schematic_text.insert(...)`).


imrishabh18 (changes requested):
It's circuit-to-svg issue, create a repro and fix in there

PR #2916, opened 2026-07-29
fix: bump @tscircuit/circuit-json-util to 0.0.102 so PCB transforms (grid layout, subcircuit anchoring) move solder paste with pads

### LINKED ISSUE #2897: Grid/array layout repositions pads without transforming their solder paste
# Grid/array layout repositions pads without transforming their solder paste

## Summary

When pads are moved by grid/array layout paths, the `pcb_smtpad` elements get their new
positions but the associated `pcb_solder_paste` elements keep their pre-layout coordinates —
the stencil apertures end up millimetres away from the pads they belong to.

## Repro (on `main`, v0.0.1531)

A `<chip>` with plain rect pads placed via a grid arrangement: after `circuit.render()`,

```
pad  shape=rect cx=1.000 cy=1.905   → its paste at x=-2.150 y=1.905   (3.15 mm away)
pad  shape=rect cx=-2.65 cy=0      → its paste at x=1.000 y=0        (3.65 mm away)
```

Every pad in the grid shows the same pattern: paste at the pad's pre-layout position. The
displacement is identical across shapes (observed with rect pads on pristine `main`, so this is
not related to #2895's polygon paste — polygon paste simply inherits the same behavior).

The constraint-layout path (`_setPositionFromLayout`) and packer moves
(`_moveCircuitJsonElements`) DO move paste; whatever path the grid layout takes bypasses both.

## Why it matters

An exported stencil for any grid-arranged board has apertures over soldermask instead of pads.
Downstream CAM can detect-and-snap it (that is how we found it), but the artifact as emitted is
wrong at the source.

## Suggested direction

Whichever layout application path the grid/array arrangements use should either route through
`_setPositionFromLayout` (which now translates all owned paste with the pad) or apply the same
delta to `pcb_solder_paste` elements alongside `pcb_smtpad`. Happy to turn the repro into a
failing test if useful.

Hero988:
## What

Bumps `@tscircuit/circuit-json-util` from `^0.0.101` to `^0.0.102` and adds a grid-layout regression test asserting every `pcb_solder_paste` stays centered on its pad.

## Why / Origin

Fixes #2897 (grid/array layout repositions pads without transforming their solder paste). The root cause was in circuit-json-util's `transformPCBElement`, which silently skipped `pcb_solder_paste` and other component-attached geometric elements — fixed upstream in tscircuit/circuit-json-util#110 and released in v0.0.102. Core's `^0.0.101` pin semver-excludes `0.0.102`, so the fix needs this explicit bump to reach grid layouts. Root-cause analysis is in https://github.com/tscircuit/core/issues/2897#issuecomment-5108529394.

## Before / After

Before: a 2×2 `pcbGrid` of 0402s emits all 8 `pcb_solder_paste` records at pre-layout coordinates, up to 3.28 mm from their pads — stencil apertures land off-pad.
After: every paste aperture follows its pad exactly (Δ = 0.000 mm).

## Testing

- New `tests/features/grid/pcb-grid5-solder-paste.test.tsx` fails on `^0.0.101` (`Received: 3.28`) and passes with `0.0.102`.
- Ran the surrounding layout surface locally (grid, board-grid, panel grids, pcb-pack, flex, solder-paste creation tests): 44 pass / 0 fail, snapshots unchanged.
- `tsc --noEmit` clean; biome format clean.


Hero988:
CI note: shard 1 flagged `subcircuit-circuit-json12` (pcbAnchorAlignment) — the PCB snapshot changed because `transformPCBElements` in circuit-json-util 0.0.102 now moves `pcb_solder_paste` with the anchored subcircuit (previously the paste was left at its pre-transform position). Verified numerically that the paste centers now coincide exactly with the pad centers after the anchor transform, so the new snapshot is the corrected rendering; regenerated it in 3404de17. Also ran the full subcircuit-circuit-json + solder-paste/anchor-related test surface locally (22 files): all pass.

imrishabh18 (commented):
PR title not matching with the snapshot diff

Hero988:
Understood the title was narrower than the fix. The 0.0.102 bump fixes `transformPCBElements` dropping `pcb_solder_paste`, which surfaces in two places: grid layout (covered by the new numeric test `pcb-grid5-solder-paste.test.tsx`, no snapshot) and subcircuit anchor alignment (the `subcircuit-circuit-json12` snapshot — the paste now moves with the anchored subcircuit instead of staying at its pre-transform position; verified the paste centers coincide with the pad centers after the transform). Same root cause, two rendering surfaces.

seveibar (approved):

imrishabh18 (approved):

PR #2918, opened 2026-07-29
repro: multiple traces between same component pair produce only one schematic_trace

itisrohit:
When two `<trace>` elements connect the same pair of components, only one `schematic_trace` is produced. The connectivity layer (`source_trace`) correctly records both connections, but the schematic rendering layer drops all but the first.


imrishabh18 (changes requested):
What is this? This is valid, the netlabels signify there is a connection

PR #2921, opened 2026-07-29
Fix capacitance display to use consistent SI unit formatting

KrishnaX12:
This fixes an issue where string values like `"1000uF"` were bypassing the SI unit formatter and rendering raw strings directly on the schematic. 

The `Capacitor` component now parses all string inputs and runs them through `formatSiUnit` so they display cleanly (e.g. `"1000uF"` -> `1mF`). 

I also ran `bun test -u` to update all the affected snapshot tests to match the new correct display values.

### Before
area_2026-07-31_19-48-42

### After
area_2026-07-31_19-45-04

Resistor and inductor seems fine to me, the issue is with capacitor !!
eg :

area_2026-07-31_19-53-04


seveibar:
try to focus on high impact, in this case this behavior is highly undesirable, we should always display what the user provided

PR #2923, opened 2026-07-29
repro: solder jumper bridge trace disconnected endpoint false positive

itisrohit:
A minimal 2-pin `<solderjumper>` with `pw=0.6604` generates a false `pcb_trace_error`.

The bridge trace has `source_trace_id: undefined` — it is created by the footprinter without connection metadata. The DRC (`checkTracesAreContiguous`) cannot find a `source_trace` for it, so `expectedPorts` is empty and it falls to a geometry-based "any pad" check. Because the trace endpoint lands exactly at the pad edge (`padWidth/2` from center), IEEE 754 compound FP error (~5.55e-17mm) makes `isPointInPad` fail, falsely reporting "disconnected endpoint."

The PCB SVG renders correctly — the trace connects pad-to-pad visually.

**Root cause:** `SolderJumper.doInitialPcbPortAttachment()` in core does not create `source_trace` entries for bridge traces or set `source_trace_id` on their `pcb_trace` entries. Fix belongs in `lib/components/normal-components/SolderJumper.ts`.


seveibar (approved):
the issue isn't geometric fwiw, the issue is related to how the trace is created without proper connection metadata

Issue #2896, opened 2026-07-28
Solder-paste aperture ratio is hard-coded at 0.7 — make it configurable


gwhthompson:
# Solder-paste aperture ratio is hard-coded at 0.7 — make it configurable

## Summary

`SmtPad.doInitialPcbPrimitiveRender()` shrinks every solder-paste aperture to `0.7 ×` the pad
dimensions, with no prop and no platform-config override:

- `lib/components/primitive-components/SmtPad.ts:157` — `radius: pcb_smtpad.radius * 0.7`
- `:208-209`, `:221-222`, `:263-264` — `width/height: pcb_smtpad.width * 0.7`

0.7 **linear** is 0.49 of the pad **area**. For fine-pitch and leadless parts that pushes the
stencil below the IPC-7525 release floor, and there is currently no way for a design to say
otherwise.

## Why it matters (measured)

IPC-7525 gives area ratio `AR = (L × W) / (2 × (L + W) × t)` for a rectangular aperture of
length `L`, width `W` in a stencil of thickness `t`, and sets the paste-release floor at
**AR ≥ 0.66**. Below that, paste preferentially stays in the aperture rather than transferring
to the land.

On a production panel (5 board designs, standard 0.12 mm laser-cut stainless stencil), the
0.7 factor put **124 of 465 apertures (27%) below the AR 0.66 floor**:

| part | package | AR at 0.7× | AR if 1:1 |
| --- | --- | --- | --- |
| TAD5242 | VQFN lead | **0.500** | 0.714 |
| OPA2210 | WSON lead | **0.605** | 0.864 |

The manufacturers' own stencil drawings for these same packages specify **1:1 apertures at
0.125 mm**, i.e. the parts are designed to be stencilled at the land size, and the 0.7 factor
is what takes them under the floor.

## Requested change

Make the ratio configurable, keeping today's behaviour as the default so nothing changes for
existing designs:

- per-pad prop on `<smtpad>` (and pass-through from `<footprint>`), e.g.
  `solderPasteMargin` (absolute inset, mirroring the existing `solderMaskMargin`) and/or
  `solderPasteRatio` (linear factor)
- board-level default in `platformConfig` / `<board>` so a design can set one value for the
  whole board rather than per pad
- default stays `0.7` for compatibility

An absolute margin in addition to a ratio is worth having: stencil practice is usually
expressed as an inset in mm (or 1:1), not as a fraction, and a fixed ratio scales the inset
[...]

PR #2930, opened 2026-07-30
fix: emit solder paste for polygon and pill smtpads

Hero988:
## What

`<smtpad shape="polygon">` and `<smtpad shape="pill">` now emit `pcb_solder_paste` — previously both branches of `SmtPad.doInitialPcbPrimitiveRender()` inserted the pad and fell through, so these pads produced no stencil aperture at all. A missing aperture violates no DFM rule, so affected boards ship silently unpasteable.

## Why / Origin

Reported with a real-world case in #2895: TI B0QFN leadless packages import as polygon lands, and 8 of 28 lands on a TPSM82913 buck module — including its only VIN land — passed three JLCPCB DFM rounds with zero aperture. That PR was closed by its author without merging; this is an independent implementation, extended to the pill branch, which has the identical fall-through (noted in #2895's "related" section). Verified on `main` first: a scene with polygon + pill + rect pads yields paste for the rect pad only.

## How

- **Polygon pads** are decomposed into axis-aligned rects (`lib/utils/pcb/decompose-polygon-into-axis-aligned-rects.ts`) with one `rect` paste per rect. Rectilinear lands — the footprint-import case — tile exactly; a slanted edge contributes the tighter of its positions at the two slab boundaries, so no aperture ever extends outside the copper. Rects are emitted 1:1 with the pad: insetting tiles would open gaps between apertures rather than insetting the outline, and manufacturer stencil drawings for leadless lands are 1:1 (#2896 tracks making the ratio configurable). Slivers under 0.05mm are dropped — no stencil releases them. `pcb_solder_paste` has no polygon variant and `circuit-json-to-gerber` throws on unknown paste shapes, so rect decomposition is the non-breaking fix.
- **Pill pads** get a `pill` paste at the same 0.7 factor as every other shape. Rotated pills get a 0.7-scaled `rotated_rect` paste: there is no rotated-pill paste variant, the scaled rect is always contained within the pill outline, and the gerber exporter already handles rotated_rect paste.
- **Layout moves**: `_setPositionFromLayout` wrote `x`/`y` (absent on polygon pads) and snapped a single paste element to the pad center — wrong for polygon pads and for multiple off-center apertures. It now translates polygon points and every owned paste element by the layout delta, and `_moveCircuitJsonElements` delegates to it. This also removes the old polygon branch's double-shift, which re-read the already-updated bounds and pushed the port a further half-delta.

## Before / After

On `main`, a scene with polygon + pill + rect pads: pads `["polygon", "pill", "rect"]` → pastes `["rect"]`. With this PR: an L-shaped polygon land tiles to rect apertures whose summed area equals the copper area with the notch uncovered, the pill pad gets one 0.7 pill aperture, and `pcbGrid` layout keeps every aperture centered on its pad.

## Visual snapshots ([74241c9](https://github.com/tscircuit/core/pull/2930/commits/1bb8fb6794b940e0a2fbf06966a7eed25272ca6b))

| PCB view | Solder paste mask |
| :---: | :---: |
| PCB snapshot: L-shaped polygon pad, pill pad, rotated pill pad | Solder paste mask snapshot: rect apertures tiling the L-shape, pill aperture, rotated-rect aperture |

Same board in both images: an L-shaped polygon pad, a pill pad, and a 30°-rotated pill pad. The paste mask (right) shows the apertures this PR emits — the L-shape tiled 1:1 by rect apertures with the notch uncovered, a 0.7-scaled pill aperture, and a contained 0.7-scaled rotated-rect aperture. On `main` the right image is empty: none of these pads produce any paste. The standard PCB SVG never draws `pcb_solder_paste`, so the mask view comes from a new `toMatchSolderPasteSnapshot` matcher added in the same commit.

## Testing

- `tests/components/primitive-components/smtpad-polygon-pill-solder-paste.test.tsx` — exact L-shape tiling (area, containment, notch uncovered), slanted-edge containment (house pentagon), mask suppression, pill 0.7 paste, rotated-pill paste, and paste tracking pads through grid layout. 5 of 6 fail without the fix (mask suppression passes trivially on `main` since nothing is emitted).
- `tests/utils/pcb/decompose-polygon-into-axis-aligned-rects.test.ts` — tiling/merge/sliver/triangle cases plus a 300-polygon seeded fuzz asserting every rect corner stays inside the polygon.
- Full suite, `tsc --noEmit`, `biome format` green locally.


lib/components/primitive-components/SmtPad.ts
```diff
@@ -313,6 +340,23 @@ export class SmtPad extends PrimitiveComponent<typeof smtPadProps> {
           subcircuit_id: subcircuit?.subcircuit_id ?? undefined,
           pcb_group_id: this.getGroup()?.pcb_group_id ?? undefined,
         } as PcbSmtPadRotatedPill)
+
+        if (shouldCreateSolderPaste)
+          // circuit-json has no rotated-pill paste variant; a 0.7-scaled
+          // rotated rect is always contained within the pill outline
```
seveibar: hack?

seveibar (changes requested):
missing visual snapshot

lib/components/primitive-components/SmtPad.ts
Hero988: I see that, I opened tscircuit/circuit-json#668 adding a `rotated_pill` variant to `pcb_solder_paste`. Once it is in, this line switches to emitting `rotated_pill` so the aperture matches the pad shape exactly.

Hero988:
Visual snapshots added in [74241c9](https://github.com/tscircuit/core/pull/2930/commits/74241c9daa83fbd554783c55527b767bf24b79a1):

| PCB view | Solder paste mask |
| :---: | :---: |
| PCB snapshot | Solder paste mask snapshot |

Same board both sides — L-shaped polygon pad, pill pad, 30°-rotated pill pad. The mask view uses a new `toMatchSolderPasteSnapshot` matcher (the standard PCB SVG doesn't draw `pcb_solder_paste`); on `main` it renders empty.

lib/components/primitive-components/SmtPad.ts
Hero988: I saw circuit-json#668 just got merged thank you, rendering support is up next: tscircuit/circuit-to-svg#645 and tscircuit/circuit-json-to-gerber#130. Once circuit-to-svg releases, this line switches to emitting `rotated_pill` (change + dep bumps ready) so the paste mask matches the pad shape exactly.

Hero988: circuit-to-svg v0.0.394 is out with the rendering support, so the switch is now in: [1bb8fb6](https://github.com/tscircuit/core/pull/2930/commits/1bb8fb6794b940e0a2fbf06966a7eed25272ca6b) emits `rotated_pill` solder paste for rotated pill pads (matching radius, no more rotated_rect approximation), bumps `circuit-json` to ^0.0.456 and `circuit-to-svg` to ^0.0.394, and regenerates the paste mask snapshot — the rotated pad's aperture now renders as a proper stadium. Body images updated to the new SHA.

techmannih (commented):
Very sloppy implementation

Hero988:
Cleaned up in [e35cbe8](https://github.com/tscircuit/core/pull/2930/commits/e35cbe84):

- Custom `toMatchSolderPasteSnapshot` matcher removed — `extend-expect-circuit-snapshot.ts` is back to main and the visual test uses the standard `toMatchPcbSnapshot` (same as #2945; it switches to `{ showSolderPaste: true }` once tscircuit/circuit-to-svg#648 releases so paste shows in the pcb snapshot itself)
- Tests split one per file
- Stale test title fixed (rotated pill emits `rotated_pill` now, not `rotated_rect`)

If a specific part of the implementation still looks off to you, point me at it and I'll fix it.

Hero988:
Rebased onto main (9ee94ca) to clear a merge conflict from the circuit-json ^0.0.457 bump on main no code changes needed, package.json now only carries the circuit-to-svg ^0.0.394 pin. All 13 targeted tests pass locally, tsc clean.

seveibar:
too much time spent on this effort that clearly isn't motivated by a user issue

Hero988:
Understood, fair call. I'll anchor future work to reported user issues rather than self-sourced enhancements. Thanks for the review time.

PR #2933, opened 2026-07-30
Fix multi-port traces in the default autorouter

MustafaMulla29:
## Summary

- preserve every source-trace endpoint when creating Simple Route JSON for the default PCB autorouter
- represent multi-port schematic traces as pairwise direct connections sharing one net ID
- update the merged schematic and PCB repro snapshots
- keep the sequential-trace router's existing two-port limitation unchanged

## Root cause

Both default-router input conversions silently selected only the first two connected ports. A trace such as `[A, B, C]` therefore became `[A, B]`, leaving `C` absent from schematic connectivity and PCB autorouting.

## Fix

The schematic conversion expands an N-terminal trace into adjacent pairwise connections with the same net ID, matching the schematic solver's pair-based input model. The PCB conversion passes all endpoints through `pointsToConnect`, which is the native multi-terminal representation consumed by the capacity autorouter.

## Validation

- `bun test tests/repros/repro-trace-path-with-more-than-two-ports.test.tsx`
- `bun test tests/components/trace tests/utils/autorouting`
- `bun test tests/components/primitive-components/trace*.test.tsx`
- `bunx tsc --noEmit`
- `bunx biome check lib/components/primitive-components/Group/Group_doInitialSchematicTraceRender/createSchematicTraceSolverInputProblem.ts lib/utils/autorouting/getSimpleRouteJsonFromCircuitJson.ts tests/repros/repro-trace-path-with-more-than-two-ports.test.tsx`
- `git diff --check`


lib/utils/autorouting/getSimpleRouteJsonFromCircuitJson.ts
```diff
@@ -364,29 +364,21 @@ export const getSimpleRouteJsonFromCircuitJson = ({
...
-          `(source_port_id: ${portA.source_port_id}) for trace ${trace.source_trace_id} does not have x/y coordinates. Skipping this trace.`,
-        )
-        return null
-      }
-      if (portB.x === undefined || portB.y === undefined) {
-        console.error(
-          `(source_port_id: ${portB.source_port_id}) for trace ${trace.source_trace_id} does not have x/y coordinates. Skipping this trace.`,
-        )
-        return null
+      for (const connectedPort of connectedPorts) {
```
seveibar, line 367: why refactor the pcb oriented-file here?

seveibar (approved):

PR #2935, opened 2026-07-30
repro: negative and zero dimensions accepted with no error

itisrohit:
Negative and zero physical dimensions (hole, via, plated hole) are accepted silently with no error.

Refs #2857


seveibar:
low impact

PR #2936, opened 2026-07-30
repro: trace connecting a port to itself accepted silently

itisrohit:
A `<trace from=".R1 > .pin1" to=".R1 > .pin1" />` resolves both ends to the same port and creates a source_trace with duplicate port IDs. No pcb_trace is generated, and the reproduction shows that this results in a downstream pcb_trace_missing_error rather than a diagnostic indicating that both trace endpoints resolve to the same port.

Refs #2859


MustafaMulla29 (commented):
You just dont throw errors for everything, its simply ignoring that trace i think otherwise we will have to emit errors for infinite cases.

itisrohit:
> You just dont throw errors for everything, its simply ignoring that trace i think otherwise we will have to emit errors for infinite cases.

Fair point , we shouldn't add errors for every edge case.

My concern is that this case doesn't seem to behave like a no-op. It still creates a `source_trace` with the same port on both ends, and that later results in a `pcb_trace_missing_error` because no PCB trace can be generated.

Would you expect a self-connection to be ignored before creating a `source_trace`, or is the downstream `pcb_trace_missing_error` considered the intended behavior?

MustafaMulla29:
If that is the case then you should make your repro better, it has no `pcb_trace_missing_error` assertions etc.

itisrohit:
> If that is the case then you should make your repro better, it has no `pcb_trace_missing_error` assertions etc.

Good point. I've updated the repro to assert the full behavior:

- a source_trace is still created with duplicate port IDs,
- no pcb_trace is generated,
- a downstream pcb_trace_missing_error is emitted,
- and no source_trace_not_connected_error is produced.

This should demonstrate the complete behavior instead of just the missing routed trace.

seveibar:
not a bug

PR #2937, opened 2026-07-30
Fix inherited routingDisabled for sequential traces

0hmX:
Depends on #2938. Merge the reproduction PR first.

## What changed

- Make sequential trace routing honor inherited `routingDisabled` values from parent boards and subcircuits.
- Apply the same inherited check to sequential net-island routing.
- Update the reproduction snapshot so the unexpected child trace disappears.

## Root cause

The async group autorouter already checks `getInheritedProperty("routingDisabled")`, but the sequential trace and net-island paths checked only the immediate subcircuit `_parsedProps.routingDisabled`.

## Before / after

The reproduction in #2938 produces a child `pcb_trace` despite the parent board using `routingDisabled`. With this change, the source trace remains but no PCB trace is generated.

## Validation

- `bun test tests/repros/repro-routing-disabled-nested-sequential-trace.test.tsx tests/examples/example19-subcircuit-overlap-traces.test.tsx tests/components/normal-components/board-routing-disabled.test.tsx tests/root-circuit/platform-routing-disabled.test.tsx tests/subcircuits/subcircuit2-independent-render.test.tsx`
- `bunx tsc --noEmit`
- `bunx biome format` on the changed files
- `git diff --check`


lib/components/primitive-components/Trace/Trace_doInitialPcbTraceRender.ts
```diff
@@ -95,7 +95,10 @@ export function Trace_doInitialPcbTraceRender(trace: Trace) {
 
   if (!parent) throw new Error("Trace has no parent")
 
-  if (subcircuit._parsedProps.routingDisabled) {
+  if (
+    trace.root?.pcbRoutingDisabled ||
```
0hmX, line 99: why this dual check

seveibar (approved):

PR #648, opened 2025-02-24
#578 Fix schematic traces

rohitbhure65:
fix #578 
/claim #578

PR #2945, opened 2026-07-30
feat: make smtpad solder paste aperture configurable via solderPasteMargin

### LINKED ISSUE #2896: Solder-paste aperture ratio is hard-coded at 0.7 — make it configurable

Hero988:
Fixes #2896

## What

Adds a **`solderPasteMargin`** prop to `<smtpad>`: an absolute per-side margin for the solder paste aperture, mirroring the existing `solderMaskMargin` sign convention (positive grows the aperture, negative shrinks it). A margin of `0` gives a 1:1 aperture. If the margin shrinks an aperture to zero, no `pcb_solder_paste` element is emitted.

Today's `0.7 ×` behaviour stays the default when the prop is not set, so nothing changes for existing designs.

Uses the prop added in tscircuit/props#767.

## Why

`SmtPad.doInitialPcbPrimitiveRender()` shrank every paste aperture to `0.7 ×` the pad dimensions (0.49 of the pad area) with no override. As measured in #2896, on a production panel that put 27% of apertures below the IPC-7525 area-ratio release floor of 0.66 (TAD5242 VQFN: AR 0.500, OPA2210 WSON: AR 0.605) — while the manufacturers' own stencil drawings for those packages specify 1:1 apertures.

## Tests

`tests/components/primitive-components/smtpad-solder-paste-config.test.tsx` (8 tests):
- default stays `0.7` with no props set (rect + circle)
- negative, zero and positive margins on rect (inset, 1:1, overprint)
- circle radius gets the margin once per side; width/height twice
- rotated_rect honors the margin and keeps its rotation
- zero-size aperture ⇒ no `pcb_solder_paste` element emitted
- visual snapshots (PCB view + solder paste mask) of the four regimes side by side

## Visual snapshot

Four identical 2×1.4mm rect pads, left to right: **no prop (default 0.7×)** · **`solderPasteMargin={0}`** (1:1) · **`solderPasteMargin={-0.2}`** (inset) · **`solderPasteMargin={0.2}`** (overprint):

| PCB view | Solder paste mask |
| :---: | :---: |
|  |  |

The first aperture is the current default; the mask only changes when the prop is set.

Full suite: no regressions vs `main` (the only failures on my machine are the same pre-existing `renderUntilSettled` timeout flakes that fail identically on unmodified `main`).


seveibar (changes requested):
still missing visual snapshot

Hero988:
Visual snapshots added in [ebb95ab](https://github.com/tscircuit/core/pull/2945/commits/ebb95ab9943a75056dafc666634ad8b47fc32b92) — PCB view + solder paste mask of the four regimes (default 0.7× / margin 0 / −0.2 / +0.2) side by side, also now in the PR body.

tests/components/primitive-components/smtpad-solder-paste-config.test.tsx
```diff
@@ -0,0 +1,222 @@
...
+
+  const [solder_paste] = circuit.db.pcb_solder_paste.list()
+  expect(solder_paste.shape).toBe("rect")
+  if (solder_paste.shape === "rect") {
+    expect(solder_paste.width).toBeCloseTo(1.8)
+    expect(solder_paste.height).toBeCloseTo(0.8)
+  }
+})
+
+test("a zero solderPasteMargin yields a 1:1 aperture", async () => {
```
techmannih: One test per file

tests/fixtures/extend-expect-circuit-snapshot.ts
```diff
@@ -49,6 +50,12 @@ async function saveSvgSnapshotOfCircuitJson({
     case "pinout":
       content = convertCircuitJsonToPinoutSvg(soup, options)
       break
+    case "solder-paste":
+      content = convertCircuitJsonToSolderPasteMask(soup, {
```
techmannih: No that's not correct I think

tests/fixtures/extend-expect-circuit-snapshot.ts
```diff
@@ -134,6 +141,34 @@ expect.extend({
         Boolean(process.env.BUN_FORCE_UPDATE_SNAPSHOTS),
     })
   },
+  async toMatchSolderPasteSnapshot(
```
techmannih: No, it should be in the pcb snapshots

tests/fixtures/extend-expect-circuit-snapshot.ts
```diff
@@ -134,6 +141,34 @@ expect.extend({
...
+    this: any,
+    received: unknown,
+    ...args: any[]
+  ): Promise<MatcherResult> {
```
techmannih: ??

techmannih (changes requested):
Lot of unrelated code

tests/components/primitive-components/smtpad-solder-paste-config.test.tsx
Hero988: Split into one test per file in [62a30be](https://github.com/tscircuit/core/pull/2945/commits/62a30bec6559ba05dcbd276d40fd66b8b04d15aa) — 7 numeric tests + 1 visual snapshot test, each in its own file.

tests/fixtures/extend-expect-circuit-snapshot.ts
Hero988: Removed in [62a30be](https://github.com/tscircuit/core/pull/2945/commits/62a30bec6559ba05dcbd276d40fd66b8b04d15aa) — `extend-expect-circuit-snapshot.ts` is reverted to main, so this PR no longer touches the snapshot fixture at all.

tests/fixtures/extend-expect-circuit-snapshot.ts
Hero988: Agreed — the matcher is removed and the visual test now uses the regular `toMatchPcbSnapshot`. Since the pcb view doesn't currently draw `pcb_solder_paste`, I opened tscircuit/circuit-to-svg#648 adding an opt-in `showSolderPaste` flag to `convertCircuitJsonToPcbSvg`; once that releases the pcb snapshot here can show the paste apertures directly.

tests/fixtures/extend-expect-circuit-snapshot.ts
Hero988: Gone in [62a30be](https://github.com/tscircuit/core/pull/2945/commits/62a30bec6559ba05dcbd276d40fd66b8b04d15aa).

Hero988:
All review points addressed in [62a30be](https://github.com/tscircuit/core/pull/2945/commits/62a30bec6559ba05dcbd276d40fd66b8b04d15aa): the copied solder-paste snapshot matcher is fully removed (`extend-expect-circuit-snapshot.ts` reverted to main), tests are split one per file, and the diff is now just `SmtPad.ts` + tests + one pcb snapshot. For showing paste in the pcb snapshots I opened tscircuit/circuit-to-svg#648 (opt-in `showSolderPaste` for the pcb view) — happy to wire it in here once it releases. The before/after paste-mask images in the PR body stay valid (pinned to the earlier commit).

techmannih (commented):
No solder mask appear on the pads?

Hero988:
> No solder mask appear on the pads?

Right — the snapshot shows bare pads because `convertCircuitJsonToPcbSvg` currently doesn't draw `pcb_solder_paste` at all: the four paste elements are in the circuit JSON, the pcb view just skips them. tscircuit/circuit-to-svg#648 fixes that with an opt-in `showSolderPaste` flag.

Here is this PR's snapshot board rendered with the #648 build — same circuit JSON, paste apertures now visible for all four regimes (default 0.7× / margin 0 / −0.2 / +0.2, left to right):

| current release (paste skipped) | with circuit-to-svg#648 `showSolderPaste: true` |
| --- | --- |
|  |  |

Once #648 merges and releases I'll bump it here and switch the test to `toMatchPcbSnapshot(import.meta.path, { showSolderPaste: true })` so the checked-in snapshot shows the paste directly.

Hero988:
Update: `showSolderPaste` landed in circuit-to-svg (#649, released in v0.0.396), so this PR no longer needs anything extra to show the paste, the standard pcb snapshot renders it.

- bumped `circuit-to-svg` to `^0.0.396`
- `smtpad-solder-paste-config.test.tsx` now asserts `toMatchPcbSnapshot(import.meta.path, { showSolderPaste: true })`, so the four `solderPasteMargin` cases are visible in the snapshot itself (4 `pcb_solder_paste` elements drawn)
- rebased onto current `main` (the branch was 34 commits behind; that staleness was also breaking `tsc` locally against the newly published deps — clean after the rebase)

| default render | `showSolderPaste: true` (this PR's snapshot) |
|---|---|
|  |  |
| paste elements exist in the circuit JSON but aren't drawn | left→right: default margin, `0`, `-0.2`, `0.2` |

@techmannih this is the "show it in the standard pcb snapshot" you asked for, now that the renderer supports it.

seveibar (approved):

PR #2948, opened 2026-07-31
Fix blob footprint URL handling

seveibar:
## Summary

- classify `blob:` footprint values as Blob URLs instead of footprint-library references
- route Blob URLs with `#ext=...` metadata through the configured footprint file parser
- ignore extensionless Blob URLs rather than emitting a false missing-library resolver error
- add visual regression coverage for both cases

## Root cause

tscircuit/core#2928 made missing footprint-library resolvers produce an `external_footprint_load_error`. `parseLibraryFootprintRef` only excluded HTTP(S) URLs, so eval's generated `blob:<uuid>` footprint was parsed as library `blob` and failed with "No footprint resolver is configured for library \"blob\"."

This blocked the automated dependency update in tscircuit/eval#3572 and therefore stopped eval's downstream release propagation.

## Impact

The current eval repro no longer reports a false external footprint error. Blob URLs that include an extension fragment can now be loaded by the matching platform file parser.

## Validation

- `bun test tests/footprint/blob-url-footprint-without-extension.test.tsx tests/footprint/blob-url-footprint-with-extension.test.tsx tests/footprint/footprint-library-map-missing-resolver-error.test.tsx`
- `bunx tsc --noEmit`
- `bun run build`
- `bun test tests/repros/create-blob-url-for-static-files-with-content.test.tsx --timeout 30000` in tscircuit/eval#3572 using this local core build


seveibar:
CC @techmannih

PR #2953, opened 2026-07-31
Emit decoupling metadata for tscircuit/checks

0hmX:
## Summary

This is a cross-package implementation of the chip-power-pin decoupling check:

- `@tscircuit/core` determines each chip pin's decoupling requirement while rendering JSX and serializes that intent into Circuit JSON.
- `@tscircuit/checks` reads the serialized metadata, evaluates circuit connectivity, and emits the warning.

This keeps component semantics in core and DRC warning ownership in checks.

## End-to-end workflow

1. Core copies explicit JSX `pinAttributes` onto the generated `source_port`.
2. Core resolves `should_have_decoupling_capacitor` using this precedence:
   - An explicit `shouldHaveDecouplingCapacitor` value wins.
   - `providesPower: true` resolves to `false` because a power-source output should not require a local decoupling capacitor.
   - An explicit `requiresPower` value resolves to the same boolean.
   - Otherwise, a conventional chip power label such as VCC or VDD resolves to `true`.
   - Unrelated pins remain unset so existing underspecified-pin diagnostics continue to work.
3. The result is present in Circuit JSON on the source port, for example:

```json
{
  "type": "source_port",
  "should_have_decoupling_capacitor": true,
  "recommended_decoupling_capacitor_capacitance": "100nF"
}
```

4. Checks builds a canonical source connectivity map and indexes source components, source ports, capacitors, and ground ports/nets.
5. For each connected `simple_chip` port that requires decoupling, checks looks for a two-pin `simple_capacitor` with:
   - one terminal connected to the chip's power rail; and
   - the other terminal connected to a ground port or ground net.
6. If that capacitor exists, no warning is produced. Otherwise checks emits a `source_pin_missing_trace_warning`, including the recommended capacitance in the message when supplied.
7. The check is exported and registered in `runAllNetlistChecks`.

## False-positive behavior

- `shouldHaveDecouplingCapacitor: false` is preserved in Circuit JSON and suppresses the warning.
- Pins with `providesPower: true` are serialized with the requirement set to `false`.
- Fully unconnected power pins are skipped here because the existing missing-connection check owns that diagnostic.
- Checks performs no VCC/VDD/GND label inference; it consumes domain metadata from Circuit JSON.
- For older Circuit JSON without the new field, checks falls back to `requires_power === true && provides_power !== true`.

## Cross-repository PRs

- Core metadata producer: https://github.com/tscircuit/core/pull/2953
- Checks warning consumer: https://github.com/tscircuit/checks/pull/175

Both changes are required for the complete JSX → Circuit JSON → connectivity check → warning workflow.

## Validation

- Core focused metadata and pin-specification regression tests
- Checks focused decoupling test and full test suite (133 tests)
- TypeScript typechecks in both repositories
- Checks build and formatting validation


seveibar (commented):
you should throw the warning in tscircuit/checks!

you may need to use tscircuit/core to add METADATA to the circuit json such that the warning can be thrown!

lib/components/primitive-components/Port/apply-default-decoupling-requirement-to-source-port.ts
```diff
@@ -0,0 +1,36 @@
...
+  }
+
+  if (
+    sourcePortLabels.some((sourcePortLabel) =>
+      POWER_NET_REGEX.test(sourcePortLabel),
+    )
+  ) {
+    sourcePortAttributes.should_have_decoupling_capacitor = true
+  }
+}
```
seveibar, line 36: can you move this stuff to capacitor

lib/components/primitive-components/Port/Port.ts
```diff
@@ -393,19 +394,24 @@ export class Port extends PrimitiveComponent<typeof portProps> {
...
 
     for (const attributes of pinAttributes) {
-      applyPinAttributesToSourcePort(portAttributesFromParent, attributes)
+      applyPinAttributesToSourcePort(sourcePortAttributes, attributes)
     }
+    applyDefaultDecouplingRequirementToSourcePort({
+      sourcePortAttributes,
+      sourcePortLabels: port_hints,
+      parentNormalComponentName: parentNormalComponent?.config.componentName,
+    })
```
seveibar, line 406: can it be moved to capacitor?

PR #2955, opened 2026-07-31
Automatically detect and constrain decoupling capacitor traces

0hmX:
## Summary

- detect capacitors connected between a qualifying chip power pin and ground
- infer power requirements from common labels such as `VCC` while honoring `requiresPower` and `shouldHaveDecouplingCapacitor: false`
- default detected decoupling capacitors and their attached traces to a 1 mm maximum length
- preserve explicit `maxDecouplingTraceLength` values and leave unrelated capacitors unchanged

This implements the trace-length portion of https://github.com/tscircuit/rfc/blob/main/rfcs/2026-07-31-automatic-decoupling-capacitor-detection-and-enforcement.md without adding a new warning or enforcement mechanism.

## Impact

Circuits matching the RFC connectivity pattern now populate the existing maximum trace-length metadata automatically, allowing the existing routing and DRC pipeline to enforce short decoupling paths.

## Validation

- `bunx tsc --noEmit`
- focused Biome checks
- RFC good-case and explicit opt-out tests
- existing explicit trace-length and crystal trace-length tests
- full suite: 1,301 passed; one unrelated differential-pair test exceeded its 5-second timeout under full-suite load and passed when rerun alone

### DIFF
--- a/lib/components/normal-components/apply-automatic-decoupling-trace-length.ts
+++ b/lib/components/normal-components/apply-automatic-decoupling-trace-length.ts
@@ -1,0 +1,123 @@
+import type { SourcePort } from "circuit-json"
+import {
+  GROUND_NET_REGEX,
+  POWER_NET_REGEX,
+} from "lib/utils/gnd-power-net-regex"
+import type { Capacitor } from "./Capacitor"
+
+const DEFAULT_MAX_DECOUPLING_TRACE_LENGTH_MM = 1
+
+const portHintsMatch = (sourcePort: SourcePort, pattern: RegExp): boolean =>
+  sourcePort.port_hints?.some((portHint) => pattern.test(portHint)) ?? false
+
+const chipPortShouldHaveDecouplingCapacitor = (
+  sourcePort: SourcePort,
+  capacitor: Capacitor,
+): boolean => {
+  if (!sourcePort.source_component_id) return false
+
+  const sourceComponent = capacitor.root!.db.source_component.get(
+    sourcePort.source_component_id,
+  )
+  if (sourceComponent?.ftype !== "simple_chip") return false
+
+  if (sourcePort.should_have_decoupling_capacitor !== undefined) {
+    return sourcePort.should_have_decoupling_capacitor
+  }
+
+  return (
+    sourcePort.requires_power ?? portHintsMatch(sourcePort, POWER_NET_REGEX)
+  )
+}
+
+const sourcePortIsGround = (sourcePort: SourcePort): boolean =>
+  sourcePort.provides_ground === true ||
+  sourcePort.requires_ground === true ||
+  portHintsMatch(sourcePort, GROUND_NET_REGEX)
+
+export const applyAutomaticDecouplingTraceLength = (
+  capacitor: Capacitor,
+): void => {
+  if (
+    capacitor._parsedProps.maxDecouplingTraceLength !== undefined ||
+    !capacitor.source_component_id
+  ) {
+    return
+  }
+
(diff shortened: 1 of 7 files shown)


lib/components/normal-components/apply-automatic-decoupling-trace-length.ts
```diff
@@ -0,0 +1,123 @@
+import type { SourcePort } from "circuit-json"
+import {
+  GROUND_NET_REGEX,
+  POWER_NET_REGEX,
+} from "lib/utils/gnd-power-net-regex"
+import type { Capacitor } from "./Capacitor"
+
+const DEFAULT_MAX_DECOUPLING_TRACE_LENGTH_MM = 1
+
+const portHintsMatch = (sourcePort: SourcePort, pattern: RegExp): boolean =>
```
0hmX: inline

lib/components/normal-components/apply-automatic-decoupling-trace-length.ts
```diff
@@ -0,0 +1,123 @@
...
+  sourcePort.port_hints?.some((portHint) => pattern.test(portHint)) ?? false
+
+const chipPortShouldHaveDecouplingCapacitor = (
```
0hmX: one function/class per files

lib/components/normal-components/apply-automatic-decoupling-trace-length.ts
```diff
@@ -0,0 +1,123 @@
...
+    sourcePort.requires_power ?? portHintsMatch(sourcePort, POWER_NET_REGEX)
+  )
+}
+
+const sourcePortIsGround = (sourcePort: SourcePort): boolean =>
+  sourcePort.provides_ground === true ||
+  sourcePort.requires_ground === true ||
+  portHintsMatch(sourcePort, GROUND_NET_REGEX)
+
+export const applyAutomaticDecouplingTraceLength = (
```
0hmX, line 15: lacking js doc do not explain type but what the function dose

lib/components/normal-components/apply-automatic-decoupling-trace-length.ts
```diff
@@ -0,0 +1,123 @@
...
+  const { db } = capacitor.root!
+  const allSourcePorts = db.source_port.list()
+  const capacitorSourcePorts = allSourcePorts.filter(
+    (sourcePort) =>
+      sourcePort.source_component_id === capacitor.source_component_id,
+  )
+
+  const chipPowerConnectivityMapKeys = new Set(
+    allSourcePorts
+      .filter(
```
0hmX: logs of filter map make me worry just make sure we are doing things smartly

lib/components/normal-components/apply-automatic-decoupling-trace-length.ts
```diff
@@ -0,0 +1,110 @@
+import type { SourcePort } from "circuit-json"
+import type { Capacitor } from "./Capacitor"
+import { chipSourcePortShouldHaveDecouplingCapacitor } from "./chip-source-port-should-have-decoupling-capacitor"
+import { sourcePortIsGround } from "./source-port-is-ground"
+
+const DEFAULT_MAX_DECOUPLING_TRACE_LENGTH_MM = 1
+
+type SubcircuitConnectivityMapKey = NonNullable<
+  SourcePort["subcircuit_connectivity_map_key"]
```
0hmX: if possible get the named type insted of deriving like this

seveibar (commented):
Not a good implementation for this, wrong render phase and inefficient db usage

PR #2956, opened 2026-07-31
Fix phased routing across transparent breakouts

seveibar:
## Issue

A transparent breakout splits each source trace into a breakout-local SRJ connection and a board-level continuation. A bus can therefore legitimately match more than one SRJ connection for the same source trace:

```tsx
<board width="20mm" height="12mm" layers={4}>
  <breakout
    name="SOURCE_BREAKOUT"
    width="6mm"
    height="6mm"
    padding="0.7mm"
    autorouter="sequential-trace"
  >
    <chip name="U1" footprint={/* two pads */} />
    <bus name="DATA" connections={["D0", "D1"]} />
  </breakout>

  <pinheader name="J1" pinCount={2} footprint="pinrow2_p2.54" />
  <trace name="D0" from=".U1 > .pin1" to=".J1 > .pin1" />
  <trace name="D1" from=".U1 > .pin2" to=".J1 > .pin2" />
</board>
```

Previously, bus conversion threw when a trace matched both connections. The board continuation could also treat the breakout segment as a foreign obstacle because the routed segment lost the root source-trace identity. Capacity routing may move a phase endpoint by a few microns, which then leaves the two segments geometrically disconnected.

## Fix

- Include every current-phase SRJ connection belonging to a bus source trace.
- Preserve `source_trace_id` on routed phase output and use it as an obstacle connectivity alias.
- Snap only near-identical routed endpoints (within 0.01 mm) back to their exact phase boundary point.

This does not add routing phases to buses or change the SRJ phase model.

## Tests

- Added a transparent-breakout bus regression covering both autorouting phases, bus membership, obstacle connectivity, source-trace identity, DRC, and PCB snapshot.
- Updated the existing autoplaced breakout snapshot for exact endpoint preservation.
- `bunx tsc --noEmit`
- `bun run build`
- Focused breakout and phased-routing tests: 7 passed, 0 failed.


seveibar:
Superseded by #2968, which is rebased on current main and contains only the focused manual-breakout autorouter selection, split-bus mapping, and regression snapshots.

PR #2962, opened 2026-08-01
Detect and constrain decoupling capacitor traces during source render

seveibar:
## Summary

- Automatically detect two-pin capacitors whose power-side port directly shares a trace with an eligible chip power-input pin and whose other port connects to ground.
- Apply the existing 1 mm default decoupling trace-length constraint before source traces render.
- Accept a direct ground port or named ground net, but do not infer decoupling through a shared power net.
- Honor `requiresPower`, inferred common power-pin labels, and explicit `shouldHaveDecouplingCapacitor` overrides.
- Preserve explicit `maxDecouplingTraceLength` values and leave unrelated traces unconstrained.

## Architecture

Detection is owned by `Capacitor.doInitialSourceRender()`. At that point, `CreateTracesFromProps` has created connectivity instances, while `SourceTraceRender` has not yet run. The capacitor asks its two `Port` instances whether they are connected to power or ground using `isConnectedToPower()` and `isConnectedToGround()`.

Power connectivity is intentionally strong: the capacitor port and eligible chip power pin must occur on the same trace. Separate traces that merely terminate at the same `net.VCC` do not qualify. Ground connectivity may terminate directly at a ground port or ground net.

The capacitor source component is inserted with the inferred `max_decoupling_trace_length`, and the existing trace render phase propagates that value to attached `source_trace.max_length` records. `Group` remains unchanged, and unresolved trace selectors continue through the normal trace diagnostic flow.

This is an alternative implementation to #2955 and implements the behavior described in the [automatic decoupling capacitor detection RFC](https://github.com/tscircuit/rfc/blob/main/rfcs/2026-07-31-automatic-decoupling-capacitor-detection-and-enforcement.md).

## Impact

Qualifying decoupling capacitors receive `max_decoupling_trace_length: 1` by default, and their attached traces receive `max_length: 1`. Capacitors connected only between shared power and ground nets are not automatically designated.

## Validation

- `bunx tsc --noEmit`
- `bun run build`
- focused direct-connectivity, shared-net exclusion, opt-out, and existing trace-length regression tests
- full suite: 1305 passed, 31 skipped, 0 failed


seveibar:
@0hmX this feels more clear to me wdyt?

astrimid:
Just wanted to drop some feedback from a user perspective.

Firstly, the motivation behind this makes total sense. Not specifying decoupling trace-length constraints leads to capacitor being useless, so catching that makes sense.

OTOH, as a recent tscircuit user, I often feel I'm fighting the placement/routing heuristics, with no high level language to communicate intent.

I wanted to share a concern that applying physical constraints automatically via netlist topology might do more harm than good.

1. Capacitor intent ambiguity

To the flattened netlist, a 100nF ceramic decoupling cap and a 220µF electrolytic bulk cap share the exact same topology (both sit between a power rail and GND, both connect a power pin and GND, not even talking about multiple chips/pins and filter networks). If max_length: 1 is silently injected onto every power-to-ground cap, the autorouter will try to cram massive bulk caps 1mm away from ICs, which will either fail the solver or create really awkward physical layouts.

2. Debugging silent "helper"

If the autorouter fails or produce strange results because of an auto-injected constraint, it might be hard for the user to debug because the 1mm constraint isn't actually visible in their source TSX/React code.

**Warning instead of autoconstraint**

I agree that autodectection is usefult in general. But I don't agree that it should lead to mutating the original TSX. I'd rather have this logic to product a DRC warning or what I call "placement advisor".

Here's the full spec if you're interested: #2964

0hmX (approved):

PR #2970, opened 2026-08-02
fix: treat pcbOffsetX/pcbOffsetY as aliases for pcbX/pcbY

Hero988:
`pcbOffsetX` / `pcbOffsetY` are declared on `pcbLayoutProps` and `commonLayoutProps` in `@tscircuit/props` (so every component accepts them, and they type-check), but the strings appear zero times in `core/lib` — they are parsed and then dropped, so setting them does nothing.

They are aliases for `pcbX`/`pcbY`, so this normalizes them onto `pcbX`/`pcbY` immediately after props are parsed. Every downstream consumer of a parsed pcb position — placement, edge anchors, calc references, manual-placement conflict detection — keeps reading only `pcbX`/`pcbY`. `pcbX`/`pcbY` win if both an alias and its canonical prop are set.

### Before / after

```tsx
<board width="30mm" height="30mm" routingDisabled>
  <resistor name="R1" resistance="1k" footprint="0402" pcbX={5} pcbY={3} />
  <resistor name="R2" resistance="1k" footprint="0402" pcbOffsetX={5} pcbOffsetY={-3} />
  <resistor name="R3" resistance="1k" footprint="0402" pcbOffsetX="-6mm" pcbOffsetY={0} />
</board>
```

| before (`main`) | after |
|---|---|
|  |  |
| R2 `(0, 0)`, R3 auto-packed to `(2.86, 0)` — both props ignored | R2 `(5, -3)`, R3 `(-6, 0)` |

### Changes

- `lib/utils/pcb/apply-pcb-position-prop-aliases.ts` (new, 17 lines) — maps `pcbOffsetX`/`pcbOffsetY` onto `pcbX`/`pcbY` when the canonical prop is unset
- `PrimitiveComponent` — applies it in the constructor and in `setProps`, the two places parsed props are produced

### Tests

- `tests/components/pcb/pcb-offset-x-y.test.tsx` — alias positions a component exactly like `pcbX`/`pcbY`, including a string distance (`"-6mm"`), with a standard pcb snapshot
- `tests/components/pcb/pcb-offset-alias-precedence.test.tsx` — `pcbX`/`pcbY` win when both are set

Full local suite: 1304 pass / 8 fail, all 8 pre-existing 5s-timeout flakes on this machine (autorouting, RP2040, differential-pair) that also fail on unmodified `main`.


seveibar (changes requested):
really really bad implementation, pcbOffsetX is an alias to pcbX

Hero988:
Thanks for being patient with me, reworked to alias semantics in 8d022dc8.

`pcbOffsetX`/`pcbOffsetY` are now normalized onto `pcbX`/`pcbY` right after props are parsed (`lib/utils/pcb/apply-pcb-position-prop-aliases.ts`, applied in `PrimitiveComponent`'s constructor and `setProps`), so placement, edge anchors, calc references and manual-placement conflict detection all keep reading a single canonical prop. `pcbX`/`pcbY` take precedence when both are set.

The additive-offset code in `NormalComponent` and `Group_doInitialPcbCalcPlacementResolution` is gone, along with its test the diff is now 2 lib files (+19 lines) and 2 tests. Body updated with the new before/after.

seveibar (changes requested):
violates our rule that _parsedProps cannot be modified inside a constructor

seveibar:
too many violations

PR #3011, opened 2026-08-05
Warn when differential pairs are not point to point

0hmX:
Implements [tscircuit/rfc#13](https://github.com/tscircuit/rfc/pull/13).

## End-to-end workflow

1. Core parses `<differentialpair>` and retains its `name`, `positiveConnection`, and `negativeConnection` in the component tree.
2. Core renders the subcircuit's source traces, source ports, source nets, and connectivity-map keys.
3. During `SourceDesignRuleChecks`, the differential-pair component validates the positive and negative connections independently.
4. For each connection, Core resolves a named trace or port selector to one subcircuit connectivity group.
5. Core collects the terminal source ports in that connectivity group.
6. Exactly two terminal ports is valid and produces no diagnostic.
7. Any other terminal count stores a `source_property_ignored_warning` in Core's Circuit JSON database. The warning identifies the pair, property, resolved net or selector, terminal pins, and suggested correction.
8. When the component updates or is removed, Core deletes the warnings previously owned by that differential pair before recalculating, preventing stale or duplicate warnings.
9. `circuit.getCircuitJson()` serializes the warning with the rest of the circuit.
10. A follow-up CLI change will classify and display this stored diagnostic when a user runs `tsci check netlist <file>`.


tests/components/primitive-components/differential-pair/failure-tests/port-selector-multiple-traces.test.tsx
```diff
@@ -2,7 +2,7 @@ import { expect, it } from "bun:test"
 import { getSimpleRouteJsonFromCircuitJson } from "lib/utils/autorouting/getSimpleRouteJsonFromCircuitJson"
 import { getTestFixture } from "tests/fixtures/get-test-fixture"
 
-it("throws when a differential pair port selector matches multiple traces", (): void => {
+it("stores a warning without changing the existing autorouter error", (): void => {
```
0hmX: why can this not be a new test in a separate files

lib/components/primitive-components/DifferentialPair.ts
```diff
@@ -3,17 +3,35 @@ import {
...
+    DifferentialPair_doInitialSourceDesignRuleChecks(this)
+  }
+
+  updateSourceDesignRuleChecks(): void {
+    DifferentialPair_doInitialSourceDesignRuleChecks(this)
+  }
+
+  removeSourceDesignRuleChecks(): void {
+    DifferentialPair_removeSourceDesignRuleChecks(this)
+  }
```
seveibar: why define all of these? Just define doInitial*

seveibar (approved):
see comments

PR #3014, opened 2026-08-05
Fix four-pin crystal trace length propagation

MustafaMulla29:
## Summary

- expose semantic `X1` and `X2` labels on crystal signal ports
- propagate a crystal's maximum trace length through those typed semantic ports
- keep GND-connected crystal ports and the wider ground network unconstrained
- add a routed regression test using the standard `footprint="crystal"`

## Root cause

`Crystal.doInitialSourceDesignRuleChecks` collected every source port belonging to a crystal. For a four-pin crystal, that included the two case/ground ports. Because those ports shared the board-wide GND connectivity map, the crystal's default 10 mm maximum trace length propagated to unrelated GND traces.

The autorouter's precheck then rejected otherwise valid boards whenever distant GND endpoints were more than 10 mm apart.

## Fix

Crystal signal ports carry the semantic labels `X1` and `X2`, included in the component's typed port map. Trace-length propagation now starts directly from `this.portMap.X1` and `this.portMap.X2`. It does not inspect physical pin numbers or scan raw port-hint strings.

Physical pin-to-role mapping remains only in `initPorts`, where a component's package pinout belongs. The design-rule behavior depends solely on the semantic signal roles.

## Impact

RP2040 and other designs using four-pin crystals can route normally without adding an artificial large `maxLength` workaround. The oscillator traces still retain the intended short-trace constraint.

## Validation

- focused four-pin crystal routing regression: pass
- all related crystal tests: 7 pass
- TypeScript `--noEmit`: pass
- Biome check: pass
- regression connects through `X1`/`X2`, asserts exactly 4 PCB traces, and asserts zero `pcb_autorouting_error` records

### DIFF
--- a/lib/components/normal-components/Crystal.ts
+++ b/lib/components/normal-components/Crystal.ts
@@ -8,14 +8,13 @@ import { formatSiUnit } from "format-si-unit"
 import { type BaseSymbolName, type Ftype } from "lib/utils/constants"
 import { NormalComponent } from "../base-components/NormalComponent/NormalComponent"
 
-type CrystalPorts = CrystalPinLabels
+type CrystalPorts = CrystalPinLabels | "X1" | "X2"
 type SourcePortId = SourcePort["source_port_id"]
 type SubcircuitConnectivityMapKey = NonNullable<
   SourceTrace["subcircuit_connectivity_map_key"]
 >
 
 const DEFAULT_CRYSTAL_MAX_TRACE_LENGTH_MM = 10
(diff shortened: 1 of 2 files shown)


seveibar (changes requested):
Fragile, should be based on pin attributes or labels, X1/X2 have w length requirement, not GND. Lots of ways to check that dont hack PIN numbers

MustafaMulla29:
@seveibar is this correct now? or we need to add pinlabels to props?

seveibar (approved):

PR #3016, opened 2026-08-05
Use canonical net label text for schematic solver bounds

MustafaMulla29:
## Problem

Core could tell the schematic trace solver that a net label had one name and width, then render a different canonical label for the same connection.

The RP2040 repro makes this visible on the shared QSPI connection. The rendered labels use `QSPI_SS`, but the solver input previously measured local trace names such as `BOOT_SW` and `BOOT_R`, and also used a different font-size assumption. For example, the solver received widths of `0.75` or `0.65` where the rendered `QSPI_SS` label is `0.96` wide. This meant routing decisions were made against undersized or incorrectly named obstacles, and exported solver bug reports did not faithfully represent the final core schematic.

## Root cause

Schematic label text was resolved independently in `createSchematicTraceSolverInputProblem` and `applyNetLabelPlacements`. Those paths had different precedence rules for source-net names, explicit display labels, trace names, and source-port labels, so solver input and rendered output could diverge.

## Fix

- Add a connection-keyed canonical schematic net-label resolver.
- Reuse that resolver when creating solver input and when applying solver label placements.
- Resolve source-net names and explicit trace display labels consistently before falling back to trace or source-port names.
- Measure solver label bounds using the same canonical text and typography used by the renderer.
- Assert the corrected `QSPI_SS`, `BOOT_SW`, and `BOOT_R` solver widths in the complete RP2040 repro.
- Update only the schematic snapshots whose canonical label bounds or placement changed.

## Impact

The schematic trace solver now receives obstacle dimensions for the labels core actually renders. This keeps standalone solver reproductions aligned with core output and prevents routing based on stale, smaller label bounds.

## Validation

- `bun test` for the five affected schematic snapshot tests — 5 passed
- `bunx tsc --noEmit`
- `bunx biome check` for the changed TypeScript files

## Follow-up

The corrected RP2040 input exposes a separate downstream solver issue: the V1V1 trace between RP2040 pins 45 and 50 can be returned through the `ADC_VREF` label after a failed detour. This PR intentionally does not add a core-side workaround for that trace-solver behavior.


tests/repros/repro170-rp2040-u1-complete-connections.test.tsx
```diff
@@ -434,5 +442,17 @@ test("repro170: complete RP2040 U1 schematic connections", async () => {
...
+  expect(
+    schematicTraceInputProblem?.directConnections
+      .filter(({ netId }) =>
+        ["QSPI_SS", "BOOT_SW", "BOOT_R"].includes(netId ?? ""),
+      )
+      .map(({ netId, netLabelWidth }) => ({ netId, netLabelWidth })),
+  ).toEqual([
+    { netId: "QSPI_SS", netLabelWidth: 0.96 },
+    { netId: "BOOT_SW", netLabelWidth: 0.96 },
+    { netId: "BOOT_R", netLabelWidth: 0.96 },
```
imrishabh18: Is `0.96` some kind of min width? Cause `BOOT_R` should have a width less compared to other's as they have more characters

MustafaMulla29: `0.96` isn’t a minimum width. These IDs represent separate direct connections(all are trace names), but they all share the `QSPI_SS` net and render the canonical `QSPI_SS` label. Therefore, each solver entry must use that label’s `0.96` width.

seveibar (approved):

PR #3018, opened 2026-08-05
Add dense hierarchical autorouting failure repro

Abse2001:
## What changed

Adds a dense Game Boy-like expected-failure repro for routing from a parent board directly into an already-routed child MCU subcircuit, without pinheaders or imported component files.

The repro uses a bounded progress watchdog after eight consecutive parent `portPointPathingSolver` cycles, allowing CI to record the routing failure without reaching Bun's test timeout.

The former minimal different-net overlap repro has moved to #3044. Its flat control was corrected so it succeeds before the nested case reaches the intended crossing snapshot.

## Why

The dense case reaches `portPointPathingSolver` after preserving more than 30 child traces, then fails to advance toward its ten parent connections. It remains useful for investigating direct parent-to-subcircuit routing without a breakout boundary.

The watchdog throws through the autorouter's existing error path, stops its timer, and leaves the expected `pcb_autorouting_error` for the final desired-behavior assertion.

## Impact

This test retains realistic failure coverage for improving the child-to-parent autorouting handoff without depending on JLCPCB imports or application-specific component files.

## Validation

- `bun test tests/repros/repro-gameboy-direct-subcircuit-routing.test.tsx`
- `bunx tsc --noEmit`
- `git diff --check`


seveibar (approved):

Abse2001:
Closing this repro-only PR because the actionable breakout routing fix and its regression coverage were merged in #3044.

PR #3025, opened 2026-08-06
Keep implicit schematic rotations available to Matchpack

mohan-bee:
motivation: here the led position is locked with 270deg only without any rotation specified explicitly 

image


tests/repros/repro161-power-section-autolayout/__snapshots__/repro161-power-section-autolayout-schematic.snap.svg
mohan-bee, line 1: will try to fix in matchpack

lib/components/primitive-components/Group/applySchematicMatchPackLayoutToTree.ts
```diff
@@ -461,9 +461,12 @@ function convertTreeToMatchPackInputProblem(
...
 
-      // A power/ground 2-pin part is locked to the single rotation that places
-      // its rail pin on the correct side (power up, ground down).
-      if (availableRotations.length === DEFAULT_AVAILABLE_ROTATIONS.length) {
+      // Non-LED power/ground parts are locked to face their rail. LEDs remain
+      // rotatable so series-branch layout can align them with adjacent parts.
+      if (
+        availableRotations.length === DEFAULT_AVAILABLE_ROTATIONS.length &&
+        child.sourceComponent.ftype !== "simple_led"
+      ) {
```
imrishabh18: I don't understand the reason for only led. What happens to other components like diodes, capacitors or resistors?

mohan-bee: agree

MustafaMulla29 (changes requested):
@mohan-bee  this is intentional, any component that connects to power/ground its rotation is locked to 270 deg.

MustafaMulla29:
I was the one who did it(which was requested by seve)

MustafaMulla29:
@mohan-bee i think you should close this pr, this doesnt solve anything

mohan-bee:
> @mohan-bee i think you should close this pr, this doesnt solve anything

but in this case the led is supposed to be placed in horizontal !! 

image

mohan-bee:
@MustafaMulla29  i think. we need to support rotation for shared net also

MustafaMulla29:
> > @mohan-bee i think you should close this pr, this doesnt solve anything
> 
> but in this case the led is supposed to be placed in horizontal !!
> 
> image

I dont think so

PR #3026, opened 2026-08-06
fix: keepout with no layers prop guards every copper layer

### LINKED ISSUE #2613: keepout silently defaults to layers=["top"] - bottom-layer routing ignores it

zkasuran:
Closes #2613

`<keepout>` with no `layer` or `layers` prop defaulted to `["top"]`, so on a 2-layer board the autorouter was free to route through the keepout on `bottom`. That is the case in the issue: a keepout paired with a routed slot held the router off on top while the net was routed along the slot on the bottom, inside the copper-to-edge clearance, with nothing to say the guard was only half active.

Nothing sets that default outside core. `layer` and `layers` are both optional in `pcbKeepoutProps`, so the `["top"]` fallback in `Keepout.doInitialPcbPrimitiveRender()` was the whole behavior.

## The change

The fallback now covers every copper layer the board has:

```ts
layers = getBoardAvailableLayers(subcircuit._getSubcircuitLayerCount())
```

`getBoardAvailableLayers` is the helper `Board.allLayers` already uses. `_getSubcircuitLayerCount()` is the same call `Via.ts` makes, so the layer count keeps one source of truth. A 2-layer board gets `["top", "bottom"]`, a `layers={4}` board gets all four copper layers. KiCad rule areas default to every copper layer too, which is the precedent the issue cites.

Precedence is untouched: explicit `layers` wins, then `layer`, then this default.

`tests/components/primitive-components/keepout-default-layers-bottom-routing.test.tsx` is the reported situation reduced to two bottom-side resistors either side of an 8mm x 4mm keepout with no `layers` prop. It collects the routed wire segments and asserts none of them cross the keepout rectangle. On `main`:

```
  expect(crossingWireSegments).toEqual([])
error: expect(received).toEqual(expected)

- []
+ [
+   [
+     { "layer": "bottom", "route_type": "wire", "width": 0.15, "x": -7.9798623750482625, "y": 0.530137624951737 },
+     { "layer": "bottom", "route_type": "wire", "width": 0.15, "x": 7.9798623750482625, "y": 0.530137624951737 },
+   ],
+ ]
(fail) keepout without layers blocks bottom layer routing
```

One wire on `bottom`, from x = -7.98 to x = +7.98 at y = 0.53, straight through a keepout spanning x = -4..4 and y = -2..2. With the change the router goes around it at y = -2.21 and the test passes.

That check is a segment against axis-aligned rectangle test rather than point-in-rectangle, because both endpoints of the offending wire sit outside the keepout while the wire crosses the full width of it. A point-only check would have passed on the broken board.

`tests/components/primitive-components/keepout-default-layers-four-layer-board.test.tsx` covers the layer-count half of the default: on a `layers={4}` board a keepout with no `layers` prop comes out as `["top", "bottom", "inner1", "inner2"]`, while a sibling keepout with `layers={["inner1"]}` is left exactly as written. It fails on `main` with `layers: ["top"]`.

## Snapshot churn (3 files)

`keepout-pcb.snap.svg`, `keepout-get-pcb-size-pcb.snap.svg` and `keepout-exclude-refs-pcb.snap.svg`. circuit-to-svg draws a keepout once per layer it names, so each keepout in those tests gains a second identical shape tagged `data-pcb-layer="bottom"`. Same geometry, same centers, nothing else in the SVG moves. I read the diffs rather than trusting the updater.

`tests/features/ten-layer-forced-autorouting.test.tsx` passes explicit `layers` to all ten of its keepouts, so it never used the default and it is unchanged.

## The behavior change, said plainly

A board that leaned on the top-only default now has its bottom and inner layers guarded as well. A route that used to slip through a keepout will detour. On a board with no way around, it becomes an autorouting failure instead of silent copper where the author asked for none, which is the trade the issue asks for. Anyone who wanted top-only writes `layers={["top"]}` and gets the old behavior back.

## Verification

- `bun test tests/components/primitive-components/keepout`: `6 pass / 0 fail`, 16 expect() calls.
- `bun test` (whole suite): `1334 pass / 37 skip / 10 fail`, 1381 tests across 1131 files. All 10 failures are timeouts rather than assertions. None of them involve keepouts. Seven hit bun's default 5s limit, which CI does not use (`bun-test.yml` runs `bun test --timeout 30000`). The other three are the slow layout repros, repro159 at its own 90s limit, repro116 at 80s and repro156 at 30s, on a machine that was also running a typecheck. Re-run with `--timeout 180000` all ten pass, in two batches: `8 pass / 0 fail` then `2 pass / 0 fail`.
- `bunx @biomejs/biome format .`: `Checked 1658 files in 147ms. No fixes applied.`

AI assistance (Claude, Anthropic) was used in developing this change. The design, review and verification were done by the author. Verified locally before submitting: the gates above.


zkasuran:
A nudge before this goes stale, it has had no review yet. All checks are green: format, type-check, smoke-test and all 10 test shards pass.

The change makes a `<keepout>` with no explicit layer guard every copper layer of the board, the way KiCad rule areas default, instead of silently defaulting to the top layer and letting the router cross it on the bottom of a two-layer board. Two tests cover it, including the reported symptom of a bottom-layer route crossing the keepout end to end, plus three refreshed keepout snapshots.

Happy to adjust if you want any change.

zkasuran:
A quick re-verification before this goes quiet, since it has had no review yet. At the current head bb711e16 all 16 CI checks are green: type-check, the 10 test shards, smoke-test, format-check and dependency-check. mergeStateStatus is CLEAN with no conflicts against main. The change stays scoped to making a keepout with no `layers` prop guard every copper layer, plus its regression test. Happy to adjust anything on review.

PR #3034, opened 2026-08-07
fix: report empty connections targets instead of a raw css-what crash

### LINKED ISSUE #2865: Empty string in `connections` crashes the render with a raw css-what parser error

zkasuran:
Closes #2865

An empty string in a `connections` prop took the whole render down with a raw CSS parser error that named neither the component nor the pin:

```tsx
  <capacitor
    name="C1"
    capacitance="1uF"
    footprint="0402"
  />
</board>
```

```
error: Expected name, found .
  at selectOne (lib/components/base-components/PrimitiveComponent/PrimitiveComponent.ts:1257:18)
```

Two things were wrong. The message is not actionable: `Expected name, found .` is css-what internals. Reading it, the natural guess is that `.R1 > .pin1`, the one target that is actually fine, is malformed. It is also a thrown exception rather than a recorded error, so the sibling valid connection, the other components and every later phase never run. One empty string produces zero output.

## Cause

`NormalComponent._createTracesFromConnectionsProp()` forwarded every entry of `props.connections` straight into a `Trace`'s `to`. An empty string is not a selector, but nothing rejected it, so `to: ""` reached `selectOne` during `SourceTraceRender`. `resolveImplicitSinglePort` prepends a `.` when the selector has no `.`/`#`/`[`, so css-what ends up parsing `"."` and throws the `Expected name, found .` several phases below the prop the user actually wrote.

The value survives prop validation too: `connections` is typed as a record of selector strings and `""` is a string, so the schema accepts it.

## The change

`_createTracesFromConnectionsProp` now skips empty and whitespace-only targets and records a `source_invalid_component_property_error` that names the component and the pin:

```
Invalid connections prop on capacitor "C1": the target for pin "pin1" is empty. Set it to a selector such as ".R1 > .pin1" or "net.GND". Remove the pin from connections if it is unconnected.
```

This mirrors the existing `_insertInvalidFootprintPropErrors` path in the same class. The message is queued while the traces are created (`CreateTracesFromProps` phase) and inserted during `SourceComponentPropertyValidation`, because `source_component_id` does not exist yet at the earlier phase. The error opens with `Invalid connections prop on ...` to match the sibling `Invalid footprint prop on ...` message.

After the change the board renders, both components are present and the valid `pin2` connection still produces its `source_trace`.

`tests/components/normal-components/connections-empty-target-error.test.tsx` is the issue reduced to two components with `pin1: ""`. On `main` it throws inside `renderUntilSettled`:

```
error: Expected name, found .
  at selectOne (lib/components/base-components/PrimitiveComponent/PrimitiveComponent.ts:1257:18)
(fail) empty connections target is reported as an invalid prop instead of crashing the render
 0 pass
 1 fail
```

With the change it records the clear error, keeps both components and keeps the sibling trace. The PCB snapshot shows the board rendered:

```
 1 pass
 0 fail
```

`tests/components/normal-components/connections-whitespace-target-error.test.tsx` covers the other two shapes the guard has to handle: a whitespace-only target (`pin1: "   "`) and an empty entry inside an array target (`pin2: [".R1 > .pin1", ""]`). Both are reported and the valid array entry still connects.

## Scope

- Only empty and whitespace-only targets are treated as invalid. A non-empty but unresolvable selector (for example `.R9 > .pin1` where R9 does not exist) already has its own `source_trace_not_connected_error` handling and is left untouched.
- The check sits in `_createTracesFromConnectionsProp`, so it covers both the string form and the array form of a `connections` entry.
- Empty targets never reached net creation (`createNetsFromProps` only acts on `net.` strings), so that path needed no change.
- Rejecting `""` in `tscircuit/props` would catch it earlier, but that is a cross-repo change with a wider blast radius. I kept the fix in core. Happy to follow up in props if you prefer that.

## Regression risk

Low by construction. The only input whose behavior changes is one that previously threw, so no passing test could have depended on the old behavior. The targeted `bun test connections trace` run is green (137 pass, 0 fail).

## Verification

- `bun test tests/components/normal-components/connections-empty-target-error.test.tsx tests/components/normal-components/connections-whitespace-target-error.test.tsx`: 2 pass / 0 fail.
- `bun test connections trace`: 137 pass / 5 skip / 0 fail.
- `bun test` (whole suite): 1348 pass / 37 skip / 2 fail across 1387 tests. Both failures are in `tests/breakout` (`fanout-soic8-sensor-to-i2c-header`, `breakout-sot23-regulator-power-rail`). They come from the external capacity-autorouter (`Failed to compute first trace`) and reproduce on a clean `main` with this change stashed, so they are unrelated to it. Nothing else fails.
- `bunx @biomejs/biome format .`: `Checked 1667 files in 138ms. No fixes applied.`

AI assistance (Claude, Anthropic) was used in developing this change. The design, review and verification were done by the author. Verified locally before submitting: the gates above. I own the change and can explain it and answer questions.


zkasuran:
Note on CI: the red test shards are two autorouter tests, `tests/breakout/fanout-soic8-sensor-to-i2c-header` and `tests/breakout/breakout-sot23-regulator-power-rail`, both failing with `Failed to compute first trace` from capacity-autorouter. They already fail on current `main` (the two were red on #2962's own shards 7 and 10 when it merged) and they reproduce locally with this change stashed. The remaining red shards are fail-fast cancellations rather than test failures.

This change only guards empty `connections` targets before they reach `selectOne`, so it does not touch routing. The rest of the local suite is green: 1348 pass, with only those two pre-existing autorouter failures.

zkasuran:
A quick re-verification before this goes quiet, no review yet. At the current head 6e05ebe2 all 16 CI checks are green (type-check, the 10 test shards, smoke-test, format-check, dependency-check) and mergeStateStatus is CLEAN with no conflicts. The fix turns a raw css-what crash on an empty connections target into a clear reported error, with a regression test. Open to any changes on review.

PR #3037, opened 2026-08-07
fix: populate inductor max_current_rating from maxCurrentRating

### LINKED ISSUE #2837: inductor drops maxCurrentRating: prop and schema field both exist, value never written

zkasuran:
Closes #2837

`maxCurrentRating` is declared on `InductorProps` and `max_current_rating` exists on the `simple_inductor` source schema, but `Inductor.doInitialSourceRender()` never wrote it. The value was accepted at the prop layer and then dropped, with no warning.

```tsx
<board width="12mm" height="10mm">
  <inductor name="L2" inductance="10uH" maxCurrentRating="2A" footprint="0402" />
</board>
```

The emitted `source_component` carried `name`, `inductance` and `display_inductance` but no `max_current_rating` anywhere.

## Cause

`doInitialSourceRender()` builds the `source_component` insert by hand and simply had no `max_current_rating` key, so the prop was never forwarded.

The value also needs converting on the way through. Capacitor's `maxVoltageRating` runs through a `voltage` zod transform in props, so `_parsedProps.maxVoltageRating` is already a number by the time the component writes it. The inductor prop is `z.union([z.string(), z.number()]).optional()` with no transform, so a raw string like `"2A"` or `"500mA"` reaches the component untouched. `parseFloat("500mA")` returns 500, which is 1000x off, so a naive parse would write a plainly wrong number into the circuit JSON.

## The change

`Inductor.ts` gains a small `_getMaxCurrentRating()` helper and writes its result into the insert. It parses with `parseAndConvertSiUnit(value, "A")`, the same call circuit-json's own `current` schema uses, so the unit prefix is honoured:

```
"2A" -> 2    "500mA" -> 0.5    "1.5A" -> 1.5    2 -> 2
```

A missing rating, a blank string or an unparseable value all resolve to `undefined`, so the field stays absent rather than being written as `null` or `NaN`. The blank-string guard matters because `parseAndConvertSiUnit("")` throws. An empty `maxCurrentRating` should be ignored the way it was before this change, not crash the render.

`tests/components/normal-components/inductor-max-current-rating.test.tsx` is the issue's exact repro. On `main` it fails:

```
expect(inductors[0].max_current_rating).toBe(2)
Expected: 2
Received: undefined
(fail) inductor writes max_current_rating from a string prop
 0 pass
 1 fail
```

With the change it reads back `2`.

`tests/components/normal-components/inductor-max-current-rating2.test.tsx` locks in the parsing: `"500mA"` becomes `0.5` (not `500`) and a numeric `3` passes straight through. An inductor with no rating leaves the field unset. Both files render a schematic snapshot so the inductors are visible.

## Scope

- Only `Inductor.ts` changes on the library side. The prop layer and the schema already had the field, so nothing there moved.
- `inductance` is still written as-is, exactly as before. This change only adds the missing `max_current_rating` key.
- I kept the parse in core rather than adding a zod transform to `maxCurrentRating` in `@tscircuit/props`. A transform would be the tidier long-term shape, but it is a cross-repo change with a wider blast radius. Happy to follow up in props if you prefer that.

## Regression risk

Low. The only observable change is that a populated `maxCurrentRating` now appears in the output where it used to be silently dropped. No existing inductor snapshot moved, because the rating is not drawn on the schematic or PCB. `bun test inductor` stays green.

## Verification

- `bun test tests/components/normal-components/inductor-max-current-rating.test.tsx tests/components/normal-components/inductor-max-current-rating2.test.tsx`: 2 pass / 0 fail, 2 snapshots.
- `bun test inductor`: 4 pass / 0 fail.
- `bun test` (whole suite): 1346 pass / 37 skip / 4 fail across 1387 tests. None of the 4 come from this change. Two are `tests/breakout` (`fanout-soic8-sensor-to-i2c-header`, `breakout-sot23-regulator-power-rail`). They regressed on the current `main` tip: they pass at `fcede5a5` and fail at `0559b381` (#2962, which changed decoupling-capacitor trace handling in source render). Stashing this diff and running them on a clean `0559b381` reproduces the same failure, so they are pre-existing here and unrelated to the inductor. The other two are heavy autorouter repros (`repro156-ti-multisheet-autolayout`, `repro116-arduino-uno-reroute`) that exceeded their per-test timeout under full-suite CPU load. Both pass when rerun on their own.
- `bunx @biomejs/biome format .`: `Checked 1667 files. No fixes applied.`


zkasuran:
The two red checks (`test (7)` and `test (10)`) are a pre-existing failure on current `main`, not from this change. The failing tests are `tests/breakout/breakout-sot23-regulator-power-rail` and `tests/breakout/fanout-soic8-sensor-to-i2c-header`.

Bisected locally against the same `bun test --timeout 30000` CI runs:

- pass at `5ec6a6d0` and at `fcede5a5`
- fail at `0559b381` (#2962, the current `main` tip)

#2962 changed decoupling-capacitor trace handling in source render, which drops the routed `pcb_trace` count in those two breakout tests below the asserted threshold (`toBeGreaterThanOrEqual(7)` gets `0`). Stashing this PR's diff and running them on a clean `0559b381` reproduces the same `0 pass / 2 fail`, so the inductor change is not involved.

This PR only touches `Inductor.ts` and its two new tests, which pass. type-check, format-check and smoke-test are green.

imrishabh18 (changes requested):
Not the right repo for fix!

zkasuran:
Closing this. Agreed it is not the right repo. Refocusing on board-driven fixes, thanks.

PR #3039, opened 2026-08-07
fix: emit schema-valid circuit JSON for board-level holes and DIP plated holes

### LINKED ISSUE #2843: board-level holes and DIP plated holes emit circuit JSON that fails any_circuit_element.parse()

zkasuran:
Closes #2843

Two board setups emit circuit JSON that fails `any_circuit_element.parse()`, the circuit-json schema parser. The board renders fine, but the emitted JSON is not schema-valid, so anything that validates it throws. core itself runs that parser in `DrcCheck.ts`.

```tsx
circuit.add(
    <hole name="H1" pcbX={-5} pcbY={5} diameter="1mm" />
    <chip name="U1" footprint="dip8" pcbX={-8} />
  </board>,
)
```

## Two defects

**1. Board-level holes emit `pcb_component_id: null`.**

`<hole />` and `<platedhole />` placed directly on a board produced an element like:

```json
{ "type": "pcb_hole", "pcb_component_id": null, "hole_shape": "circle", ... }
```

Every id on the `pcb_hole` and `pcb_plated_hole` schemas is `z.string().optional()`, which accepts a string or `undefined` but rejects `null`. A hole inside a footprint has an owning `pcb_component`, so its id resolves to a real string. A board-level hole has no owning component, so the resolved value stayed `null` and reached the output.

**2. DIP plated holes emit `circular_hole_with_rect_pad` without `hole_shape` / `pad_shape`.**

`<chip footprint="dip8" />` generates pin 1 as a `circular_hole_with_rect_pad` plated hole. The `pcb_circular_hole_with_rect_pad` schema requires `hole_shape: "circle"` and `pad_shape: "rect"`, but the insert in `PlatedHole.ts` omitted both:

```json
{ "type": "pcb_plated_hole", "shape": "circular_hole_with_rect_pad", "hole_diameter": 0.8, ... }
```

The neighbouring pill-with-rect-pad branches in the same file already set `hole_shape` and `pad_shape`, so this was a gap in one branch rather than a design choice.

## The change

- `Hole.ts` and `PlatedHole.ts`: the `pcb_component_id` lookup now falls through to `undefined` when neither the parent nor the primitive container owns a `pcb_component`, instead of leaving `null`. `?? undefined` is the same normalization the surrounding `subcircuit_id` and `pcb_group_id` fields already use in these files.
- `PlatedHole.ts`: the `circular_hole_with_rect_pad` insert now sets `hole_shape: "circle"` and `pad_shape: "rect"`, matching the schema and the sibling branches.

Three lines in the emitters. No change to geometry, so the rendered SVG is unchanged.

`tests/components/primitive-components/holes-emit-valid-circuit-json.test.tsx` builds a board with a board-level hole, a board-level `circular_hole_with_rect_pad` plated hole and a `dip8` chip, then runs every emitted `pcb_hole` and `pcb_plated_hole` through `any_circuit_element.safeParse`. On `main` it fails: the board-level hole and the DIP pin-1 pad do not parse. With the change every hole element parses.

The change also flips one existing inline snapshot in `plated-hole-pill-rotated2.test.tsx`: a board-level pill plated hole that was snapshotting the invalid `"pcb_component_id": null` now records `undefined`. That is the bug being corrected, captured in a snapshot, so the update is the fix showing up.

## Scope

- Only touches the two hole emitters. A hole with a real owning component is unchanged, because its id already resolved to a string before this.
- The issue's own sweep noted `pcb_silkscreen_text` and `pcb_fabrication_note_text` carry the same `null` id, but their schemas require a non-optional string, so normalizing to `undefined` would not make them valid. That is a separate design question and is left out of this change.

## Regression risk

Low. The only values that change are a `null` that was already invalid (now `undefined`) and two literal fields that were missing (now present). No passing consumer could have relied on the invalid output. `bun test tests/components/primitive-components tests/drc` is green.

## Verification

Ubuntu, bun 1.3.14.

- `bun test tests/components/primitive-components/holes-emit-valid-circuit-json.test.tsx`: 1 pass / 0 fail.
- `bun test tests/components/primitive-components tests/drc`: 237 pass / 1 skip / 0 fail, 16 snapshots.
- `bun test` (whole suite, run in three shards to fit the local runner): 1339 pass / 37 skip / 4 fail.
  - `tests/components` (the changed area): 621 pass / 4 skip / 0 fail.
  - `tests/repros tests/breakout`: 206 pass / 13 skip / 4 fail.
  - everything else: 512 pass / 20 skip / 0 fail.
  All 4 failures live in the autorouter and none touch holes. `fanout-soic8-sensor-to-i2c-header` and `breakout-sot23-regulator-power-rail` fail the same way on a clean `main` with this change stashed (the external capacity-autorouter returns "Failed to compute first trace"). `repro156-ti-multisheet-autolayout` and `repro116-arduino-uno-reroute` are heavy autolayout tests that time out only under load. Both pass on rerun in isolation.
- `bunx @biomejs/biome format .`: `Checked 1666 files. No fixes applied.`


imrishabh18 (changes requested):
Board level holes can have pcb_component_id as null cause they are not part of component

zkasuran:
Closing this. Agreed, board-level holes can carry a null pcb_component_id, so it is not needed here. Thanks for the correction.

PR #3052, opened 2026-08-07
fix(fanout): explain how to fix a breakout that has no room to escape

seveibar:
## Problem

When a `<breakout>` hugs a fine-pitch package there is genuinely not enough room to escape its pins, and the solver correctly refuses. But `FanoutAutorouter.solveFanout` rethrew the solver's internal phrasing verbatim:

```
FanoutSolver: best layer assignment routed 0/27 connections
```

Nothing there tells a board author what to change — it reads like a solver defect rather than a design constraint. I chased it as a solver bug and got as far as opening a reproduction against tscircuit/fanout-solver before realising the fanout simply needed more space.

## Fix

Translate the failure into board terms at the point where core owns the context:

```
Fanout failed for U1: only 2 of 24 connections could escape to the breakout
boundary. The breakout boundary is 0.05mm from the pads at its narrowest point,
which all 24 escapes have to share. Give the fanout more room by increasing the
breakout's padding, or by extending the breakout to include the parts crowding
it (decoupling capacitors, series resistors) so their pads sit inside the
boundary instead of against it.
```

Three things it adds over the old message:

1. **Names the component.** `componentNamesById` is plumbed from `Group` through the local-autorouter strategy, so obstacles' `pcb_component_id` values are resolved to source names — `U1`, not `pcb_component_0`.
2. **Quotes the number that matters.** The narrowest gap between the fanned-out pads and the boundary is the space every escape shares, so that is what gets measured and reported.
3. **Names both remedies**, because which one is right depends on the layout: widen the breakout, or pull the crowding parts inside it so the boundary sits outside them rather than between them.

New files are small and single-purpose: `get-fanout-space-error-message.ts` builds the message, `get-pcb-component-names-by-id.ts` does the id → name lookup.

## Tests

`tests/breakout/breakout-insufficient-fanout-space-error.test.tsx` — a QFN56 in a `padding="0.05mm"` breakout with 24 pins escaping. Asserts the `pcb_autorouting_error` names `U1`, explains the failure, names both remedies, and no longer leaks `best layer assignment`.

Typecheck clean. `tests/breakout` is 12 pass / 2 skip / 2 fail, and both failures (`fanout-soic8-sensor-to-i2c-header`, `breakout-sot23-regulator-power-rail`) reproduce unchanged on a clean checkout of this branch point — they are not from this change.

## Note

The gap is only reported when a shared boundary is available. Worth a separate look: for a `<breakout>` core currently passes neither `sharedBoundary` nor `componentBounds` to the solver — `solverOptions` is just `{ borderDistribution, compactBusTracks }` — which sits awkwardly against the fanout solver's "all footprints in one problem escape through the same shared boundary" invariant.

### DIFF
--- a/lib/components/primitive-components/Group/Group.ts
+++ b/lib/components/primitive-components/Group/Group.ts
@@ -29,6 +29,7 @@ import type { GraphicsObject } from "graphics-debug"
 import type { PrimitiveComponent } from "lib/components/base-components/PrimitiveComponent"
 import { AutorouterError } from "lib/errors/AutorouterError"
 import type { AutorouterOptions } from "lib/utils/autorouting/CapacityMeshAutorouter"
+import { getPcbComponentNamesById } from "lib/utils/autorouting/get-pcb-component-names-by-id"
 import { FanoutAutorouter } from "lib/utils/autorouting/FanoutAutorouter"
 import type { GenericLocalAutorouter } from "lib/utils/autorouting/GenericLocalAutorouter"
 import type { SimplifiedPcbTrace } from "lib/utils/autorouting/SimpleRouteJson"
@@ -1328,6 +1329,7 @@ export class Group<Props extends z.ZodType<any, any, any> = typeof groupProps>
               busFanoutDirections: routingPhasePlan.busFanoutDirections,
               fanoutBounds: routingPhasePlan.fanoutBounds,
               fanoutRoutingLayers: routingPhasePlan.fanoutRoutingLayers,
+              componentNamesById: getPcbComponentNamesById(db),
             })
           }
 
--- a/lib/utils/autorouting/FanoutAutorouter.ts
+++ b/lib/utils/autorouting/FanoutAutorouter.ts
@@ -25,6 +25,7 @@ import type {
   SimplifiedPcbTrace,
 } from "./SimpleRouteJson"
 import { getFanoutSharedBoundary } from "./get-fanout-shared-boundary"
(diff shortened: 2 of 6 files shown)


seveibar:
Two corrections to this PR, both found while actually applying the advice it gives.

**1. The remedy named the wrong prop.** The message said "increasing the breakout's padding", but `padding` pads the group's layout and does not move the fanout boundary. I verified by dumping the resolved `fanoutBounds`:

| breakout prop | resolved fanoutBounds |
|---|---|
| `padding="1.2mm"` | `minX -6.38, maxX 6.68, minY -6.28, maxY 4.012` |
| `padding="4.5mm"` | `minX -6.38, maxX 6.68, minY -6.28, maxY 4.012` — identical |
| `fanoutBoundaryPadding="1.2mm"` | grows by exactly 1.2mm on every side |

Pushed a follow-up commit so the message names `fanoutBoundaryPadding` and notes that `padding` will not do it. Following the old wording would have left a user changing a number that does nothing — worse than the original bare error.

Worth considering separately whether `<breakout padding>` should feed the fanout boundary, or at least warn when it is set alone on a breakout that then fails to fan out.

**2. My closing note above was wrong.** I said core "passes neither `sharedBoundary` nor `componentBounds`". `solveFanout` does pass `sharedBoundary` whenever `this.options.fanoutBounds` is set — I had dumped `getFanoutSolverOptions()` before the spread that adds it. `componentBounds` does still appear to go unset. Sorry for the noise.

PR #3058, opened 2026-08-07
Add failing netlabel routing repro with schematics disabled

0hmX:
## Summary

- add one focused dynamic-render regression test
- mark it with `test.failing` to document the current broken behavior
- add a PCB snapshot showing two components without the expected routed trace
- render an automatic result annotation showing expected versus actual route count
- snapshot the one-point SRJ net and assert that one PCB trace should be generated

## Problem

When schematic rendering is disabled, connectivity introduced through a `<netlabel>` is not preserved for PCB autorouting. The generated Simple Route JSON contains only the directly connected endpoint, so the autorouter receives a one-point connection and produces zero routed traces.

## Expected behavior

Netlabel-derived electrical connectivity should remain available to PCB and autorouting phases regardless of whether schematic output is enabled. The two connected pads should produce one routed PCB trace.

## Validation

- `BUN_UPDATE_SNAPSHOTS=1 bun test tests/repros/repro-netlabel-schematic-disabled-srj.test.tsx`

The focused test currently passes as an expected failure with one PCB snapshot and three assertions. Once fixed, it can be changed from `test.failing` to `test`.


seveibar (changes requested):
????????? PCB snapshot but for a schematic issue?

seveibar (approved):
ok i think i get it

PR #3066, opened 2026-08-07
fix(fanout): keep the routing area consistent with the fanout boundary

seveibar:
Follow-up to tscircuit/fanout-solver#30, fixing the incoherence at its source rather than tolerating it in the solver.

## Problem

There are two boundaries in play and they are meant to be nested:

| | meaning | source |
|---|---|---|
| `simpleRouteJson.bounds` | where copper may exist | the breakout's `pcb_group` box |
| `fanoutBounds` / `sharedBoundary` | the edge escapes terminate **exactly** on | the same box — until `fanoutBoundaryPadding` moves it |

By default they are derived from the same `pcb_group` and are therefore identical, which is why nothing noticed they are different concepts. `fanoutBoundaryPadding` then moves one and not the other: `resolveFanoutBounds` updates `routingPhasePlan.fanoutBounds` and the `pcb_group`, but `simpleRouteJson.bounds` keeps its earlier value.

Captured from a real board with `fanoutBoundaryPadding="1.2mm"`:

```
srj.bounds   x -6.38..6.68,  y -6.28..4.01
fanoutBounds x -7.58..7.28,  y -7.43..5.21   <- outside on every side
```

Every exit then sits outside the region copper may occupy. The fanout solver rejected all of them, so the one prop for giving a fanout more room made it route nothing.

## Fix

Widen `simpleRouteJson.bounds` to contain the resolved `fanoutBounds`, right where the `pcb_group` is already updated to match. The two stay nested by construction.

Also adds a guard in `FanoutAutorouter.solveFanout` for the case where they are still inconsistent:

```
Fanout boundary extends 1.2mm outside the routable area, so its escapes would
have to terminate where copper cannot go. This usually means the fanout
boundary was widened without widening the routing bounds to match.
```

Previously that situation produced a bare `routed 0/N connections` with nothing pointing at the cause.

## Tests

`tests/breakout/breakout-fanout-boundary-padding-widens-routing-bounds.test.tsx` — a breakout with `fanoutBoundaryPadding` must not produce the out-of-bounds error. Verified non-vacuous: reverting the widening makes it fail with exactly that error.

### Snapshot changes

Two breakout snapshots shift by 0.08%: the routable area is marginally larger, so one 45-degree bend and a pair of plane-fanout vias move a fraction. I inspected both diffs — routes connect the same pads by the same paths otherwise. Only these two are updated.

`tests/breakout` is 14 pass / 2 fail. Both failures (`fanout-soic8-sensor-to-i2c-header`, `breakout-sot23-regulator-power-rail`) reproduce on a clean checkout and are deliberately left alone rather than re-baselined. Typecheck errors about `pcbPackSolverTimeoutMs` and `beta_pipeline9` are likewise pre-existing on `main`, both awaiting a props release.

## Related

- tscircuit/fanout-solver#30 (merged) made the solver tolerate the boundary being outside `srj.bounds`; this stops core producing that input
- tscircuit/core#3053 warns when `padding` is set without `fanoutBoundaryPadding`


seveibar:
CI caught a mistake of mine — correcting the record.

**The two snapshot updates in this PR were spurious and are now reverted.** My local `node_modules` had `@tscircuit/capacity-autorouter` 0.0.750 while `package.json` requires `^0.0.765`, so the snapshots I regenerated came from an older router. Re-checked with the correct version: this change leaves both snapshots byte-identical, and `tests/breakout` is **16 pass / 0 fail**.

So the PR is now just the two lib changes plus the new test, with no snapshot churn at all — which is the right shape for it.

**A second correction, affecting my earlier comments here and on #3053:** I described `fanout-soic8-sensor-to-i2c-header` and `breakout-sot23-regulator-power-rail` as pre-existing failures on `main`. They are not. They also fail only under the stale 0.0.750 autorouter and pass with 0.0.765. Sorry for the noise — I had verified they failed on a clean checkout, but that checkout shared the same stale install, so the check confirmed nothing.

PR #3071, opened 2026-08-08
fix: correct pcb_component size and anchor for rotated components

### LINKED ISSUE #3070: pcb_component width/height are not swapped for components rotated 90/270 (and wrong at arbitrary angles)
While laying out a board I rotated a chip 90 degrees and noticed its emitted `pcb_component` had the wrong `width`/`height`: the values are not swapped for the rotation, so they describe the unrotated footprint rather than what is actually on the board.

## Repro

```tsx
<board width="10mm" height="10mm">
  <chip
    name="U1"
    pcbRotation={90}
    footprint={
      <footprint>
        <smtpad pcbX={0} pcbY={0} layer="top" shape="rect" width="1mm" height="0.4mm" portHints={["pin1"]} />
      </footprint>
    }
  />
</board>
```

The pad is 1.0mm x 0.4mm. Rotated 90 degrees it occupies 0.4mm wide by 1.0mm tall, so `pcb_component` should report `width: 0.4, height: 1.0`. On `main` (0.0.1625) it reports `width: 1.0, height: 0.4`, unswapped. The emitted `pcb_smtpad` itself is correct (its width/height are swapped at 90/270), so the component size disagrees with its own pad.

## Knock-on effects

- `pcbPositionAnchor` (the corner and edge anchors) is computed from the component bounds, so a rotated component with an anchor is placed off by up to `|padWidth - padHeight|` per axis.
- Arbitrary angles are wrong too. A 3.9mm x 0.9mm pad rotated 45 degrees yields a skewed component and board bounding box instead of the correct symmetric one.

I have a fix and will open a PR that references this issue.

zkasuran:
Closes #3070

A component rotated 90 or 270 degrees emits a `pcb_component` whose `width` and `height` are not swapped, so they describe the unrotated footprint. The emitted `pcb_smtpad` is already swapped for the rotation, so the component ends up disagreeing with its own pad. Arbitrary angles are wrong the same way, the bounding box is not rotated at all.

## Repro

```tsx
<board width="10mm" height="10mm">
  <chip
    name="U1"
    pcbRotation={90}
    footprint={
      <footprint>
        <smtpad pcbX={0} pcbY={0} layer="top" shape="rect" width="1mm" height="0.4mm" portHints={["pin1"]} />
      </footprint>
    }
  />
</board>
```

The 1.0 x 0.4 mm pad rotated 90 degrees occupies 0.4 mm wide by 1.0 mm tall. `pcb_component` should report `width: 0.4, height: 1.0`; on `main` it reports `1.0 x 0.4`.

## Cause

`SmtPad.getPcbSize()` returns a rect or pill pad's `width`/`height` with no rotation awareness:

```ts
if (props.shape === "rect") {
  return { width: props.width!, height: props.height! }
}
```

`doInitialPcbPrimitiveRender()` in the same file already swaps them at 90/270 (and emits a rotated shape for arbitrary angles). `getBoundsOfPcbComponents()` builds the component bounds from `getPcbSize()`. Those bounds feed `doInitialPcbComponentSizeCalculation()` (which writes `pcb_component.width/height`) and the `pcbPositionAnchor` alignment, so the size and any corner or edge anchor come out wrong for every rotated component.

## The change

`getPcbSize()` now applies the component's PCB rotation to the rect and pill branches, mirroring the emit path: it swaps width/height at 90/270 and uses the existing `getAxisAlignedSizeFromRotatedRect` helper for arbitrary angles. Axis-aligned (0/180) is unchanged, so non-rotated boards are untouched.

## Verification

- New test `repro-pcb-component-size-90-rotation.test.tsx`: fails on `main` (width 1.0), passes after (width 0.4).
- `repro-simple-route-json-45-degree` now reports a symmetric bounding box (`±3.697`) for a 45 degree rect where `main` gave a skewed `3.95 x 2.45`.
- Snapshots refreshed for the rotated components they contain (jumper-pcb-rotation, pinout-pico, bga36 breakout); the diffs are the corrected rotated geometry.
- `bunx tsc --noEmit` clean, `biome format` clean.

## Disclosure

AI assistance (Claude, Anthropic) was used in developing this change. Verified locally before submitting: the repro and 45 degree tests, the affected rotation snapshots, `bunx tsc --noEmit` and `biome format`.


imrishabh18 (changes requested):
The width, height and rotation stays as source of truth in circuit-json. We don't swap the pcb_component dimensions just cause of rotation

zkasuran:
Makes sense, thanks for the quick look. If width, height and rotation are the source of truth then folding the rotation into width/height is the wrong move. Closing this and the issue too.

PR #3072, opened 2026-08-08
Expose Matchpack in solver debugging tools

mohan-bee:
### Motivation
- Make schematic Matchpack runs available to solver debugging tools such as Runframe's Solvers tab.

### Before
- Core ran `LayoutPipelineSolver` without lifecycle events or a solver registry entry.

### After
- Core registers `LayoutPipelineSolver` and emits its exact constructor input when schematic auto-layout starts.
- Core emits `solver:ended` with solved, failed, iteration, and error state after Matchpack finishes.


seveibar (approved):
very very nice, but you're missing the solver ended event!

PR #3080, opened 2026-08-08
fix: parse SI-prefixed fuse current and voltage ratings

### LINKED ISSUE #3079: fuse currentRating="500mA" emits current_rating_amps: 500 (500 A) instead of 0.5
### Bug

A fuse whose rating uses an SI prefix is parsed with `parseFloat`, which drops the prefix. `parseFloat("500mA")` is `500`, so a 500 mA fuse ends up with `current_rating_amps: 500` (500 A), a 1000x error. Its schematic label also reads `500A`.

Milliamp fuses are very common (500mA, 100mA and so on), so this hits realistic boards.

### Repro

```tsx
const circuit = new RootCircuit()
circuit.add(
  <board width="12mm" height="10mm">
    <fuse name="F1" currentRating="500mA" voltageRating="250V" />
  </board>,
)
circuit.render()
const sc = circuit
  .getCircuitJson()
  .find((e) => e.type === "source_component" && e.ftype === "simple_fuse")
```

### Expected

- `current_rating_amps`: `0.5`
- `display_current_rating`: `"500mA"`
- schematic label: `500mA / 250V`

### Actual

- `current_rating_amps`: `500`
- `display_current_rating`: `"500A"`
- schematic label: `500A / 250V`

### Cause

`lib/components/normal-components/Fuse.ts` parses `currentRating`/`voltageRating` with `parseFloat`. `fuseProps.currentRating` is `z.union([z.number(), z.string()])` with no transform, unlike `resistorProps.resistance` / `capacitorProps.capacitance` / `inductorProps.inductance` which transform the string to a number in `@tscircuit/props`. `parseFloat` cannot handle an SI prefix. The existing fuse tests only use whole-unit values (`"10"`, `"1A"`, `"2A"`, `"24V"`, `"220"`) where `parseFloat` happens to agree, which is why this was never caught.

zkasuran:
### Summary

Fuse parsed `currentRating` and `voltageRating` with `parseFloat`, which drops the SI prefix. A `500mA` fuse emitted `current_rating_amps: 500` (500 A) instead of `0.5` and its schematic symbol read `500A`. This honors SI-prefixed ratings.

Closes #3079.

### Change

`lib/components/normal-components/Fuse.ts` now uses `parseAndConvertSiUnit` from `format-si-unit` (already a dependency, already imported in this file for `formatSiUnit`) instead of `parseFloat`, both for the emitted `source_component` values and for the schematic display label.

- `"500mA"` -> `0.5` A and `"250mV"` -> `0.25` V
- plain values like `"10"`, `"2A"`, `"24V"` or `"220"` are unchanged, so no existing snapshot moves

### Test

`tests/components/normal-components/fuse-si-prefixed-ratings.test.tsx` builds a `500mA / 250V` fuse and asserts `current_rating_amps === 0.5`, `voltage_rating_volts === 250`, the display strings and `symbol_display_value === "500mA / 250V"`, plus a schematic snapshot.

### Verified locally

- `bun test` on the new test, the existing `fuse.test.tsx` and the fuse bounding-box test: all pass
- `bunx tsc --noEmit`: clean
- `biome format`: clean
- existing fuse snapshots unchanged (all use whole-unit values)

### AI assistance

AI (Claude, Anthropic) was used to help find and fix this. The design, review and verification were done by the author. Verified locally before submitting: the tests, typecheck and format checks listed above.


lib/components/normal-components/Fuse.ts
```diff
@@ -26,11 +26,9 @@ export class Fuse extends NormalComponent<typeof fuseProps, PassivePorts> {
     const rawCurrent = this._parsedProps.currentRating
     const rawVoltage = this._parsedProps.voltageRating
 
-    const current =
-      typeof rawCurrent === "string" ? parseFloat(rawCurrent) : rawCurrent
+    const current = parseAndConvertSiUnit(rawCurrent, "A").value ?? undefined
```
imrishabh18: Don't fallback to `undefined`

imrishabh18 (approved):
See comments, and resolve the conflicts

zkasuran:
Thanks for the review. Both done:

- Dropped the `?? undefined` fallback at all four sites and read `parseAndConvertSiUnit(...).value` directly, since `formatSiUnit` already accepts null. The `source_component` insert now omits `voltage_rating_volts` when no `voltageRating` is given rather than writing null.
- Merged latest main and regenerated the rp2040 schematic snapshots so they carry both the new matchpack layout and the 500A to 500mA correction. On the autolayout snapshot the only visual delta versus main is the fuse label.

tsc, biome and the affected tests are green locally.

seveibar (approved):

PR #3097, opened 2026-08-08
test: add resistor tolerance repro

0hmX:
## Summary
- add a core-native repro and PCB snapshot for dropped resistor tolerance

## Validation
- `bun test tests/repros/repro-resistor-tolerance.test.tsx`


seveibar (commented):
i don't think the desired behavior is necessarily correct here- tolerance may not be the source component property key, we will definitely support display_tolerance however so you could add that

seveibar:
e.g. resistance_tolerance could be used, could be a string etc., there are too many possibilities that need to be grounded in a use case

it's also possible we should be using display_resistance to hold the tolerance value. unless you need this fix don't try to fix it

PR #3119, opened 2026-08-09
Fix validation for invalid capacitor capacitance

anil08607:
Fixes https://github.com/0hmX/capacitor-invalid-capacitance
## Summary

- Reject capacitor capacitance values that do not parse to finite numbers
- Prevent invalid Circuit JSON values such as `null` and display values such as `NaNpF`
- Keep validation at the standard component prop boundary
- Replace the previous `NaNpF` expectation with a validation-error regression test
- Preserve automatic decoupling trace-length behavior

## Testing

- Added invalid capacitance regression coverage
- Verified capacitor display-value tests
- Verified automatic decoupling tests
- 10 related tests pass
- Biome and conflict-marker checks pass

### DIFF
--- a/lib/components/normal-components/Capacitor.ts
+++ b/lib/components/normal-components/Capacitor.ts
@@ -1,17 +1,23 @@
 import { capacitorProps } from "@tscircuit/props"
 import type { SourceSimpleCapacitorInput } from "circuit-json"
+import { formatSiUnit } from "format-si-unit"
 import {
-  FTYPE,
   type BaseSymbolName,
+  FTYPE,
   type PolarizedPassivePorts,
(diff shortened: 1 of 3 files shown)


mohan-bee:
Not a great issue to fix ig , i think ai is our user I never seen ai did that to me.

anil08607:
@mohan-bee  Im not sure, ohm has reported that bug, so i try to fix it.

imrishabh18 (commented):
@seveibar Not sure what is the right behaviour is for this case, `NaN` which will keep the schematic clean or throwing Zod error

seveibar (changes requested):
Wrong repo

PR #3133, opened 2026-08-10
fix: emit inductor inductance as a parsed number not a raw string

### LINKED ISSUE #3111: inductor: unit strings violate the Circuit JSON inductance type
A documented string value such as `inductance="10uH"` is emitted as a string even though `SourceSimpleInductor.inductance` requires numeric henries.

Expected: parse the unit string to a numeric henry value, or reject it.
Actual: a string reaches Circuit JSON.

Standalone repro: https://github.com/0hmX/tscircuit-repro-inductor-string-inductance

zkasuran:
## What

`<inductor inductance="10uH" />` stored `source_component.inductance` as the raw string `"10uH"`. `SourceSimpleInductor.inductance` is numeric henries, so a string there is the wrong type in the emitted Circuit JSON. Fixes #3111.

## Why

`doInitialSourceRender` read `this.props.inductance` (the raw input) instead of the zod-parsed `_parsedProps.inductance`. Every other passive emits the parsed value: capacitor writes `props.capacitance`, resistor writes `props.resistance`, crystal writes `props.frequency`. Inductor was the only one reaching for the raw prop, so numeric input worked while the documented string form did not.

`display_inductance` already used `_parsedProps`, so the schematic label looked right while the stored value was wrong. That is why it went unnoticed.

## Change

One line in `lib/components/normal-components/Inductor.ts`: `this.props.inductance` becomes `props.inductance` (the parsed value). Now `"10uH"` is stored as `0.00001` henries, matching the schema and the sibling components.

## Test

Added `tests/components/normal-components/inductor-string-inductance.test.tsx`. It renders an inductor with `inductance="10µH"` and asserts the emitted `inductance` is the number `0.00001`, that its type is `number`, that `display_inductance` stays `"10µH"`, plus a schematic snapshot. The test fails on main (the value comes back as the string `"10µH"`) and passes with this change.

## Verified locally

- `bun test` on the normal-components suite and the inductor, spice and repro tests that use inductors: 362 pass, 3 skip, 0 fail
- `bunx tsc --noEmit`: clean
- `bunx @biomejs/biome format .`: clean
- `bun run build`: success
- `bunx @tscircuit/dependency-check`: passed

No existing snapshot changed, since `display_inductance` was already correct.

---

AI assistance (Claude) was used in developing this change. The design, review and verification were done by the author. Verified locally before submitting: bun test (normal-components plus the inductor, spice and repro sets, 362 pass 3 skip 0 fail), bunx tsc --noEmit, bunx @biomejs/biome format ., bun run build, bunx @tscircuit/dependency-check.


zkasuran:
Closing this. Refocusing on board-driven fixes rather than small changes. Thanks for the steer.

Issue #3100, opened 2026-08-08
board: anchorAlignment is ignored


0hmX:
`anchorAlignment` is documented as the preferred board anchoring prop but the runtime reads only deprecated `boardAnchorAlignment`.

Expected: `anchorAlignment="top_left"` positions the board so its top-left is at `boardAnchorPosition`.
Actual: the value is ignored.

Evidence:
- Core repro PR: https://github.com/tscircuit/core/pull/3094
- Standalone repro: https://github.com/0hmX/repro-board-anchor-alignment-20260809

PR #3135, opened 2026-08-10
feat: anchorAlignment prop for board components

ArnavK-09:
#3100


seveibar (changes requested):
create the repro snapshots separately (in a separate PR than the fix) so that we can see what changed

PR #3145, opened 2026-08-10
repro: board ignores anchorAlignment

ArnavK-09:
https://github.com/tscircuit/core/pull/3135#pullrequestreview-4899256138

image

repro for #3100


seveibar (commented):
The snapshot doesn’t have sufficient information. You also need to say the size of the board in order for the Center to make sense relative to the alignment.

PR #3151, opened 2026-08-11
Add repro for GND symbol intersecting trace on up-facing pinheader

techmannih:


tests/repros/__snapshots__/repro-pinheader-up-facing-net-connections-schematic.snap.svg
techmannih, line 1: Image

MustafaMulla29 (approved):

PR #3152, opened 2026-08-11
Orient enclosure apertures from the component frame

addibble:
## Summary

- prefer footprint-local `cutoutApertureDirection` over `insertionDirection`, with nearest-edge detection remaining the fallback when neither is authored
- preserve the continuous aperture axis through component rotation and bottom-layer transforms instead of quantizing it to a wall direction
- anchor the transformed axis at the component datum so the enclosure solver can choose the first wall it intersects
- derive horizontal aperture profile roll from the PCB component frame rather than CAD asset alignment offsets

## Layering

All prerequisites for this final layer are merged:

- #3166 — `assembly.device` compatibility container
- #3174 — staged two-part solver integration
- #3180 — separate base and lid CAD components
- #3196 — component body, aperture offsets and explicit depth
- tscircuit/3d-viewer#974 — compatibility appearance controls for the assembled enclosure

This PR is now rebased onto Core `v0.0.1670` and contains only the component-relative aperture-axis layer.

## Circuit JSON compatibility

This deliberately preserves the existing synthetic enclosure representation. It does not emit unreleased typed enclosure records or persist viewer appearance state in Circuit JSON.

## Engineering guidance

This PR also adds the repository-local geometry rules learned while implementing the axis transform: explicit frame declarations, canonical direction names, composed transforms, proper bottom-layer rotations, and discriminating geometry tests. Keeping the guidance with the implementation prevents the same frame defects from returning.

## Testing

- `bun run format`
- `bunx tsc --noEmit`
- dependency check
- build and `smoke-test:dist`
- duplicate-helper check
- focused enclosure tests: 4 passed
- full Core suite: 1,400 passed, 37 skipped, 0 failed; 58 snapshots and 7,961 assertions

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -144,6 +144,81 @@ circuit.selectOne("resistor") // Find first resistor
   at an untrusted boundary; redundant internal guards add noise and hide broken
   type contracts.
 
+## Coordinate frames, side names, and transforms
+
+### Declare the frame at every boundary
+
+Any function or record carrying geometry states explicitly, in its docstring:
+**which frame** (footprint-local, board/circuit world, renderer scene), **what
+the axes mean**, **units** (mm throughout tscircuit), **handedness and which way
+is up**, and **whether the value is a point or a direction** — a point picks up
+translation, a direction must not. Most of the transform defects here were two
+functions silently disagreeing about which frame they were handed.
+
+### Validate a convention before copying it
+
+Matching surrounding code is the default *only* once you have confirmed the
+surrounding code is intentional. Prevailing patterns in this area have
+repeatedly turned out to be bugs, or compensations for bugs elsewhere. Before
+adopting one, find its origin commit, the test that pins it, or a measurement
+that confirms it — and prefer removing a compensation at its source over adding
+a matching one.
+
+### Canonical direction names
+
+Directions name **where something is**, in board space: +X `right`, −X `left`,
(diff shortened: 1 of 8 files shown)


lib/components/primitive-components/EnclosureCutoutAperture/get-component-body.ts
```diff
@@ -0,0 +1,136 @@
...
+  cad,
+  authoredModel,
+}: {
+  cad: CadComponent
+  authoredModel?: Record<string, any>
+}): number | undefined => {
+  // `modelBounds` is staged in Props before it becomes durable Circuit JSON.
+  // Prefer the eventual record field when present, but read the parsed prop in
+  // this migration PR so Core can consume it without a schema fork.
+  const bounds =
```
seveibar, line 30: Circuit json spec issue? We can add model_bounds probably if there isnt anything already there for that-

addibble: yes I use it for core to locate the middle of the model relative to the board edge to center the aperture. we do have a size attribute and a model_origin_position but unless we read the model vertices we do not know the extent of the positive and negative vertices from the origin and I thought it was best to leave model loading to the renderers. It might be worth implementing more assembly/clearance DRC to figure out if this is sufficient for that purpose as well before canonicalizing it in circuit-json.

seveibar (approved):

PR #3160, opened 2026-08-11
feat: resolve noSchematicRepresentation support for Normal Component

### LINKED ISSUE #3117: pushbutton: noSchematicRepresentation is ignored
Accepted `noSchematicRepresentation` does not suppress the pushbutton schematic representation.

Expected: no schematic representation when enabled.
Actual: the representation is still emitted.

Standalone repro: https://github.com/0hmX/repro-pushbutton-no-schematic-representation-20260809

ArnavK-09:
fix #3117

because it inherits ChipProps
so prop is doc-ed


seveibar (commented):
this should probably be in NormalComponent, idk why it would be a push button specific thing

seveibar:
waiting for tests to pass before reviewing because idk if this will have negative effects

seveibar (approved):

PR #3166, opened 2026-08-11
Add assembly.device as a compatibility container

addibble:
Adds the authoring/container layer independently from the enclosure solver migration.

- exports `assembly.device`
- uses the assembly-specific `isAssemblyDeviceContainer` guard rather than overloading Core's existing root terminology
- allows board discovery through the wrapper
- avoids electrical group/subcircuit semantics
- preserves current Circuit JSON and emits no assembly record
- enables an existing `enclosure.fdm.box` under the wrapper

This is layer 1 of the staged Core migration. Layer 2 is reviewable at https://github.com/addibble/core/pull/2.


lib/components/base-components/is-root-container.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * A root container is a product-level wrapper (today only `<assembly.device>`)
+ * that may sit above a `<board>` without being an electrical group or
+ * subcircuit. Components that would otherwise assume "my parent is a
+ * subcircuit" use this to opt out.
```
seveibar: ehhh i don't think this is a good idea- RootCircuit / RootContainer doesn't make much sense to me- in general we want to be careful about re-using the name "Root" since there is only supposed to be "one root thing"

I think if you are just more specific and say `isAssemblyDeviceContainer` or something that would be fine?

seveibar (commented):
i think root container could be short sighted? devices go inside a larger assembly

lib/components/base-components/is-root-container.ts
addibble: now it's called AssemblyDeviceContainer, not RootContainer anymore

seveibar (approved):

PR #3176, opened 2026-08-12
Fix pushbutton externally connected pins

### LINKED ISSUE #3118: pushbutton: externallyConnectedPins is ignored
Accepted `externallyConnectedPins` does not affect pushbutton connectivity.

Expected: specified pins are treated as externally connected.
Actual: the prop has no effect.

Standalone repro: https://github.com/0hmX/repro-pushbutton-externally-connected-pins-20260809

rushabhcodes:
## Summary

Fix `externallyConnectedPins` support for `<pushbutton />` components.

`PushButton` now creates the requested source-level traces, matching the behavior already implemented by `Chip`. A regression test verifies that `SW1.pin1` connects to `SW1.pin2` in the generated circuit JSON.

## Validation

- Focused pushbutton tests pass
- `bunx tsc --noEmit` passes
- Formatting and diff checks pass

## Issue

Fixes https://github.com/tscircuit/core/issues/3118


imrishabh18 (commented):
I am confused, can you explain more about the issue?

rushabhcodes:
@imrishabh18 
The issue is that `<pushbutton />` accepts the `externallyConnectedPins` prop, but currently ignores it during rendering.

For example:

```tsx
<pushbutton
  name="SW1"
  externallyConnectedPins={[["pin1", "pin2"]]}
/>
```

This should create a `source_trace` connecting `SW1.pin1` and `SW1.pin2`, similar to how `<chip />` handles the same prop. Before this PR, no `source_trace` was created even though the prop passed TypeScript/Zod validation.

This PR adds that missing trace creation behavior and includes a regression test for it.

seveibar:
@rushabhcodes pushbuttons don't usually have externallyConnectedPins

seveibar:
i'm sort of confused why you want this

seveibar:
i don't think this is based on a real issue, we shouldn't spend our time on stuff that isn't actually used

PR #3179, opened 2026-08-12
Add repro for incorrect component bounds on rotated rectangular pads

techmannih:
Adds a regression repro for `jlcpcb:C1526234` placed at `-90°` next to a `cap0402`.

The generated `pcb_component` bounds incorrectly overlap C2 by approximately `0.64mm`, even though the actual courtyards have a clear gap of approximately `0.996mm`.

## Root cause demonstrated

The pad centers are transformed according to the component rotation, but the width and height of non-square rectangular pads are not rotated when calculating the component bounds.

## Snapshot annotations

- Red dashed: generated `pcb_component` bounds
- Red filled: false `0.64mm` overlap
- Magenta: actual courtyards
- Green: real `0.996mm` courtyard clearance

## Scope

This PR only adds the repro and snapshot. It does not implement the bounds calculation fix.


tests/repros/repro-c1526234-rotated-component-bounds.test.tsx
imrishabh18, line 1: Cleanup this test, and create a minimal repro. Too hard to follow

techmannih: Done!

imrishabh18 (changes requested):
You just moved it to circuit-json instead of tsx, I mean't can you reduce this to something simpler which is not having too many pads!

seveibar (approved):

imrishabh18 (approved):

PR #3180, opened 2026-08-12
Emit separate base and lid CAD components

addibble:
Layer 3 of the staged Core enclosure migration, following merged #3174.

Uses `output.parts` to emit two existing `cad_component` records—base and lid—sharing the current synthetic enclosure source/PCB owner and placement. Existing renderers already enumerate both records, and the compatibility Appearance control classifies/toggles both through their shared owner.

This intentionally adds no Circuit JSON fields and does not infer base/lid role names from IDs. Durable roles and independent controls remain in the later typed enclosure schema.

The existing USB/PoppyGL regression now asserts two CAD records, distinct IDs, shared visibility/hidden-edge flags, and refreshed separate-mesh snapshots.


addibble:
Temporarily closing/reopening to retrigger the test matrix: all shards failed during Sharp libvips download with a transient socket hang-up before tests ran.

addibble:
The CI install failure is upstream in `bun-match-svg → looks-same@9 → sharp@0.32.6`, not this enclosure diff. Fix: tscircuit/bun-match-svg#14. After it merges/releases, this PR will rebase and bump `bun-match-svg`; no Core override will be carried.

seveibar (approved):

PR #3187, opened 2026-08-13
Repro schematic section title overlap

Jk2006k:
## Reproduction

Adds a minimal visual reproduction for overlapping schematic section titles.

Three sections contain one resistor each and use long titles with `sectionTitleFontSize=0.32in`. On current `main`, the generated snapshot shows the titles occupying overlapping section space because section packing only considers component bounds.

This PR intentionally contains only the reproduction and current before-fix snapshot. The proposed fix is in #3186.

## Validation

- `bun test tests/repros/repro-schematic-section-large-title-overlap.test.tsx`
- `bunx tsc --noEmit`


imrishabh18:
Not a valid issue, they are supposed to be added with X and Y coordinates

Jk2006k:
@imrishabh18 
In schematic section X and Y coordinates are not working

PR #3197, opened 2026-08-13
Document geometry frame and transform rules

addibble:
## Summary

Document the geometry conventions validated during the parametric-enclosure work:

- declare coordinate frames, units, handedness, and point-vs-direction semantics at boundaries
- use canonical axis/direction names and retire ambiguous `front`/`back`
- distinguish direction `top`/`bottom` (±Y) from PCB layers (±Z)
- compose transforms from the component's reference matrix
- treat bottom-layer placement as a rotation rather than an inversion
- require discriminating geometry tests across rotations and layers

Documentation only; no runtime behavior changes.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -144,6 +144,81 @@ circuit.selectOne("resistor") // Find first resistor
   at an untrusted boundary; redundant internal guards add noise and hide broken
   type contracts.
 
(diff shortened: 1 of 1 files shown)


addibble:
Folded into #3152 so the transform guidance lands with the component-relative aperture-axis implementation.

PR #3198, opened 2026-08-13
Reproduce default breakout one-point net error

Abse2001:
## Summary

- reproduce default `<breakout>` routing emitting a one-point net connection
- use no explicit autorouter, fanout, or routing-constraint props
- preserve the enclosing subcircuit, outside R1, and parent VCC traces from the real hierarchical case
- capture both the PCB result and the autorouting phase inputs
- keep the expected behavior as a `test.failing` regression

This PR intentionally contains no production-code fix and no assertion for the separate parent-routing endpoint bug.

## Bug

A manual port-based `<breakoutpoint>` with no `autorouter` prop makes Core split breakout routing into a default escape phase and a fanout phase. For a trace from `U1.pin1` to `net.VCC`, Core emits these as separate connections:

```text
source_net_*: [U1.pin1]
breakout:...: [U1.pin1, breakoutPoint]
```

The fanout phase receives the one-point `source_net_0` connection and fails with:

```text
FanoutSolver: connection "source_net_*" has no target beyond its BGA pad
```

The test inspects every generated routing-phase input and requires that no connection have fewer than two points. It also requires that routing complete without a `pcb_autorouting_error` once the malformed connection is fixed.

Using `autorouter="auto"` avoids this default phase split and hides the failure. The outside R1 and its parent VCC connections intentionally remain in this repro, although they cannot route while the child breakout phase fails. Fixing this bug will let the same hierarchical circuit advance to the separate parent-route/breakout-endpoint issue without that prop or an extra component.

## Validation

- `BUN_UPDATE_SNAPSHOTS=1 bun test tests/breakout/default-breakout-one-point-net.test.tsx --timeout 30000`
- `bun test tests/breakout/default-breakout-one-point-net.test.tsx --timeout 30000`
- `bunx tsc --noEmit`


seveibar (approved):

tests/breakout/parent-net-uses-child-breakoutpoint.test.tsx
```diff
@@ -0,0 +1,58 @@
...
+    parentVccConnection!.pointsToConnect.some(
+      (point) => point.pointId === breakoutPoint.pcb_breakout_point_id,
+    ),
+  ).toBe(true)
+  expect(
+    parentVccConnection!.pointsToConnect.some(
+      (point) => point.pointId === internalPort.pcb_port_id,
+    ),
+  ).toBe(false)
+})
```
seveibar: add autorouting phase snapshot

seveibar (approved):

PR #3208, opened 2026-08-14
Fix custom passive symbol port placement

imrishabh18:
## Issue

A single capacitor component with 8 pins were initially represented with only 2 default ports

image

## Summary

- render every footprint-derived port for passive components that provide a custom schematic symbol
- infer port attachment points from unambiguous open-path endpoints on the custom symbol boundary
- keep the inference in a reusable schematic utility built on `@tscircuit/math-utils` bounds, distance, direction, and transform helpers
- centralize the existing default schematic-port distance instead of duplicating `0.4` literals
- recreate the standard 0.4-unit EasyEDA pin stems and keep explicit symbol ports and user-defined port arrangements unchanged
- add an exact C62857 / 6124B104K500NT repro with eight SMT pads and a schematic snapshot

## Root cause

Capacitors and resistors retained their built-in two-port schematic-symbol metadata even when a custom `symbol` prop was supplied. The additional footprint-derived ports therefore had no native symbol port definitions. The initial workaround used generic schematic-box dimensions, which placed the port endpoints too far from the imported EasyEDA artwork.

This change treats the custom symbol as the schematic geometry source. When the symbol has no explicit ports and its boundary path endpoints match the footprint pin count exactly, core maps those endpoints in CCW pin order and renders the external stems from the original symbol boundary.

## Validation

- `bun test --timeout 15000` across 8 focused capacitor and custom-symbol regression tests: 8 passed
- isolated C62857 repro: 6 assertions passed
- `bun run build`: passed, including declaration generation
- new schematic inference utility passes Biome checks
- `git diff --check`: passed

## Schematic snapshot

C62857 capacitor array with eight aligned schematic ports


seveibar (changes requested):
why doesn't the symbol have it's ports defined?

PR #3211, opened 2026-08-14
keep dense inline net labels attached to their traces

mohan-bee:
before : 

image

after: 

image


mohan-bee:
@copilot  revert the latest commit

seveibar (approved):

PR #3224, opened 2026-08-14
Fix trace connectivity through explicit vias

Abse2001:
## What changed

- When a trace terminates at a port of a net-connected explicit via, source-trace rendering adds that net to the existing `source_trace.connected_source_net_ids`.
- Leaves `Group_doInitialSourceAddConnectivityMapKey` unchanged; the normal connectivity-map phase now receives the correct serialized source relationship.
- Keeps a net-only explicit via represented as `source_manually_placed_via` plus `pcb_via`; no synthetic `source_trace` is created for the via.
- Converts the merged repro from `test.failing` to a passing test and updates its PCB snapshot to show the false overlap DRC is gone.

## Root cause

An explicit via with `connectsTo="net.GND"` already assigned GND to the via body. However, a separate trace ending at `.VIA_GND > .top` serialized only its component and via-port endpoints, leaving `connected_source_net_ids` empty. Downstream connectivity and DRC therefore treated the intentional trace-to-via contact as different nets touching.

The fix resolves the net through the trace's via endpoint while that existing source trace is created. Connectivity-map generation remains generic and derives the shared key from the corrected source trace, port, and net relationship.

## Impact

Traces can terminate on any port of an explicit net-connected via without producing a false overlap DRC. Existing port-connected via behavior remains unchanged.

## Validation

- 10 focused via, subcircuit, and connectivity-map tests
- `bunx tsc --noEmit`
- Biome check on the changed implementation and repro


seveibar (changes requested):
a via doesn't establish the connectivity map- there are many lines of code added to the source add connectivity map file that all represent a giant hack

PR #3241, opened 2026-08-15
Add repro for incorrect bottom-layer CAD model rotation

anil08607:
## Summary

Adds a reproduction for incorrect CAD model rotation when the imported
C2891263 (`XC6206P282MR-G`) component is placed on the bottom PCB layer.

The footprint is mirrored correctly, but the CAD model is rotated 180° relative
to its pads. The bottom-view 3D snapshot clearly captures the mismatch.

This PR only adds the reproduction and snapshots; it does not include the fix.

## Testing

- Added PCB snapshot
- Added direct bottom-view 3D snapshot
- `bun test tests/repros/repro-c2891263-bottom-layer-cad-rotation.test.tsx`


anil08607:
Screenshot 2026-08-15 at 17 22 36

seveibar (changes requested):
this doesn't repro a core bug

PR #3265, opened 2026-08-16
fix(subcircuit): decline caching, loudly, for anything the inflators cannot rebuild

addibble:
Subcircuit isolation renders a subtree in its own circuit, throws the components
away, keeps the AnyCircuitElement[] and rebuilds components from those records
with inflate-circuit-json. That is only transparent for what an inflator knows
how to rebuild; everything else takes one of three exits, and two of them are
silent:

- no record at all (<enclosure.cutoutaperture>) -- dropped, no error anywhere
- a record with no inflator (<netlabel>) -- dropped, no error anywhere
- an ftype the inflator's switch has no case for (<pinheader>) -- throws
  mid-render, so an optimization decides whether the board renders at all

The round trip is now checked before it is taken:
isolation-round-trip.ts holds an allowlist keyed by componentName whose value
names the inflator that rebuilds each entry, so the pairing with
inflate-circuit-json.ts can be checked by reading. A subtree is isolated only
when every declaration in it is on that list; anything else renders normally --
correct output, no cache -- and warns once per circuit, naming the offending
components and why each cannot come back.

The walk stops at components the inflators rebuild whole (a chip's footprint,
pads and ports come back with the chip, so they are not separate declarations to
vet) but keeps descending for declarations that emit no Circuit JSON at all,
which have nowhere to come back from at any depth.

### DIFF
--- a/lib/components/primitive-components/EnclosureCutoutAperture/EnclosureCutoutAperture.ts
+++ b/lib/components/primitive-components/EnclosureCutoutAperture/EnclosureCutoutAperture.ts
@@ -14,6 +14,13 @@ import {
 export class EnclosureCutoutAperture extends PrimitiveComponent<
   typeof enclosureCutoutApertureProps
 > {
+  /**
+   * No Circuit JSON record of its own, so it cannot survive being round-tripped
+   * through one. Subcircuit isolation checks this and declines to isolate a
+   * subtree containing it, rather than dropping it silently.
+   */
+  isEphemeralDeclaration = true as const
+
   get config() {
     return {
       componentName: "EnclosureCutoutAperture",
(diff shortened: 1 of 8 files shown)


addibble:
Closing note, for whoever finds this later: the maintainers want to prioritize subcircuit caching properly rather than take a guard around it, so this isn't the right change to land.

The position it leaves behind: **subcircuit caching (`_subcircuitCachingEnabled`) is broken, and turning it on is at your own risk.** Isolation replaces a subtree with Circuit JSON and rebuilds it from the inflators, so anything without an inflator is dropped silently (a `<netlabel>` vanishes), and an `ftype` the inflator's switch has no case for (`<pinheader>`) throws mid-render. It is off by default, and nothing in eval/cli/runframe turns it on.

Measurements, if they're useful when this does get picked up: the cache was worth 1.00x on 4 identical modules and 1.33x on 24.

Issue #3115, opened 2026-08-08
pushbutton: documented default internal pin connections are absent


0hmX:
The documented default internal connections between each pair of pushbutton pins are missing from Circuit JSON.

Expected: internal pin connections are emitted.
Actual: no documented default internal connections.

Standalone repro: https://github.com/0hmX/repro-pushbutton-default-internal-pins-20260809

Core repro PR: not yet created.

PR #3268, opened 2026-08-16
fix(pushbutton): create a port for every leg, not just two

Reiss-Cashmore:
Hello,

Full disclosure, a LLM helped me prepare this PR but this is not a hallucinated issue. I made these designs using TSCircuit, had them printed by PCBWAY and then realised the mistake once they arrived and then had an LLM help me with the PR to fix it. I am a self admitted noob when it comes to PCB design and software. So if this PR is off the mark initially please do provide feedback and I'll be happy to learn and fix.

Here is a photo of the printed board with a bad soldering job that works around the issue :D :

image

## Problem

`PushButton.initPorts()` hardcodes `pinCount: 2`. A 4-leg tactile switch
therefore only ever gets two ports — pins 3 and 4 exist as copper pads in the
footprint but get no `pcb_port`, so they cannot join a net and are invisible to
connectivity analysis.

This is not theoretical: I shipped a board where both push-button ground legs
were left unrouted. The build succeeded, no DRC complained, and the buttons were
dead on arrival, because from the netlist's point of view those pads did not
exist.

The same hardcoded count also leaves the `pin3Port` / `pin4Port` lookups
`undefined` while typed non-null:

```ts
const pin3Port = ports.find((p) => p.props.pinNumber === 3)! as Port
```

which is what throws `cannot set schematicSymbolPortDef of undefined` whenever
`internallyConnectedPins` is passed to a `<pushbutton>`.

## Fix

Derive the pin count from the resolved footprint instead of hardcoding it, so a
4-leg part gets four ports and a 2-pin part gets two. With no footprint there is
nothing to derive from, so the previous 2-pin default is kept — a bare
`<pushbutton />` is still a two-terminal switch schematically.

The schematic symbol genuinely has only two terminals, and the existing loop
already maps pins 2..4 onto the symbol's second terminal; it now skips absent
pins rather than dereferencing `undefined`.

## Verification

Ports on a 4-leg tactile footprint:

| | source_ports | pcb_ports |
|---|---|---|
| before | 4 | 2 |
| after | 4 | 4 |

`pin4` goes from `pcb_port_id: null` to a real port.

Full suite, compared against a pristine checkout of the same base commit:

| | pass | fail |
|---|---|---|
| `main` (baseline) | 1376 | 25 |
| this branch | 1380 | 23 |

No regressions — and it fixes two pre-existing failures. The remaining 23 are
present on `main` too.

Adds `pushbutton-four-leg-ports.test.tsx`: a 4-leg footprint yields four ports,
and a footprint-less pushbutton still yields two.

## Related

Related to #3115 (pushbutton default internal pin connections absent) but does
not close it. That issue asks for pin1↔pin3 / pin2↔pin4 to be emitted as
internal connections; this PR does not add those. It is arguably a prerequisite,
since those pins previously did not exist to be connected.

### DIFF
--- a/lib/components/normal-components/PushButton.ts
+++ b/lib/components/normal-components/PushButton.ts
@@ -28,8 +28,19 @@ export class PushButton extends NormalComponent<
   }
 
   override initPorts() {
+    // A 4-leg tactile switch has four physical legs. Hardcoding pinCount: 2
+    // meant pins 3 and 4 never got ports — so those pads had no pcb_port, could
+    // not join a net, and vanished from connectivity analysis entirely. (It also
+    // left the pin3/pin4 lookups below undefined, which is what crashed when
+    // internallyConnectedPins was set.)
+    //
+    // With a footprint, let NormalComponent derive the ports from it, so a
+    // 4-leg part gets 4 and a 2-pin part gets 2. With no footprint there is
+    // nothing to derive from, so keep the historical 2-pin default — a bare
+    // <pushbutton /> is still a two-terminal switch schematically.
+    const hasFootprint = Boolean(this.props.footprint)
     super.initPorts({
-      pinCount: 2,
+      ...(hasFootprint ? {} : { pinCount: 2 }),
       ignoreSymbolPorts: true,
     })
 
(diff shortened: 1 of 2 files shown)


Reiss-Cashmore:
Would be good to get a bit of discussion going regarding this, not sure if this MR is a "holistic" fix to this issue

Reiss-Cashmore:
I'll probably just let this close if I can't get any visibility on it. Thats a shame

PR #3269, opened 2026-08-17
Run grid via stitching after copper-pour resolution for same net copper pour present on both top/bottom layers

imrishabh18:
## What changed

- integrates `@tscircuit/via-stitch-solver` as a normal core render solver
- moves `PcbViaStitchRender` after `PcbCopperPourRender`
- runs the solver once per subcircuit from its copper-pour components
- stitches existing same-net copper pours and inserts the resulting vias
- removes the previous thick-power-trace and generated-corridor behavior
- keeps stitching vias clear of component bounds, pads, plated holes, and board holes, including same-net bottom-layer pads
- adds five board-wide fixtures plus convex and concave partial-board polygon-pour fixtures, all with exact snapshots and no diff threshold
- spreads connected components across the polygon fixtures so each circuit has components both inside and outside the stitched region

The implementation is provided by [via-stitch-solver #3](https://github.com/tscircuit/via-stitch-solver/pull/3), and core is pinned to its immutable head commit.

## Behavior

The solver follows the standard EDA operation: existing copper pours on two layers are grouped by net, their overlapping copper is sampled on a regular grid, and a via is added only where its full annulus and clearances fit inside both pours without intersecting PCB obstacles. The solver does not create pours or derive geometry from traces.

Fixed polygon outlines are respected directly: the via grid is clipped to the actual shared copper, so unpoured board regions and concave polygon cut-ins receive no stitching vias. The polygon examples are full routed circuits on larger boards; only one portion of each circuit has the top/bottom GND polygon and stitching grid.

## Validation

- 7 exact via-stitching PCB snapshot tests without diff thresholds
- assertions that polygon fixtures contain components both inside and outside the poured region
- polygon-boundary and component/pad clearance assertions
- TypeScript typecheck
- package build
- dependency check
- formatting


package.json
```diff
@@ -58,6 +58,7 @@
     "@tscircuit/schematic-trace-solver": "^0.0.134",
     "@tscircuit/solver-utils": "^0.0.16",
     "@tscircuit/soup-util": "^0.0.41",
+    "@tscircuit/via-stitch-solver": "github:tscircuit/via-stitch-solver#271c95d937eb5a8ff60c30269047d9d6fe53b938",
```
seveibar: i think we might not be allowed to have github deps in this module (can't remember) i think it might break the library

seveibar (commented):
really i'm confused by the snapshots, what's the point of via stitching without a ground pour?

package.json
imrishabh18: other deps are using

https://github.com/tscircuit/core/pull/3269/changes/BASE..92857d23ca76dd0f0a6651e8879726dbf37ea070#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R37

PR #3274, opened 2026-08-17
Apply bus layer props as fanout constraints

seveibar:
## Summary

- pass bus `preferredLayer` and `preferredLayers` props into Simple Route JSON
- translate those public props into per-bus `allowedLayers` only at the fanout-solver boundary
- intersect the requested fanout layers with any existing `pcbAllowedLayers` constraint
- preserve the original preference metadata in Simple Route JSON for downstream consumers
- update `@tscircuit/props` to the release containing the new bus props
- use the published `@tscircuit/fanout-solver@0.0.30` release
- verify a four-layer BGA fanout never escapes outside `inner2` and `bottom`

## Dependency

The solver prerequisite landed in tscircuit/fanout-solver#71 and is published in `@tscircuit/fanout-solver@0.0.30`.

The props prerequisite has already landed in tscircuit/props#804 and is published in `@tscircuit/props@0.0.624`.

## Validation

- released solver integration: 2 pass, 0 fail
- focused fanout, breakout, and autorouting suite: 57 pass, 2 skip, 0 fail
- prior full core suite on this branch: 1415 pass, 37 skip, 0 fail
- `bunx tsc --noEmit`
- `bunx biome format .`
- `bun run build`
- `git diff --check`


tests/features/autorouter-fanout.test.tsx
```diff
@@ -88,6 +88,8 @@ test('autorouter="fanout" escapes an inner BGA bus before board routing', async
       <bus
         name="DATA"
         connections={innerPinNumbers.map((_, busIndex) => `D${busIndex}`)}
+        preferredLayer="inner2"
```
0hmX, line 91: so an an whole route the bus in inner2 and for the first child of the bus use bottom

PR #18, opened 2024-09-01
Fix Rotations for SMTPads and SilkscreenPaths, Fix Traces routing through other traces

seveibar:
- **fix trace overlapping**
- **fix smtpads not rotating**
- **silkscreen path transforming properly**
- **adjust rotation for jumper**

Issue #17, opened 2024-09-01
Livestream Issues Sep 1st


seveibar:
- [x] Resistors and capacitors PCB footprints aren't rotating
- [x] Traces are autorouting through other traces

PR #3280, opened 2026-08-18
Place winding-aware breakout points before routing

ShiboSoftwareDev:
## Summary

- keep breakout-point topology selection in core's dedicated breakout-point phase, as requested in review
- pin the exact winding-aware breakout-point-solver commit and the upstream fanout-solver component-identity fix
- render only the solved boundary handoffs, then let fanout route physically to those fixed endpoints
- preserve global routing ownership beyond the breakout boundary
- verify the real QFP16 + header + passives board uses four required handoffs, zero vias, and zero DRC errors

## Stack

Depends on core #3350, breakout-point-solver #18 (repro #17), and fanout-solver #74 (repro #80).

The core head and base are both branches in `tscircuit/core`, so GitHub renders this stack. The breakout-point-solver repository still denies upstream branch pushes to this account; #17/#18 are the existing verified fork PRs for that dependency.

## Review response

The rejected implementation that moved breakout ownership into fanout was fully reverted. This version keeps a dedicated core breakout-point phase: breakout-point-solver selects winding-aware topology, fanout-solver routes to the fixed endpoints, and the global autorouter continues after them.

## Validation

- four affected real-board breakout/fanout tests
- PCB and autorouting-phase snapshots
- TypeScript check
- formatter check
- diff check


seveibar (changes requested):
disagree that breakout should be in the fanout solver- why? it doesn't make sense there AFAIK

ShiboSoftwareDev:
Agreed—the rejected ownership was wrong. I reverted that implementation and reworked this PR: core keeps a dedicated breakout-point phase, breakout-point-solver assigns winding-aware boundary endpoints, fanout-solver physically routes to those fixed endpoints, and the global autorouter continues beyond them. The real QFP fixture now has 4 true crossings, 0 vias, and 0 DRC errors; the PR body and generated SVGs show the result.

PR #3307, opened 2026-08-20
Fix internally connected custom-symbol ports

MustafaMulla29:
## What changed

- Resolve internally connected pin groups from all ports owned by a component, including ports nested inside a custom React symbol.
- Preserve alias-based pin resolution and the existing missing-port error.
- Convert both exact DMT6007LFG-7 repros from expected failures into passing regression tests.

## Root cause

NormalComponent._getInternallyConnectedPins used portMap, which only searches direct child ports. Imported custom symbols place their ports beneath a Symbol primitive container, so valid pins such as pin1 could not be found.

The fix uses the existing _getAllPortsFromChildren path that core already relies on while initializing custom-symbol ports. It does not special-case the MOSFET, supplier part, or pin numbers.

## Validation

- Both merged custom-symbol repros pass as regular tests.
- Nine internal-connectivity tests pass across chips, interconnects, jumpers, and solder jumpers.
- bunx tsc --noEmit passes.
- Existing schematic snapshots remain unchanged.

### DIFF
--- a/lib/components/base-components/NormalComponent/NormalComponent.ts
+++ b/lib/components/base-components/NormalComponent/NormalComponent.ts
@@ -671,9 +671,14 @@ export class NormalComponent<
       {},
       {
         get: (target, prop): Port => {
-          const port = this._getAllPortsFromChildren().find((port) =>
-            port.isMatchingNameOrAlias(prop as string),
+          const symbol = this.children.find(
+            (child) => child.componentName === "Symbol",
           )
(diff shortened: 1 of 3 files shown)


seveibar (approved):
is a group a normal component? If so i think this is very dangerous to merge- you should be doing [...this.children, ...this.symbol.children]

MustafaMulla29:
> is a group a normal component? If so i think this is very dangerous to merge- you should be doing [...this.children, ...this.symbol.children]

Updated

PR #3308, opened 2026-08-20
repro: trace matching fails on rotated rectangular SMT pads

itisrohit:
When a rectangular SMT pad has a ccw_rotation (such as 90° on vertical chip footprints), isPointInSmtPad checks containment using the pad's unrotated width and height. This can cause false overlap matches with adjacent pads and incorrect source_trace_id assignments during post-routing trace matching.


itisrohit:
Closing: after full verification, this bug produces no user-visible symptoms. The incorrect source_trace_id assignment is only observable through internal circuit.json metadata. This does not meet our user-facing repro criteria.

PR #3311, opened 2026-08-20
Pass trace maximum via counts to autorouter input

ShiboSoftwareDev:
## Summary

- add `maxViaCount` to Simple Route JSON connections
- propagate source trace `max_via_count` into autorouter input
- extend the source-trace regression test through the SRJ boundary

## Tests

- `bun test tests/components/primitive-components/trace-max-via-count.test.tsx`
- `bunx tsc --noEmit`
- `bun run build`

### DIFF
--- a/lib/utils/autorouting/SimpleRouteJson.ts
+++ b/lib/utils/autorouting/SimpleRouteJson.ts
@@ -86,6 +86,8 @@ export type SimpleRouteConnection = {
   isOffBoard?: boolean
   netConnectionName?: string
   nominalTraceWidth?: number
+  /** Maximum number of vias permitted across the routed connection. */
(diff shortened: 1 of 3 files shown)


lib/utils/autorouting/SimpleRouteJson.ts
```diff
@@ -86,6 +86,8 @@ export type SimpleRouteConnection = {
   isOffBoard?: boolean
   netConnectionName?: string
   nominalTraceWidth?: number
+  /** Maximum number of vias permitted across the routed connection. */
+  maxViaCount?: number
```
seveibar, line 90: interesting- not sure if we should support this but i'm supportive of introducing the API

PR #3315, opened 2026-08-20
keep schematic sheet frames at the origin

mohan-bee:
### Motivation
- Preserve Circuit JSON positions and keep schematic sheets spec-compliant.

### Before
- Core wrote a non-spec sheet center derived from the schematic contents.

### After
- Sheet frames use the origin without rewriting schematic content, covered by the full battery power snapshot.


lib/utils/schematic/moveSchematicSheetContentsInsideFrame.ts
```diff
@@ -0,0 +1,103 @@
...
+    bounds.maxY - bounds.minY <= sheetMaxY - sheetMinY
+  // Leave oversized content unchanged so warnings identify its actual bounds.
+  if (!contentFitsInsideSheet) return
+
+  let translateX = 0
+  let translateY = 0
+  if (bounds.minX < sheetMinX) translateX = sheetMinX - bounds.minX
+  else if (bounds.maxX > sheetMaxX) translateX = sheetMaxX - bounds.maxX
+  if (bounds.minY < sheetMinY) translateY = sheetMinY - bounds.minY
+  else if (bounds.maxY > sheetMaxY) translateY = sheetMaxY - bounds.maxY
```
imrishabh18: This is overriding the actual position of the components, you can't do this

imrishabh18 (changes requested):

PR #3318, opened 2026-08-20
Pin fanout solver fix for existing trace obstacles

ShiboSoftwareDev:
## Stack

Review and merge the core reproduction first: https://github.com/tscircuit/core/pull/3321

This PR is based on the reproduction branch, so its diff contains only the dependency pin and the same board test changing from the observed broken behavior to the corrected behavior.

Solver stack:

- solver reproduction: https://github.com/tscircuit/fanout-solver/pull/78
- solver fix: https://github.com/tscircuit/fanout-solver/pull/79

## Summary

- pin `@tscircuit/fanout-solver` to the immutable fix commit `dad9d8bde92b8bee5bc731665c9d66422385c3c7`
- change the reproduction's assertion from no bottom-layer escape to a required bottom-layer escape
- replace the expected top-layer overlap error with clean routing and clearance error tables
- update the same PCB snapshot from crossing copper to a bottom-layer underpass

## Before and after

The reproduction routes a vertical `PRE_ROUTED` trace on `top` in phase 0. Phase 1 receives that emitted trace in its `SimpleRouteJson` and fans out a horizontal `SIGNAL` across the same X coordinate, with both `top` and `bottom` available.

In #3321, released `@tscircuit/fanout-solver@0.0.30` keeps `SIGNAL` on `top`, crosses the earlier different-net trace, and produces one `pcb_trace_error`.

With this pin, the same signal drops to `bottom`, passes underneath the phase 0 copper, and returns to `top` without DRC errors.

phase 1 fanout avoids phase 0 copper

## Verification

- focused real-board/fanout/breakout set: 3 passed, 0 failed
- `bunx tsc --noEmit`
- `bun run build`
- full core suite with the pinned solver: 1,425 passed, 37 skipped, 0 failed; 58 snapshots and 8,108 assertions across 1,205 files


ShiboSoftwareDev:
Replaced by #3323 so both stack branches live in tscircuit/core and GitHub can render the native stack preview.

PR #3320, opened 2026-08-20
restore ports on async four-pad pushbuttons

mohan-bee:
### Motivation
- asynchronous footprint ports must be visible before pcb pad matching.

### Before
- stale selector caches hid inferred internal ports, leaving the real KiCad switch pads without pcb port ids or routed connections.

### After
- clear selector caches after asynchronous port initialization so all four imported pads receive pcb ports and the pcb snapshot shows the routed switch connections.


lib/components/base-components/NormalComponent/NormalComponent.ts
```diff
@@ -1163,6 +1163,8 @@ export class NormalComponent<
 
   updateInitializePortsFromChildren(): void {
     this.initPorts()
+    this._cachedSelectAllQueries.clear()
+    this._cachedSelectOneQueries.clear()
```
seveibar, line 1167: this is unfortunately very illegal for any render phase to do, the fix is elsewhere

seveibar: you know honestly maybe this isn't a huge deal- i'll think about this a bit more- i just don't think it should be inside NormalComponent

seveibar: i would look for another solution but keep this open

PR #3324, opened 2026-08-20
fix: invalidate selector caches on tree mutations

seveibar:
### Motivation
Async footprint loading can add inferred ports after a component has cached selectOne/selectAll results. The four-pad pushbutton then reads its stale two-port result and leaves the loaded pads unattached.

### Fix
- invalidate selector caches at the PrimitiveComponent add/remove mutation boundary
- propagate invalidation only through affected ancestor selector scopes
- retain cached result identity while the component tree is unchanged
- update the repro snapshot so all four switch pads are attached and routed

This is a stacked fix for #3319 and an alternative to #3320; it avoids unconditional cache clearing in NormalComponent render phases.

### Validation
- bun test: 1426 passed, 37 skipped, 0 failed
- bunx tsc --noEmit
- Biome check on changed TypeScript files
- git diff --check


seveibar (commented):
CC @mohan-bee

PR #3338, opened 2026-08-20
Reproduce duplicate fanout output trace identity

ShiboSoftwareDev:
## Reproduction

This runs the core `FanoutAutorouter` twice with the exact output trace from the first solve included in the second solve's `SimpleRouteJson.traces` input.

The test records the current bug: the second solve returns a new fanout trace with the same deterministic `pcb_trace_id` as the trace already present in its input. When phases accumulate their traces, that creates duplicate identities and ambiguous copper ownership.

The existing trace remains a routed trace in `SimpleRouteJson`; it is not converted into an obstacle.

## Expected behavior

The host adapter should suppress only returned fanout traces whose `pcb_trace_id` is already present in its input while leaving the existing trace in `SimpleRouteJson.traces` for the solver.

## Test

- `bun test tests/utils/autorouting/fanout-autorouter-reemits-input-trace-id.test.ts`
- `bunx biome check tests/utils/autorouting/fanout-autorouter-reemits-input-trace-id.test.ts`

This is the reproduction PR. The fix is intentionally kept in the next PR in the stack.

### DIFF
--- a/tests/utils/autorouting/fanout-autorouter-reemits-input-trace-id.test.ts
+++ b/tests/utils/autorouting/fanout-autorouter-reemits-input-trace-id.test.ts
@@ -1,0 +1,65 @@
+import { expect, test } from "bun:test"
+import { FanoutAutorouter } from "lib/utils/autorouting/FanoutAutorouter"
+import type { SimpleRouteJson } from "lib/utils/autorouting/SimpleRouteJson"
+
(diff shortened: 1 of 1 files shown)


ShiboSoftwareDev:
Superseded by the exact reproduction in tscircuit/fanout-solver#89 and stacked fix #90. Moving the stack to the producing repository keeps direct solver consumers correct as well as Core.

PR #3339, opened 2026-08-20
Deduplicate fanout output against input traces

ShiboSoftwareDev:
## Summary

- collect `pcb_trace_id`s already present in the fanout adapter's input traces
- filter only duplicate IDs from the newly returned fanout-trace list
- keep the existing routed copper in downstream `SimpleRouteJson.traces`; no trace is converted into an obstacle
- turn the reproduction into a regression that requires no duplicate emitted trace while confirming the input trace remains in downstream SRJ

## Stack

- Reproduction: #3338
- Fix: this PR

This branch is based directly on the reproduction branch so GitHub can show the stack.

## Tests

- `bun test tests/utils/autorouting/fanout-autorouter-reemits-input-trace-id.test.ts tests/features/breakout-fanout-props.test.tsx`
- `bunx biome check lib/utils/autorouting/FanoutAutorouter.ts tests/utils/autorouting/fanout-autorouter-reemits-input-trace-id.test.ts`

### DIFF
--- a/lib/utils/autorouting/FanoutAutorouter.ts
+++ b/lib/utils/autorouting/FanoutAutorouter.ts
@@ -480,7 +480,12 @@ export class FanoutAutorouter implements GenericLocalAutorouter {
     const output = fanoutSolver.getOutput()
     const fanoutSimpleRouteJson =
       output.simpleRouteJson as unknown as SimpleRouteJson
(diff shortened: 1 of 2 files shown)


ShiboSoftwareDev:
Superseded by the solver-owned stack tscircuit/fanout-solver#89 and #90. The duplicate ID already exists in FanoutSolver.getOutput(), so fixing it in Core would only mask an invalid upstream output contract.

PR #3360, opened 2026-08-21
repro: false clearance errors for unconnected thermal vias

GokulPandi-M:
## What happens

Thermal vias fully contained by an exposed SMD pad inherit its connectivity only after that pad is connected to a trace or net. While the component pin is still unconnected, core leaves the vias electrically separate and the via-to-pad DRC reports one false clearance error per via.

Thermal vias inside an exposed PowerPAD are intentional; TI documents using them to connect the pad to internal or bottom copper planes in the [TPS768-Q1 datasheet](https://www.ti.com/lit/gpn/TPS768-Q1).

## Reproduction

The new test creates a regulator-style board placement with:

- input and output capacitors
- correctly spaced VIN and VOUT pads
- one unconnected exposed thermal pad
- four vias fully contained by that thermal pad
- no EasyEDA or converter dependency

It verifies that the pad already has a connectivity identity, the four vias do not inherit it, and the checker produces four false via-to-pad clearance errors.


GokulPandi-M:
Moved this reproduction to the checks layer as suggested: tscircuit/checks#231. The smaller repro shows that `checkViasInPads` respects `is_via_in_pad_allowed`, while `checkViaPadClearance` ignores it and reports four false errors. Closing this draft in favor of the checks PR.

PR #3362, opened 2026-08-21
feat: add automatic implicit copper pours

imrishabh18:
## Summary

- add a `PcbImplicitCopperPourRender` phase after routing and before explicit copper-pour rendering
- enable the phase per board through the public `automaticPoursEnabled` prop from `@tscircuit/props@^0.0.637`
- remove the temporary root-level `_featureImplicitCopperPours` rollout switch
- convert coarse implicit power-net regions into `CopperPour` components so the exact solver handles final BRep geometry and clearances
- preserve zero-clearance seams between touching same-net regions while maintaining physical board-edge, different-net, trace, pad, cutout, and keepout clearances
- keep PCB DRC ordered after both copper-pour phases
- update the nRF52810 regression to opt in on `<board automaticPoursEnabled />` and verify final pours exclude the RF keepout

## Dependencies

- pin [implicit-copper-pour-solver#8](https://github.com/tscircuit/implicit-copper-pour-solver/pull/8) at commit `47f9301`
- use [copper-pour-solver#85](https://github.com/tscircuit/copper-pour-solver/pull/85) for physical board-boundary handling

## Testing

- `bun test` — 1,459 passed, 40 skipped, 0 failed
- `bunx tsc --noEmit`
- `bun run build`
- `bun run smoke-test:dist`
- `bun run format`
- focused nRF52810 implicit-pour visual and geometric regression
- explicit copper-pour regression suite


seveibar (approved):

seveibar:
CC @imrishabh18 merging

PR #3375, opened 2026-08-22
fix: preserve trace widths when reversing routes

seveibar:
## Summary

- add one shared, immutable PCB-route reversal helper
- preserve physical copper in both default/constant and interpolated thickness modes when route direction changes
- use it for Group output orientation, trace-by-trace orientation, and both `mergeRoutes` reversal paths
- keep `through_pad` endpoints/layers correct and preserve the unused terminal width of each wire run so reversing twice is lossless

## Root cause and reproduction

In default/constant thickness mode, Circuit JSON stores a wire segment's width on the segment's starting route point. Reversing only the route-point array therefore moves variable widths to the opposite physical segments. In interpolated mode, widths are point-local and must instead remain attached to their points as the route reverses. Core did not distinguish those contracts across four reversal paths:

1. `Group.ensureRouteStartsAtSourceTraceStart`
2. `Trace_doInitialPcbTraceRender.ensureRouteStartsAtObjective`
3. the first-route reversal in `mergeRoutes`
4. the subsequent-route reversal in `mergeRoutes`

The first two had duplicate helpers that handled `through_pad` direction but not thickness-mode semantics; the two `mergeRoutes` paths called `.reverse()` directly and also mutated their input arrays. Group now passes the output trace's `route_thickness_mode` to the shared helper. Trace-by-trace routing and `mergeRoutes` retain the default/constant behavior.

This was reproduced with quick-config's deterministic FR2355 autorouter input (`phase-0.input.simple-route.json`) on current autorouter main. Its 69 output traces contain 28 consecutive wire segments shorter than 0.001 mm, including 2 nm `0.30 -> 0.50 -> 0.30` width transitions. The transition pair comes from `power-trace-expander`; its paired inside/outside points currently keep the pad-neck width safe in either route direction despite Core's raw reversal. It exposed the segment-start-width reversal bug directly.

## Tests

- `reverse-pcb-trace-route.test.ts`: three default/constant variable-width wire runs separated by a via and a through-pad plus an interpolated route; verifies every reversed physical segment/point width, through-pad endpoint/layer reversal, double-reversal identity in both modes, and input immutability
- `merge-routes-variable-width.test.ts`: drives both first-route and subsequent-route reversal branches and verifies widths plus input immutability

Validation completed:

- `bun test tests/utils/reverse-pcb-trace-route.test.ts tests/utils/merge-routes-variable-width.test.ts`
- `bunx biome check` on all six changed files
- `bunx tsc --noEmit`
- `bun run build`
- `git diff --check`

## Deferred expander cleanup

This PR intentionally does not remove `power-trace-expander`'s `epsilon = Math.min(1e-6, length / 4)` pad-boundary pair. Removing it independently breaks the existing connected-pad safety/coverage regressions because released Core versions can still reverse routes without remapping widths.

After this Core behavior is released and consumed, tscircuit/power-trace-expander#18 can replace the epsilon pair with a normal boundary transition. That follow-up should assert that it introduces no positive wire segment below 1 µm, preserves pad-neck geometry in both route directions, and covers the exact FR2355 input. The unrelated pre-power sub-micron terminal/topology slivers should not be removed by a broad dedupe without separate DRC-backed analysis.


seveibar:
CI's repro79 snapshot delta is the expected effect of the fix: the panel route is reversed, and its variable-width segments now retain the correct segment widths instead of being shifted by one segment. I reproduced the failure locally; the snapshot token diff is the intended width sequence shift. Commit b523e776 updates that baseline, and the focused repro passes.

seveibar:
The next CI run surfaced the same intended width-sequence correction in the three repro116 region-reroute snapshots, which exercise both mergeRoutes reversal branches. I reproduced all three locally (0.01% visual delta), reviewed the SVG token changes (same path geometry/order reversal plus corrected segment-width assignment), updated their baselines in bf56ee7c, and reran the three focused tests: 3 pass, 273 assertions.

PR #3377, opened 2026-08-22
fix: preserve implicit breakout layers in autorouting

0hmX:
## Dependency

Depends on tscircuit/circuit-json#722 and its npm release: https://github.com/tscircuit/circuit-json/pull/722

Until that release is available and Core updates its registry dependency, typecheck/CI may fail because the published `PcbBreakoutPoint` type does not yet include `layer`.

## Summary

- carry the layer selected by the implicit breakout-point solver into native `pcb_breakout_point.layer`
- use that Circuit JSON layer in both child breakout routing and parent cross-boundary SRJ endpoints
- preserve the existing pad-layer or `top` fallback for legacy/manual breakout points with no layer

## Problem

Core passes bus `preferredLayers` into the implicit solver, which returns each placement with `solvedPoint.layer`. Previously the layer was lost before Circuit JSON emission, so later SimpleRouteJson conversion could emit the breakout target on `top` even when the solver selected an inner layer.

The selected routing layer now lives with the breakout point’s x/y routing constraint in Circuit JSON. This avoids mutable sidecar state and avoids scanning the live component tree during SRJ conversion.

## Checks

Run locally against the linked circuit-json producer branch:

- `bun test tests/breakout` (19 pass, 2 skipped)
- `bunx tsc --noEmit`
- targeted `bunx biome check` on the four changed production files
- `bun run build`
- focused AM62L byte-1 route: 11 SoC fanouts, 11 RAM fanouts, and 11 global routes with zero autorouter errors
- layer consistency: 11/11 matched; shorts check: none


0hmX:
The native Circuit JSON dependency for this fix is now proposed in tscircuit/circuit-json#722: https://github.com/tscircuit/circuit-json/pull/722

It adds backward-compatible `pcb_breakout_point.layer?: LayerRef`, so Core can emit the solver-selected layer and the SRJ converter can consume it directly without scanning the live component tree. The simplified Core refactor is validated locally and is waiting for that contract to merge and release before it can be pushed with a normal registry dependency.

seveibar (approved):

PR #3378, opened 2026-08-22
feat: make pipeline9 the default autorouter

seveibar:
## Summary

- make Pipeline9 the default local autorouter and map `latest` to it
- preserve `beta_pipeline7` as the explicit legacy Pipeline7 escape hatch
- de-duplicate unchanged preloaded traces and shared boundary vias returned across phased/fanout routing stages
- rely on the upstream Pipeline9 via-preservation fix instead of repairing autorouter output in Core
- refresh affected PCB and autorouting snapshots

## Dependency

Includes the merged fix from tscircuit/tscircuit-autorouter#2204 via published `@tscircuit/capacity-autorouter@0.0.836`.

## Validation

- published `@tscircuit/capacity-autorouter@0.0.836`: both targeted Core breakout tests pass with unchanged snapshots, including the no-layer-change-without-via assertion
- upstream: `bun run build`, `bunx tsc --noEmit`, and all 48 non-visual Pipeline9 feature tests
- earlier full Core suite for the Pipeline9-default change: 1,427 pass, 40 skip
- Core formatting check on touched TypeScript files
- `git diff --check`

## Props compatibility

`@tscircuit/props` v0.0.628 already includes `beta_pipeline7` in `AutorouterVersion`, its Zod schema, and its autorouter tests, so no props PR is needed.


seveibar:
Update after validating the merged autorouter release:

- @tscircuit/capacity-autorouter 0.0.836 fixes the original unrelated-winding breakout regression.
- The full Core suite exposed a second missing-via route in the A03 dependency (repro116 Arduino center reroute).
- Root fix: tscircuit/high-density-a01#89.
- Autorouter dependency bump: tscircuit/tscircuit-autorouter#2206.
- With both fixes linked and the autorouter rebuilt, the center reroute passes: 1 test, 124 assertions, with no Core workaround.

This Core PR should wait for those stacked fixes to merge and release before its final dependency/snapshot refresh.

PR #3381, opened 2026-08-23
Route automatic breakout exits toward external endpoints

seveibar:
## Summary

- choose a nearer breakout edge for safe standalone outlier connections while preserving the shared edge for buses, differential pairs, and dense fanouts
- center each split edge group toward its external endpoints
- lock the QFP16 Pipeline9 regression fixture to zero vias and update its PCB/autorouting snapshots

## Why

The routing-quality regression surfaced in #3378 because every connection in the single breakout region was forced through the aggregate right edge. The RESET/R1 connection has its external endpoint on the left, so that start/end choice created a long cross-region escape and unnecessary layer changes. The right-side header exits were also centered on the internal pads instead of the external header endpoints.

This keeps the original winding solve as the baseline feasibility check and only splits ungrouped singleton outliers when at least two routes remain on the original edge. Custom solvers and multi-region breakouts are unchanged.

The affected Pipeline9 fixture goes from 3 vias to 0.

## Testing

- bun run build
- bun test tests/breakout (19 pass, 2 skip, 0 fail)
- bun test (1,431 pass, 40 skip, 0 fail)
- biome check on changed TypeScript files
- git diff --check

Related to #3378.


seveibar:
Superseded by the properly split implementation: winding solver behavior in tscircuit/winding-breakout-point-solver#6, the public endpoint contract in tscircuit/props#809, and Core integration/snapshots in #3382. Closing this Core-owned solver workaround in favor of the winding-owned fix.

PR #3405, opened 2026-08-24
Fix fallback label bounds for inline schematic traces

MustafaMulla29:
## Summary

- provide the solver with the regular fallback-label width when a trace is eligible for an inline label
- derive both inline and fallback dimensions from the existing schematic net-label text measurement helper
- add a regression assertion and update the TB67S579FTG schematic snapshot

## Root cause

When inline placement was blocked, the solver fell back to an anchored label. Core supplied the inline label dimensions but omitted netLabelWidth for that fallback, so the solver used its generic 0.45 mm width while AGC_OUT rendered at 0.96 mm. The unrelated GND trace therefore appeared clear to the solver but crossed the rendered label.

## Testing

- bun test tests/repros/repro-tb67s579ftg-inline-net-labels.test.tsx
- 15 related inline-label, fallback-label, cross-boundary, and cross-sheet schematic tests
- bunx tsc --noEmit
- bunx biome check on the changed TypeScript files
- bun run build


tests/repros/__snapshots__/repro-tb67s579ftg-inline-net-labels-schematic.snap.svg
imrishabh18, line 1: Why are the traces on the right getting split?

MustafaMulla29: Will fix

MustafaMulla29: You are right; this is a regression. netLabelWidth also participates in trace-line routing, so using it for fallback geometry enlarged the routing-time text padding. OUT_A_P, OUT_A_N, OUT_B_N, OUT_B_P, and SERIAL_IN then failed initial routing and became paired terminal stubs.

I separated those concerns in tscircuit/schematic-trace-solver#860: fallbackNetLabelWidth is used only by anchored fallback placement/collision stages and is explicitly excluded from routing decisions. The end-to-end TB67S579FTG repro now keeps all five right-side traces continuous while the GND trace still clears the full AGC_OUT label. I have marked this core PR draft until that solver change is released and the dependency is updated here.

PR #3409, opened 2026-08-24
Add repro for inner pcbPack exceeding board bounds

GokulPandi-M:
Adds a minimal three-component reproduction where `pcbPack` is enabled on an inner group inside a 14 mm × 10 mm board.

The inner group packs R1 partially outside the parent board, producing one `pcb_component_outside_board_error`. The desired zero-error assertion is included as a TODO, and the PCB snapshot captures the invalid placement.

Isolation performed:
- The inner-group reproduction produces one outside-board error.
- Moving `pcbPack` to the board with the same dimensions and components produces zero outside-board errors.
- The inner group has no explicit width/height, so its packing input receives no bounds; the parent board dimensions are not inherited.

This PR intentionally adds only the reproduction for maintainer confirmation.

Verification:
- `bun test tests/repros/repro-pcbpack-inner-group-board-bounds.test.tsx`
- Biome format check for the new test


mohan-bee (commented):
try removing the width and height ,

0hmX (approved):

0hmX:
you will need some more approvals ig @seveibar and @mohan-bee

imrishabh18 (changes requested):
I don't think this is a valid repro, the courtyard and also the pcbGap is insufficient to pack them together. Remove the `width` and `height` and it will work

GokulPandi-M:
> I don't think this is a valid repro, the courtyard and also the pcbGap is insufficient to pack them together. Remove the `width` and `height` and it will work

yeah it is working !! sorry for the confusion closing the pr

PR #3411, opened 2026-08-24
fix: improve phased fanout autorouting

seveibar:
## Summary

- map canonical edge-first fanout positions into explicit solver exit positions while preserving all legacy NinePointAnchor behavior
- route the AM62L upper and lower DDR byte buses to upper and lower bands on the right boundary
- preflight resolved breakout fanout regions and reject phase domains that overlap or lack a via-safe channel
- carry completed source-footprint keepouts into later fanout and global phases
- preserve standalone sequential board fanout phases and the fanout/global endpoint handoff
- reduce the AM62L breakout padding to 2mm on the SoC and 2.5mm on the DRAM, leaving a 2mm routing channel

## Dependencies

- @tscircuit/props ^0.0.634, released from https://github.com/tscircuit/props/pull/812
- @tscircuit/fanout-solver 0.0.36, released from https://github.com/tscircuit/fanout-solver/pull/91

The temporary fanout-solver Git commit pin has been removed. Both Props constraints now use the published release.

## Verification

- fresh bun install resolves @tscircuit/props 0.0.634 and @tscircuit/fanout-solver 0.0.36
- bunx tsc --noEmit --pretty false
- 10 focused fanout, separation, compatibility, and AM62L tests pass with 176 assertions
- AM62L phase 0 asserts all 16 SoC fanout exits lie on maxX, with BYTE0 in the upper band and BYTE1 in the lower band
- bun run build
- bun run smoke-test:dist
- git diff --check

## Snapshot

Updates the AM62L PCB snapshot to show both DDR buses exiting through the right boundary bands before the global route.


seveibar:
CI status: the current test, typecheck, and smoke-test failures all have the same expected dependency cause: published @tscircuit/props does not yet export canonicalBusFanoutDirectionValues or the new edge-first types. The solver is already pinned to PR #91 by immutable commit. Once https://github.com/tscircuit/props/pull/812 merges and publishes, I will bump both Props constraints and rerun the matrix.

seveibar:
NinePointAnchor compatibility follow-up: all nine legacy directives remain on the original inferred/destination-guided path in scalar and object form. The shared center spelling is now kept unconstrained instead of being forwarded as an explicit exitPosition, and the pre-existing center_right integration fixture is restored.

seveibar:
Addressed the keepout-ID parsing concern in ccb53fa3:

- `componentId` is now the sole source-component identity.
- `isFanoutSourceKeepout` explicitly marks the obstacle role.
- phase handoff no longer inspects, slices, prefixes, or deduplicates by `obstacleId`.
- added a regression with opaque keepout IDs and an unmarked prefix-shaped impostor to prevent ID parsing from returning.

Validated with the AM62L repro, the focused metadata regression, and `tsc --noEmit`.

PR #3415, opened 2026-08-24
Reject uniformly undersized autorouter output

seveibar:
## Summary
- validate local autorouter stage output before accepting it, emitting `autorouting:end`, or writing it to the core route cache
- resolve routed connection aliases back to explicitly-sized source traces and reject routes whose body never reaches the supplied minimum width
- cover fresh custom-solver output and poisoned cache hits with a 0.5 mm source trace / 0.1 mm output regression
- update the nRF52810 repro to assert that its known 0.18 mm -> 0.15 mm capacity-autorouter result is rejected and no undersized copper is inserted

Terminal pad-entry tapering remains allowed; the check rejects traces that are uniformly below the source minimum. Capacity-autorouter still needs a separate solver fix to satisfy or reroute hard-width requests.

## Validation
- `bun test` — 1440 pass, 40 skip, 0 fail
- `bunx tsc --noEmit`
- `bun run build`
- `bunx biome format --write ...`
- `git diff --check`


seveibar:
Closing after an architecture review. Core should accept a successful autorouter result and retain its copper; it should not independently reinterpret the result and discard the entire routing stage. The existing checks package already owns requested-versus-routed width as a coordinate-bearing, non-blocking warning (checks #132). This gate also used maximum route-point width, so it was not a sound end-to-end width DRC. Solver-declared failures still propagate normally; this PR's extra core rejection layer is the part being withdrawn.

PR #3417, opened 2026-08-24
Emit hard per-connection minimum trace widths in SRJ

seveibar:
## Summary

- add an explicit per-connection minTraceWidth hard constraint to core's Simple Route JSON boundary
- derive it from source_trace.min_trace_thickness for direct traces and breakout connections
- use the strictest source-trace minimum when several traces are grouped into one net connection
- retain nominalTraceWidth and the deprecated width field as routing preferences/compatibility fields
- raise connection minima when a routing phase specifies a stricter minTraceWidth

This is the core-side SRJ producer change coordinated with tscircuit/tscircuit-autorouter#2221, which makes capacity autorouter enforce the field. It complements the defensive output gate in tscircuit/core#3415.

## Tests

- bun test tests/utils/autorouting/simple-route-json-connection-nominal-trace-width.test.tsx tests/utils/autorouting/simple-route-json-net-min-trace-width.test.tsx tests/features/autoroutingphase-region-reroute.test.tsx
- bunx tsc --noEmit
- bun run build
- biome format on changed TypeScript files
- git diff --check


seveibar:
Closing together with tscircuit/tscircuit-autorouter#2221 after reviewing the established width policy. source_trace.min_trace_thickness is currently carried as the requested/advisory connection nominalTraceWidth, with an undersized result reported as a warning. This PR silently upgraded that existing field into a new solver-fatal hard constraint without a distinct user-facing hard-width property, branch-specific semantics for merged nets, or a terminal neck-down policy. A future hard per-connection contract should begin as an explicit cross-repo design rather than reinterpret the existing requested width.

PR #3446, opened 2026-08-26
fix: synchronize fanout exits across named nets

ShiboSoftwareDev:
Screenshot 2026-08-26 at 10 45 10 AM

## Summary

- matches local fanout and global routing connections by their shared breakout-point identity when named-net expansion gives them different source trace IDs
- updates each global routing endpoint to the compacted fanout exit
- centralizes SRJ point matching with documented board-world coordinates, stable point identity, and position-plus-layer fallback semantics
- adds physical copper-continuity assertions at every handoff
- updates the same 4x4 BGA PCB SVG so R1-R3 visibly route all the way to U1

The otherwise-unconnected resistor pins are compact visual endpoints for this routing reproduction, not a functional pull-up network.

## SVG before and after

- base snapshot in #3445: the three resistor stubs and three BGA fanout traces end at different coordinates
- fixed snapshot in this PR: each resistor route reaches the corresponding fanout exit; the BGA itself and component placement stay unchanged

## Stack

- base repro: #3445
- this PR contains only the handoff fix, continuity assertions, shared SRJ comparison utility, and resulting SVG change

## Validation

- focused visual repro: 1 passed with 9 assertions
- SRJ endpoint comparison contract: 1 passed with 6 assertions
- paired exit-target layer test: 1 passed
- breakout tests: 22 passed, 2 skipped
- memory-capped bun run build: passed, including declaration generation


seveibar (approved):
The windong on that fanout doesn’t look good 🤔

PR #3448, opened 2026-08-26
Pass ground metadata to schematic solver

mohan-bee:


imrishabh18 (commented):
What is the use for this? why only ground and not power?

mohan-bee:
> What is the use for this? why only ground and not power?

i need that metadata for this fix to avoid GND string hacks. 

https://github.com/tscircuit/schematic-trace-solver/pull/919  

i don't find any need for power here right now.

MustafaMulla29 (approved):

imrishabh18 (approved):

Issue #910, opened 2025-06-12
for pinheader with 1 pin schfacingdirection prop does not apply


Anshgrover23:
![Image](https://github.com/user-attachments/assets/db3b7f74-614d-4446-9678-9c58d4baa2e6)

for pinheader with 1 pin schfacingdirection prop does not apply, only the facingDirection applies.
see: 

![Image](https://github.com/user-attachments/assets/dd21f401-ebd5-447d-a32c-aa2e87e248a7)

![Image](https://github.com/user-attachments/assets/4f12fc5b-3d57-4f42-9d72-15a68ff5a2ad)

CC: @seveibar @imrishabh18 ?

PR #3450, opened 2026-08-26
Support inline labels on multi-pin signal nets

MustafaMulla29:
## Dependency

> [!IMPORTANT]
> **tscircuit/schematic-trace-solver#910 must be merged and released before this PR can merge.**

Local verification used `bun link` against solver PR #910 at commit `83a983b`. The machine-local link is intentionally not committed. After #910 is published, this PR should bump `@tscircuit/schematic-trace-solver` to that released version before merge.

## Summary

- allow explicitly named, non-power port-to-net signals spanning multiple components to request inline labels
- preserve explicit `<netlabel>` semantics and the existing power/ground exclusions
- update the merged DRV8305 repro to require all target remote signals inline with zero matching anchored tags
- add a focused three-component signal-net regression test
- update only `toMatchSchematicSnapshot` schematic SVG snapshots
- remove stale anchored-label fallbacks from affected snapshots while retaining their inline labels

## Verification

- linked solver focused suite: 23 passed
- DRV8305 motor repro with linked solver: passed (5 assertions)
- focused multi-pin regression: passed
- semantic SVG audit: the affected auto-layout snapshot changed from one anchored plus three inline `SDA` labels to zero anchored plus three inline `SDA` labels
- targeted affected schematic snapshots regenerated only through `toMatchSchematicSnapshot`
- the large schematic-section auto-layout case writes and matches the corrected snapshot but exceeds its existing 15-second local test timeout
- `bun run format`
- solver `bun run build` (ESM and declarations)
- `git diff --check`

The full core suite was not run because it takes hours locally; snapshot regeneration was limited to tests using `toMatchSchematicSnapshot`.


seveibar (approved):

MustafaMulla29:
Got like 1/2 regressions, will try to fix them

PR #3455, opened 2026-08-26
Compare BGA fanout solvers on AM62L to LPDDR4

seveibar:
## Summary

- add `@tscircuit/bga-fanout-solver` as a GitHub-pinned dev dependency
- extract the existing AM62L-to-LPDDR4 two-byte-bus repro into a shared fixture that accepts a custom fanout `algorithmFn`
- keep the existing `@tscircuit/fanout-solver` regression and add a second regression that runs `FixedTargetBgaFanoutSolver` for both BGA breakouts
- label and snapshot each solver output separately for direct visual comparison

## Comparison result

Both solver paths run the same circuit and the same 136 assertions covering fanout trace count, via count, bus ordering, BYTE0 skew, global winding, source keepouts, phase I/O, and DRC output. After normalizing the on-board solver label, the two generated PCB SVGs are byte-identical.

## Checks

- `bun test tests/repros/repro-am62l-lpddr4-two-bus-fanout.test.tsx`
- `bun test tests/repros/repro-am62l-lpddr4-two-bus-bga-fanout.test.tsx`
- `bunx tsc --noEmit`
- `bun run build`
- Biome formatting and `git diff --check`


seveibar:
CC @0hmX , they are identical????

PR #3460, opened 2026-08-26
Add configurable schematic sheet dimensions

Sang-it:
## Summary

- emit A4 as the default schematic sheet size and support ANSI B
- emit explicit sheet_width and sheet_height values from sheetWidth and sheetHeight props
- use explicit dimensions when checking for elements outside the sheet
- update circuit-json and props to their merged schema releases

## Verification

- targeted schematic sheet tests
- bunx tsc --noEmit
- bun run format
- git diff --check


seveibar (changes requested):
you're not setting sheetWidth/sheetHeight when the user provides only sheetSize

Sang-it:
Addressed: core now resolves the selected sheet size to concrete sheet_width and sheet_height values before emitting schematic_sheet data. Explicit dimensions still override the defaults. ANSI B emission is covered by the focused test.

seveibar (approved):

PR #3470, opened 2026-08-27
fix: scope local autorouting cache by solver

seveibar:
## Summary

- centralize capacity autorouter solver-name resolution without changing the current Pipeline7 default
- version local phase cache keys by resolved solver, capacity depth, target minimum capacity, and effort
- prove explicit Pipeline7 and Pipeline9 renders cannot reuse each other's cached routes

This prepares the cache boundary for a later default-solver change while preserving current routing behavior. No snapshots are changed.

## Validation

- 9 focused cache and solver-selection tests pass
- bunx tsc --noEmit
- bun run build
- targeted Biome formatting check
- git diff --check

### DIFF
--- a/lib/components/primitive-components/Group/Group.ts
+++ b/lib/components/primitive-components/Group/Group.ts
@@ -29,7 +29,10 @@ import type { GraphicsObject } from "graphics-debug"
 import type { PrimitiveComponent } from "lib/components/base-components/PrimitiveComponent"
 import { isAssemblyDeviceContainer } from "lib/components/base-components/is-assembly-device-container"
 import { AutorouterError } from "lib/errors/AutorouterError"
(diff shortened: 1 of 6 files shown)


seveibar:
CI note: all 10 Bun shards fail before test collection during `bun install` because current `main` pins `@tscircuit/copper-pour-solver@https://pkg.pr.new/@tscircuit/copper-pour-solver@85`, which currently fails to resolve. I reran the failed workflow and reproduced the same dependency-resolution failure. Format, typecheck, smoke build, and the focused local cache-identity suite are green; no shard reached this PR’s tests.

seveibar:
Rebased onto current `main` v0.0.1784 and revalidated the exact tree: all 3 focused cache suites pass (29 assertions), TypeScript and ESM+DTS build pass, and the worktree is clean. Fresh hosted CI is running.

PR #3471, opened 2026-08-27
Prevent duplicate preloaded traces between routing phases

seveibar:
## Summary

- replace accumulated PCB traces by `pcb_trace_id` when a routing stage returns preloaded traces
- use the same replacement boundary before reconnecting a rerouted region
- preserve multiple split sections returned by one stage, even when they share a trace ID
- add a fast Pipeline9-like region integration regression plus Arduino logical-route uniqueness coverage

## Why

Pipeline9 can return its preloaded traces together with newly routed traces. `Group` previously prepended the same preloaded traces again in follow-up-stage and region-reroute paths, materializing duplicate copper. The mocked region regression produced 9 traces before this fix and 6 unique traces after it.

This uses the existing trace-ID replacement helper at the two accumulation boundaries instead of silently canonicalizing or deduplicating malformed stage output.

## Validation

- `bun test tests/utils/autorouting/accumulate-stage-output-traces.test.ts`
- `bun test tests/features/autoroutingphase-reroute-preloaded-traces.test.tsx`
- `bun test tests/features/autorouter-fanout.test.tsx`
- `bun test tests/features/autorouter-beta-pipeline9-multiple-phases.test.tsx`
- `bun test tests/repros/repro-gameboy-direct-subcircuit-routing.test.tsx`
- `bun test tests/repros/repro116-arduino-uno-reroute.test.tsx`
- `bun test tests/features/autoroutingphase-reroute.test.tsx`
- `bun test tests/features/autoroutingphase-imported-circuit-json-reroute-builtin.test.tsx`
- `bunx tsc --noEmit`
- `bun run build`
- `git diff --check`

### DIFF
--- a/lib/components/primitive-components/Group/Group.ts
+++ b/lib/components/primitive-components/Group/Group.ts
@@ -1618,7 +1618,11 @@ export class Group<Props extends z.ZodType<any, any, any> = typeof groupProps>
         if (transformedSimpleRouteJson?.traces) {
           stageOutputTraces = transformedSimpleRouteJson.traces
         } else if (usesPreviousStageOutput) {
-          stageOutputTraces = [...(simpleRouteJson.traces ?? []), ...traces]
+          stageOutputTraces =
+            getAccumulatedPcbTracesWithStageOutputReplacements({
+              accumulatedPcbTraces: simpleRouteJson.traces ?? [],
+              stageOutputPcbTraces: traces,
+            })
         }
         const outputSimpleRouteJson = {
           ...(transformedSimpleRouteJson ?? simpleRouteJson),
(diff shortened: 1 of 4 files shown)


seveibar:
Vercel failure is also before build/test collection: `bun install` cannot resolve the `@tscircuit/copper-pour-solver@https://pkg.pr.new/@tscircuit/copper-pour-solver@85` pin inherited from current `main`. All 10 hosted Bun shards, format, typecheck, and smoke build are green.

seveibar:
Rebased onto current `main` v0.0.1784 and revalidated the exact tree: mocked Pipeline9-like region reroute, Arduino repro116, and accumulation helper all pass (3 tests / 36 assertions); TypeScript and ESM+DTS build pass; `git diff --check` is clean. Fresh hosted CI is running.

PR #3478, opened 2026-08-27
Add AM62L LPDDR4 address-control fanout bus

seveibar:
## Summary

- add the AM62L/LPDDR4 address-control fanout bus alongside both byte buses
- route CA0–CA5, CS, and CKE using their verified package balls on a dedicated inner6 signal layer
- preserve all 212 GND/DDR-power plane dogbones and through-all via checks
- enforce phase-local skew limits for all three buses and prove the straight global handoff has no winding conflicts
- rename the shared fixture and regular repro from two-bus to generic/three-bus naming while keeping the BGA-solver comparison signal-only
- use released `@tscircuit/fanout-solver@0.0.40`, which contains fanout-solver PR #99

## Validation

- three-bus repro on the released solver: 1 pass, 5,045 assertions, 50.2 s test body / 50.7 s wall
- adjacent plane-termination and fanout/global-handoff regressions: 3 pass, 60 assertions
- `bunx tsc --noEmit`
- dependency install resolved `@tscircuit/fanout-solver@0.0.40`
- `git diff --check`

## Routed groups

- DDR_BYTE0: DQ0–DQ7, local fanout skew <= 8 mm
- DDR_BYTE1: DQ8–DQ15, local fanout skew <= 14.5 mm
- DDR_ADDR_CTRL: CA0–CA5, CS, CKE, local fanout skew <= 15 mm

Clock remains a separate differential timing class and reset remains a separate control trace; those are intentionally not folded into this single-ended ADDR_CTRL bus.


seveibar:
CC @0hmX

PR #3483, opened 2026-08-27
BGA fanout vs default fanout example

0hmX:
Follow-up to merged #3455. The linked timeline event was the merge event, and the original PR checks were green.

The existing two-bus comparison remains intact and now exercises the intended eight-layer power-plane input. This PR also adds a separate source-routed regression for the validated full AM62L-to-LPDDR4 board instead of replacing or weakening either existing test.

Changes:
- preserve both existing two-bus tests and snapshots
- keep the existing BGA comparison aligned with the eight-layer GND and LPDDR4 power-pour input
- add a dedicated 60 x 30 mm full-board TSX fixture at the validated U1/U2 positions and 30.675 mm center separation
- route all three 11-net DDR buses through two FixedTargetBgaFanoutSolver breakouts plus the capacity channel autorouter
- pin @tscircuit/bga-fanout-solver to merged PR #9 commit e67a81f0 and use only public package dependencies
- assert 33/33 connections, four GND layers, 243 PCB traces, 138 TOP-to-BOTTOM vias, 144 explicit GND pad connections, 72 GND via drops, 0.086788 mm minimum via-to-pad clearance, and no routing/clearance errors
- add a dedicated full-board PCB snapshot generated from TSX; no precomputed Circuit JSON is imported

The original 42 x 26 mm comparison board and 19.116917 mm U1/U2 spacing remain unchanged. Its regular and BGA snapshots are byte-identical to their pre-full-board hashes.

Verification:
- bun test tests/repros/repro-am62l-lpddr4-full-bga-fanout.test.tsx
- bun test tests/repros/repro-am62l-lpddr4-two-bus-fanout.test.tsx tests/repros/repro-am62l-lpddr4-two-bus-bga-fanout.test.tsx
- bun test tests/utils/autorouting/simple-route-json-unbroken-copper-pour-obstacles.test.tsx tests/examples/example39-unbroken-copper-pour-escape-vias.test.tsx tests/features/autorouter-fanout-plane-termination.test.tsx
- bunx tsc --noEmit
- bun run format
- bunx biome format .


tests/fixtures/am62l-lpddr4-full-bga/generate-am62l-altium-reference.ts
```diff
@@ -0,0 +1,604 @@
+import { basename, dirname, join, resolve } from "node:path"
```
seveibar, line 1: ~~can you remove this script? it is a bad reference for core i think~~

seveibar: just locate this script inside the tests/fixtures/am62l-lpddr4-full-bga directory

seveibar (approved):
see important comments

also i'm seeing the default fanout solver is able to solve this board, which makes me a bit skeptical about the configuration

CleanShot 2026-08-27 at 09 35 57@2x

PR #3491, opened 2026-08-27
chore: update schematic-trace-solver to 0.0.170

seveibar:
## Summary
- update `@tscircuit/schematic-trace-solver` from `^0.0.168` to `^0.0.170`
- regenerate the visual snapshots against the new solver; existing schematic snapshots remain current, so no snapshot files changed

## Testing
- `bunx tsc --noEmit`
- `bun run build`
- 11 focused schematic snapshot tests (all passed)
- `BUN_UPDATE_SNAPSHOTS=1 bun test --timeout 120000` (1,476 passed, 40 skipped; 2 unrelated inline snapshots failed because Bun 1.3.2 serializes a tiny float in decimal rather than scientific notation)

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,7 @@
     "@tscircuit/ngspice-spice-engine": "^0.0.20",
     "@tscircuit/props": "^0.0.641",
     "@tscircuit/schematic-match-adapt": "^0.0.18",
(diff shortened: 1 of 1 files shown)


seveibar:
CC @Sang-it this is a bump for your version

PR #3504, opened 2026-08-28
Render inline labels across sections on the same sheet

MustafaMulla29:
## Summary

- render canonical inline net labels when connected endpoints span schematic sections on the same schematic sheet
- aggregate aliases for one electrical net into a single solver connection so every disconnected endpoint gets the canonical label text
- keep connections across different schematic sheets on the anchored net-label path
- preserve explicit net-label behavior, source-trace ownership, and anchored fallback placement

## Validation

- `bun run build`
- `bunx tsc --noEmit`
- `bunx biome check` on changed files
- existing section-boundary, subcircuit-boundary, and RP2040 repro tests
- cross-sheet repro140 and repro173 regression tests

The affected existing schematic snapshots were regenerated by CI and audited; redundant standalone feature tests and snapshots were removed.


tests/components/primitive-components/__snapshots__/schematic-section-2-schematic.snap.svg
MustafaMulla29, line 1: This might be a schematic-trace-solver issue

tests/repros/__snapshots__/repro170-rp2040-u1-complete-connections-schematic.snap.svg
MustafaMulla29, line 1: This as well might be a trace-solver issue

seveibar (commented):
@MustafaMulla29 so inline net labels ARE allowed if both items are on the same sheet, but NOT ALLOWED across different sheets- i think you're missing that distinction?

MustafaMulla29:
That distinction makes sense. I have updated the implementation so inline-label eligibility now explicitly requires both endpoints to have the same schematic_sheet_id and different schematic-section identities. Connections across different sheets remain anchored. I also consolidated same-net aliases under the canonical displayed net label, removed the trace-solver experiment/link, and kept snapshot updates out of this PR.

tests/features/subcircuit-circuit-json/__snapshots__/subcircuit-circuit-json13-schematic.snap.svg
MustafaMulla29, line 1: This change is intentional: the net crosses the S1 subcircuit boundary, so both schematic fragments need reciprocal anchored labels—R2_pin1 inside S1 and R1_pin1 beside R2.

seveibar (approved):

PR #3511, opened 2026-08-28
Fix FabricationNoteText layer handling

rushabhcodes:
## Why

FabricationNoteText currently writes every pcb_fabrication_note_text element to the top layer, even when the JSX specifies layer="bottom" or the note belongs to a component whose footprint is flipped to the bottom side.

That produces incorrect Circuit JSON for bottom-side fabrication notes. Downstream renderers and fabrication tooling receive the note on the wrong side, and importing existing Circuit JSON also loses the original fabrication-note text layer.

## What changed

- Resolve FabricationNoteText from its layer prop instead of hard-coding top.
- Apply the existing PCB primitive layer-flip helper so notes follow bottom-side component and footprint placement consistently with other fabrication-note primitives.
- Validate that the resolved fabrication-note text layer is top or bottom.
- Preserve pcb_fabrication_note_text.layer when rebuilding components from Circuit JSON.
- Add a visual regression test covering both an explicit bottom layer and a note inherited through a bottom-layer component container.

## Testing

- bun test tests/components/primitive-components/fabrication-note-text-layer.test.tsx tests/components/primitive-components/fabrication-note-text-rotation.test.tsx tests/components/normal-components/fabrication-notes.test.tsx
- bunx tsc --noEmit


seveibar (approved):
you're showing a repro right? fabricationnotetext in a group doesn't flip properly?

rushabhcodes:
> you're showing a repro right? fabricationnotetext in a group doesn't flip properly?

even normally it does not flip

rushabhcodes:
top layer was hard coded

PR #3518, opened 2026-08-29
fix: render polarity pin label suffixes as symbols

rushabhcodes:
## Why

Pin labels serve two different roles in core:

- They are identifier-safe aliases used by selectors and connectivity.
- They are also used as the visible text rendered on schematic pins.

The identifier form intentionally allows letters, numbers, and underscores, so common polarity labels are written as `IN_POS` and `IN_NEG` instead of `IN+` and `IN-`. Rendering those identifiers verbatim makes schematics less conventional and harder to scan, even though keeping them identifier-safe is useful for code and selectors.

Core is the boundary that translates source pin identifiers into Circuit JSON display fields. This PR therefore preserves the original labels in `source_port` names, port hints, and selectors, while formatting only the emitted schematic display label:

- `IN_POS` renders as `IN+`
- `IN_NEG` renders as `IN-`
- Labels without a trailing `_POS` or `_NEG` are unchanged

This keeps expressions such as `U1.IN_POS` stable and avoids widening the pin-label identifier syntax or requiring renderers to reinterpret source-domain identifiers.

## What changed

- Added one shared schematic pin-label display formatter.
- Applied it to schematic boxes, regular/custom ports, and schematic-symbol ports so all core emission paths behave consistently.
- Added a visual regression test that verifies the rendered symbols and confirms the original selectors still resolve.
- Updated existing snapshots that intentionally contain `_POS` and `_NEG` pin labels.

## Verification

- `bun test tests/components/normal-components/chip-polarity-pin-label-display.test.tsx`
- Relevant schematic and pin-label regression suite: 11 tests passed
- `bunx tsc --noEmit`
- Biome checks for all modified TypeScript files


seveibar (changes requested):
i don't agree with this change. It looks nice but can create confusion since IN+ and IN- are difference, the user will need to add these aliases manually

PR #3522, opened 2026-08-29
fix: keep explicit labels inline on branched traces

### LINKED ISSUE #3521: Trace schDisplayLabel on branched connectivity becomes schematic_net_label
## Problem

A real named `<trace schDisplayLabel="V+">` on branched connectivity is emitted as a `schematic_net_label`, even though the TSX contains no `<netlabel>` component. This happens in both standalone and composed subcircuit renders.

It makes it impossible to reproduce source schematics that put a short signal name over a real branched wire without changing it into a net-label endpoint. This was isolated while reproducing the V+ and V- motor-current/shunt input traces in `tscircuit/ti#134`.

## Reproduction

```tsx
<trace
  name="V-PLUS-INPUT"
  path={["J1.V_PLUS", "R_SHUNT.pin2", "R_FILTER.pin1"]}
  schDisplayLabel="V+"
/>
```

The same result occurs when the three-terminal connectivity is expressed as a displayed two-endpoint trace plus a second trace joining the shunt.

## Expected

The render contains `schematic_text` owned by the real source trace:

```json
{
  "type": "schematic_text",
  "text": "V+",
  "source_trace_id": "..."
}
```

## Actual

```json
{
  "type": "schematic_net_label",
  "text": "V+",
  "source_trace_id": null,
  "source_net_id": "...connectivity_net0"
}
```

The label changes display semantics and is placed like a net-label endpoint rather than over the trace. In a composed parent the same element remains a `schematic_net_label` and only the connectivity-net identifier is remapped.

## Version

- `@tscircuit/core`: 0.0.1787
- `tscircuit`: 0.0.2460

Integration: https://github.com/tscircuit/ti/pull/134

Abse2001:
## Summary

- allow an explicit trace `schDisplayLabel` to opt one branch of a multi-port connectivity net into inline labeling
- retain the exact source-trace label only in core's solver-input metadata and strip it before invoking the schematic solver
- assign a multi-point path's explicit label to its first expanded branch so the text is not duplicated
- add a focused three-terminal shunt/filter regression and schematic snapshot

## Motivation

`schDisplayLabel` already means “draw this name on the trace” for two-terminal connections. The inline-label eligibility check rejected the same explicit intent whenever the electrical connectivity net had a third tap, causing a tag-shaped `schematic_net_label` instead of trace-owned `schematic_text`.

This was isolated while reproducing the V+ and V- shunt-input traces in https://github.com/tscircuit/ti/pull/134.

Closes #3521.

## Validation

- focused branched trace regression passes and emits one trace-owned `V+` schematic text with no `V+` schematic net label
- all seven schematic inline-label tests pass
- trace display-label precedence and ground-symbol tests pass
- `bunx tsc --noEmit`
- `bun run build`
- visually inspected the new schematic snapshot


tests/examples/__snapshots__/example11-net-symbol-schematic.snap.svg
imrishabh18, line 1: Why in this schematic there is no `GND` symbol? All of them are inline netLabel, but I think some of them should have the downward `GND` symbol rather

Abse2001: Correction: GND1, GND2, and GND3 in this example are user-supplied schDisplayLabel strings, not connections to a source net marked as ground. Keeping those labels inline is intended. Actual ground nets remain excluded from inline-label eligibility and still render as rail_down symbols, including when the trace has schDisplayLabel="GND"; the dedicated ground-symbol tests pass. I reverted my previous change in c0f20a1b.

Abse2001:
Closing this approach after validating the minimal circuit against the latest Core and tscircuit.com. Explicitly declaring the nets with isGroundNet and isPowerNet correctly produces rail_down and rail_up symbols, respectively. The solver repro did not preserve that real Core metadata, so it was not a valid reproduction of the product behavior. No fix is needed from this PR.

PR #3529, opened 2026-08-29
Set default footprint silkscreen text size to 0.6mm

mohan-bee:
### Motivation

keep the default font size from 0.4 to 0.6

image


seveibar (changes requested):
the change belongs in footprinter, and no modification should be made on the footprinter font size

mohan-bee:
Closing because this change belongs only in Footprinter. Replaced by https://github.com/tscircuit/footprinter/pull/859

PR #3538, opened 2026-08-30
Preserve custom fanout algorithms during preset normalization

0hmX:
## Summary
- preserve supplied configuration fields such as `algorithmFn` when normalizing fanout presets
- treat a custom fanout algorithm as the complete routing stage instead of appending the built-in follow-up router
- add normalization and end-to-end dispatch regression coverage
- mark the AM62L two-bus BGA repro as `test.failing` now that it reaches the real solver
- remove the invalid snapshot that was generated by the built-in fanout solver

## Why
`getPresetAutoroutingConfig` reconstructed fanout preset objects from fixed fields and silently discarded `algorithmFn`. Core therefore instantiated the built-in fanout solver even when callers explicitly supplied a custom solver.

After this fix, the AM62L repro genuinely invokes `FixedTargetBgaFanoutSolver`, which currently fails to finish within its iteration budget. The expected-failure test records that solver limitation without retaining misleading default-solver output.

## Validation
- `bun test tests/repros/repro-am62l-lpddr4-two-bus-bga-fanout.test.tsx tests/utils/autorouting/getPresetAutoroutingConfig.test.ts tests/features/custom-fanout-autorouting-algorithmfn.test.tsx`
- `bunx tsc --noEmit`


0hmX:
The AM62L two-bus repro now reaches `FixedTargetBgaFanoutSolver` instead of silently falling back to the built-in Core fanout solver. It remains `test.failing` until the fixed-target solver completes this fixture within its iteration budget. The previous snapshot was removed because it depicted output from the fallback solver, not the requested BGA solver.

PR #3553, opened 2026-08-31
fix(netlabel): preserve projection identity for shared schematicsymbol ports

### LINKED ISSUE #3505: Net labels lose identity across shared schematicsymbol supply projections
## Description

Two independently placed native `schematicsymbol` projections can map their V+/V- symbol ports to the same physical chip supply pins (for example, the A and B units of one dual op amp). Native `netlabel` connections to both projections preserve source connectivity, but schematic rendering does not keep the projection identity consistently.

Observed with `tscircuit@0.0.2460` / `@tscircuit/core@0.0.1787` while implementing the TLV2316-Q1 stage from TI TIDA-01421.

## Minimal shape

```tsx
<net name="V5" isPowerNet />
<net name="GND" isGroundNet />

<chip name="U3" noSchematicRepresentation pinLabels={dualOpAmpPins} />

<schematicsymbol
  name="U3A"
  chipRef=".U3"
  symbolName="opamp_with_power_right"
  connections={{
    inp1: "U3.pin3",
    inp2: "U3.pin2",
    out: "U3.pin1",
    "V+": "U3.pin8",
    "V-": "U3.pin4",
  }}
  schX={0}
  schY={0}
/>
<schematicsymbol
  name="U3B"
  chipRef=".U3"
  symbolName="opamp_with_power_right"
  connections={{
    inp1: "U3.pin5",
    inp2: "U3.pin6",
    out: "U3.pin7",
    "V+": "U3.pin8",
    "V-": "U3.pin4",
  }}
  schX={5}
  schY={0}
/>

<netlabel net="V5" connection="U3A.pin5" anchorSide="bottom" schX={-0.03} schY={0.39} />
<netlabel net="GND" connection="U3A.pin3" anchorSide="top" schX={-0.02} schY={-0.39} />
<netlabel net="V5" connection="U3B.pin5" anchorSide="bottom" schX={4.97} schY={0.39} />
<netlabel net="GND" connection="U3B.pin3" anchorSide="top" schX={4.98} schY={-0.39} />
```

## Actual behavior

- U3A and U3B projection ports correctly share the physical U3 source-port IDs.
- Only one projection's shared V+/V- `schematic_port.is_connected` values are set. The other projection remains false even though the physical pin is on the correct source net.
- In a parent composite using `exposedNets`, the automatically generated child-to-parent power-net trace can add a second rail label at a different projection. For GND the generated text can be inferred as `U2_GND`, overlapping an explicit native `GND` symbol.
- Adding direct same-net traces does not reliably fix the projection flag; it can instead create another automatically generated rail label.
- If the four `netlabel` elements omit `schX`/`schY` so they should attach directly to the connected projection ports, rendering throws `Couldn't find position for schematic_port` during `NetLabel.computeSchematicPropsTransform`, because the `schematicsymbol` port position is not yet available in that phase.

## Expected behavior

- `connection="U3A.pin5"` and `connection="U3B.pin5"` retain the selected visual projection through schematic rendering, even though both map to physical U3 pin 8.
- Each explicit native rail symbol attaches to and marks its selected projection port connected.
- Exposing the child V5/GND nets does not generate another endpoint label over an explicit label.
- A coordinate-free `netlabel` connected to a `schematicsymbol` port waits for or derives the native port position instead of throwing.

## Downstream workaround

The downstream reference circuit renders the rail labels at exact native-stem coordinates and, in the composite only, places the six labels in the parent wrapper so all automatically exposed endpoints reuse the plain `V5`/`GND` names. No fake nets or decorative wires are used. This produces correct connectivity and clean output, but core should preserve projection identity directly.

marcos452652258-gif:
## Summary

Fixes #3505.

Two problems when netlabels connect to schematicsymbol projections that share physical source ports (e.g. both units of a dual op-amp mapping V+/V- to the same chip pins):

### 1. Coordinate-free netlabels crashed

`<netlabel net="V5" connection="U3A.pin5" />` (no `schX`/`schY`) threw `Couldn't find position for schematic_port` from `NetLabel.computeSchematicPropsTransform`, because the pre-layout position of a schematicsymbol projection port is not available at that phase.

**Fix:** use the already-rendered `schematic_port.center` when available, falling back to the pre-layout position (plain schematic boxes).

### 2. Only one projection's shared power/ground port was marked connected

The second projection's V+/V- `schematic_port.is_connected` stayed `false` because an existing source_trace caused an early `continue` before the update.

**Fix:** mark the projection port connected whenever its netlabel resolves, even when the trace already exists; coordinate-free netlabels also mark their connected ports.

## Tests

- `schematic-symbol-3505.test.tsx` — dual op-amp (U3A/U3B) with shared V+/V- pins:
  - coordinate-free netlabels render (no throw) and all 4 V+/V- schematic ports are connected;
  - explicit-coordinate netlabels mark both projections' shared power ports connected.

## Verification

- New tests pass (2/2).
- Existing `netlabel*` tests: no new failures (the 2 pre-existing failures on `main`, `netlabel-center` snapshot 0.02% and `netlabel-connection` timeout, reproduce unchanged without this patch).

### DIFF
--- a/lib/components/primitive-components/NetLabel.ts
+++ b/lib/components/primitive-components/NetLabel.ts
@@ -92,8 +92,17 @@ export class NetLabel extends PrimitiveComponent<typeof netLabelProps> {
     if (props.schX === undefined && props.schY === undefined) {
       const connectedPorts = this._getConnectedPorts()
       if (connectedPorts.length > 0) {
-        const portPos =
-          connectedPorts[0]._getGlobalSchematicPositionBeforeLayout()
+        // For ports whose schematic position comes from a schematicsymbol
+        // projection (schematicSymbolPortDef), the pre-layout position is not
+        // available until the symbol's schematic ports have been created. Use
+        // the rendered schematic_port center when available, and fall back to
+        // the pre-layout position otherwise (e.g. plain schematic boxes).
+        let portPos = this._tryGetPortSchematicPosition(
+          connectedPorts[0],
+        )
+        if (!portPos) {
+          portPos = connectedPorts[0]._getGlobalSchematicPositionBeforeLayout()
+        }
         const parentCenter = applyToPoint(
           this.parent?.computeSchematicGlobalTransform?.() ?? identity(),
           { x: 0, y: 0 },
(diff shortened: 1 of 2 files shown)


marcos452652258-gif:

marcos452652258-gif:

marcos452652258-gif:

marcos452652258-gif:

PR #3562, opened 2026-08-31
Only use paired breakouts as fanout exit targets

ShiboSoftwareDev:
Stops ordinary component pins from becoming fanout exit targets while preserving paired-breakout guidance. The stacked AM62L snapshot changes from 0/3 routed to fully routed.


seveibar (changes requested):
bad repro, use tsx

ShiboSoftwareDev:
Addressed: removed the hand-authored SimpleRouteJson test and moved paired exit-target coverage into the existing two-breakout TSX integration test. The stacked AM62L repro remains a real TSX board with its PCB snapshot.

seveibar (approved):

PR #3570, opened 2026-09-01
Avoid duplicating preloaded fanout traces

ShiboSoftwareDev:
Follow-up solvers already return their preloaded traces, so Core no longer appends them a second time. The stacked PCB snapshot changes the real HDMI circuit from unrouted to routed.


lib/components/primitive-components/Group/Group.ts
```diff
@@ -1617,8 +1617,6 @@ export class Group<Props extends z.ZodType<any, any, any> = typeof groupProps>
         let stageOutputTraces = traces
         if (transformedSimpleRouteJson?.traces) {
           stageOutputTraces = transformedSimpleRouteJson.traces
-        } else if (usesPreviousStageOutput) {
-          stageOutputTraces = [...(simpleRouteJson.traces ?? []), ...traces]
```
ShiboSoftwareDev, line 1621: Pipeline 9 already returns these preloaded fanout traces in its stage output. Concatenating them here creates duplicate pcb_trace_id entries; the existing stage accumulator preserves prior traces.

PR #3576, opened 2026-09-01
repro: capacitor bounds do not match their rotated orientation

GokulPandi-M:
## Motivation

Overlapping capacitor labels make the schematic difficult to read. This repro captures a bounds mismatch after capacitor rotation so it can be investigated and used to verify a future fix.

Screenshot 2026-09-01 at 10 26 16 PM

## Current behavior

In the LED Blinker schematic, C1 and C3 automatically rotate to 270°, but their reserved layout bounds are smaller than the space occupied by their symbols and reference/value text. Their rendered bounds also overlap.

## Expected behavior

The reserved layout space should account for the capacitor’s rotated orientation and its text.

## Changes

- Adds a reproduction using the LED Blinker schematic.
- Checks the reserved sizes against the rendered bounds and confirms the overlap.
- Includes a schematic snapshot with bounding boxes.


imrishabh18 (commented):
The description and the PR title is confusing, it's simply the bounds are not being rotated for the capacitor's on rotation

GokulPandi-M:
> The description and the PR title is confusing, it's simply the bounds are not being rotated for the capacitor's on rotation

Sorry for the confusions @imrishabh18 , the issue is like after the capacitors rotate, the space reserved for them is smaller than the space their symbols and text actually occupy.

tests/repros/__snapshots__/repro-led-blinker-matchpack-input-under-reserves-rotated-text-schematic.snap.svg
imrishabh18, line 1: Not a bug, seems okay to me, move them apart so there is no confusion.

imrishabh18 (changes requested):

PR #3587, opened 2026-09-02
Route the full OSM-S AM62L fanout

ShiboSoftwareDev:
Pins tscircuit/fanout-solver#136 so the real OSM-S AM62L fanout repro routes all 96 connections. Existing fanout snapshots remain unchanged.


ShiboSoftwareDev:
Superseded by the corrected core stack in #3595 and #3596. The replacement keeps the E2E test separate and makes the Pipeline 10 integration change in core.

PR #3595, opened 2026-09-02
Add AM62L LPDDR4 Pipeline 10 E2E

ShiboSoftwareDev:
## Summary

- add an end-to-end TSX circuit using the repository's full 373-ball AM62L32 and 200-ball MT53E1G16D1ZW package models
- connect four real LPDDR4 byte-0 data signals as a bus on an 8-layer HDI board
- render the circuit through core's complete pipeline and store the processed PCB SVG snapshot

## Why

This provides a realistic visual control for the dedicated two-BGA Pipeline 10 flow. Core currently does not recognize the `beta-pipeline10` preset, so this PR records the existing fallback routing. The stacked fix will register the shipped Pipeline 10 solver and update only the generated PCB SVG.

## Testing

- `bun test tests/repros/repro-am62l-lpddr4-pipeline10-e2e.test.tsx`
- `bunx tsc --noEmit`

PR #3596, opened 2026-09-02
Enable Pipeline 10 BGA fanout preset

ShiboSoftwareDev:
## Summary

- register the shipped `AutoroutingPipelineSolver10_BgaFanout` in core's solver catalogue
- select it when a circuit uses `autorouter="beta-pipeline10"`
- update only the E2E's processed PCB SVG; the test source is unchanged in this PR

## Visual result

The base snapshot falls through to Pipeline 7 and draws four direct routes through the AM62L escape area: 0 vias and two PCB trace DRC errors (trace spacing and accidental pad contact). With this change, Pipeline 10 detects the two BGAs, fans out both packages, and globally routes the four LPDDR4 byte-lane signals using 8 vias.

A stricter verification run reported no autorouting, connectivity, pad-to-trace, via-to-via, or via-to-trace errors. The stacked SVG diff shows the routing change directly.

## Stack

Depends on #3595.

## Testing

- `bun test tests/repros/repro-am62l-lpddr4-pipeline10-e2e.test.tsx`
- `bunx tsc --noEmit`
- `bun run build`


ShiboSoftwareDev:
CI note: all relevant checks and the other nine test shards pass. Test shard 9 is the same unrelated schematic warning assertion already fixed by #3594. The Pipeline 10 E2E itself passes in CI, and this stacked PR has no test-source changes.

ShiboSoftwareDev:
Closing because registering Pipeline 10 is not the production DDR fix and the resulting routing is not acceptable. The replacement work targets the Pipeline 9 fanout-to-global handoff in core.

PR #3601, opened 2026-09-02
Route the OSM-S carrier fanout

ShiboSoftwareDev:
Pins tscircuit/fanout-solver#136. The realistic carrier repro changes from 95/96 to 96/96 routed connections without clearance errors.


ShiboSoftwareDev:
Closing until core #3594 fixes the unrelated Bun shard-9 failure currently present on main; this stack should not be presented with red CI.

PR #3614, opened 2026-09-03
reproduce missing solder jumper pcb ports

mohan-bee:
### Motivation
- Capture the failure caused by solder jumpers without explicit footprints.

### Before
- A footprint-less solder jumper creates no PCB ports and autorouting throws while resolving connected ports.

### After
- A TSX-only expected-failing test and PCB snapshot reproduce the missing jumper pads.


mohan-bee:
Superseded by direct upstream repro PR #3616.

PR #3615, opened 2026-09-03
prevent autorouter crashes when solder jumper footprints are omitted

mohan-bee:
**motivation**
solder jumpers without explicit footprints never inflate pcb pads or ports, so connected tsx traces reach an undefined endpoint and the autorouter throws while reading its coordinates.

**before**
the repro creates 0 pcb ports for jp1, logs an autorouting error, and the pcb snapshot contains only the two resistors with no jumper between them.

**after**
solder jumpers derive `solderjumper2` or `solderjumper3` footprints from their pin configuration and reuse bridge resolution for normally closed variants; jp1 now has 2 pcb ports, the routed repro has 0 autorouting errors, and the real pcb snapshots show open and bridged two-pin and three-pin defaults.


mohan-bee:
Superseded by direct upstream stacked fix PR #3617.

PR #3617, opened 2026-09-03
prevent autorouter crashes when solder jumper footprints are omitted

mohan-bee:
**motivation**
solder jumpers without explicit footprints never inflate pcb pads or ports, so connected tsx traces reach an undefined endpoint and the autorouter throws while reading its coordinates.

**before**

`<solderjumper name="JP1" pinCount={2} bridged={false} />`

the stacked repro creates 0 pcb ports for jp1, logs an autorouting error, and its pcb snapshot contains only the two resistors with no jumper between them.

**after**

`<solderjumper name="JP1" pinCount={2} bridged={false} />`

image

solder jumpers derive their footprinter string directly from the typed `pinCount`, defaulting to `solderjumper2`; jp1 now has 2 pcb ports, the routed repro has 0 autorouting errors, and the updated real pcb snapshot shows the open two-pad jumper between both resistors.


imrishabh18 (changes requested):
We don't want to imply any footprint if the footprint prop is not passed. The circuit json should have the `pcb_footprint_missing_error` in this case and that's good

mohan-bee:
> We don't want to imply any footprint if the footprint prop is not passed. The circuit json should have the `pcb_footprint_missing_error` in this case and that's good

i don't find any footprint error 

image

PR #3640, opened 2026-09-04
reproduce missing schematic section dividers for small gaps

mohan-bee:
**motivation**
the real kicad benchmark board declares four schematic sections, but the divider calculation treats their distinct component bounds as overlapping.

**before**
the real board emits zero schematic divider lines, leaving its four intended regions visually unpartitioned in the schematic svg.

**after**
a focused 23-line regression records the zero-divider failure with numeric assertions and the complete real board schematic svg.


mohan-bee:
migrated to direct branch pr #3642

PR #3643, opened 2026-09-04
render schematic section dividers when margins overlap

mohan-bee:
### Motivation
- preserve schematic section boundaries when the one millimeter cell margin connects otherwise separate section contents

### Before
- the real benchmark board feeds one overlap cluster into the boundary solver and emits zero dividers

### After
- detect the connected expanded input and use raw component bounds, producing three dividers and four visible regions


lib/components/primitive-components/SchematicSection.ts
```diff
@@ -135,14 +135,18 @@ export class SchematicSection extends PrimitiveComponent<
...
-    const dividers = calculateCellBoundaries(
+    let dividers = calculateCellBoundaries(
       allSectionsWithBounds.map((s) => ({
         minX: s.rawBounds.minX - CELL_MARGIN,
         maxX: s.rawBounds.maxX + CELL_MARGIN,
         minY: s.rawBounds.minY - CELL_MARGIN,
         maxY: s.rawBounds.maxY + CELL_MARGIN,
       })),
     )
+    if (dividers.length === 0)
```
imrishabh18: Why only `0`? very confusing.

imrishabh18 (changes requested):
Seems like a hack, the actual problem may belong somewhere else

PR #3646, opened 2026-09-04
use library cell margins to preserve schematic section dividers

mohan-bee:
**motivation**
compact schematic sections lose their dividers when core expands their bounds before boundary calculation.

**before**
the real benchmark's four expanded sections produce zero dividers in the reproduction snapshot.

**after**
core uses the published calculate-cell-boundaries 0.0.21 release and passes raw section bounds with a one millimeter preferred margin. the library owns margin handling. the real-board snapshot shows three dividers forming four sections, and the three protected section snapshots remain unchanged.


imrishabh18 (commented):
Add snapshot of the reference you are trying to fix and match with

seveibar (commented):
doesn't this fix belong in caclulateCellBoundaries? we don't want algorithm code in core generally

imrishabh18 (approved):

PR #3651, opened 2026-09-04
add four-pin pushbutton self-connection repro

mohan-bee:
### Motivation
- reproduce a four-pin pushbutton whose rendered contacts resolve to the same net when internal pin groups are omitted.

### Before
- the schematic silently draws wake_mr_n through the switch back to wake_mr_n.

### After
- a focused assertion and schematic snapshot capture the one-connectivity-key failure for the follow-up fix.


mohan-bee:
Superseded by the reduced Circuit JSON repro in tscircuit/checks#262.

PR #3652, opened 2026-09-04
report pushbuttons connected to themselves

mohan-bee:
### Motivation
- prevent a pushbutton whose two schematic contacts resolve to one net from passing silently.

### Before
- the four-pin repro renders wake_mr_n through the switch back to wake_mr_n with zero component errors.

### After
- core emits one fatal component-configuration error naming the internal-pin and footprint mapping, while valid internally grouped pushbuttons remain unchanged.


mohan-bee:
Superseded by tscircuit/checks#263, where post-render netlist validation is owned.

PR #3654, opened 2026-09-04
feat: use Pipeline 9 as the default autorouter

seveibar:
Use Pipeline9 for the implicit local default and `autorouterVersion="latest"`, while preserving explicit `beta_pipeline7` selection.

Update `@tscircuit/capacity-autorouter` to `^0.0.885`, including the merged connected-pad, repair, and preloaded fanout simplification fixes. Merge current `main`, including the named-net differential-pair connection fix.

Component-owned footprint copper without a source trace remains a fixed obstacle rather than a preloaded route. The solver selector is shared with cache diagnostics so reported and selected pipeline versions agree.

Refresh affected PCB/autorouting snapshots. The BGA36 decoupling-capacitor second phase now has four straight connections and four single-bend connections, no vias, and no DRC errors. Snapshot workflows use Bun 1.4.0, matching normal CI. Align the primary Arduino reroute test's timeout with the suite's 30-second budget.

Validation on the updated branch:
- 71 focused routing/reproduction tests pass, plus two pre-existing skips; the additional AM62/Arduino follow-up passes without snapshot updates.
- Typecheck, package build, and distribution smoke test pass.
- Two CI-only snapshot differences were regenerated on the same Linux runner/runtime as normal tests; snapshot tolerances were not relaxed.
- Full CI at `9d281a26` passed 9/10 test shards; the only failing test was the nRF52810 implicit-pour reproduction. Format, typecheck, and smoke checks passed.
- Per request, temporarily skip that one nRF test in `0c3c1c39`, retaining its assertions and adding a re-enable TODO. Verified locally: one skip, zero failures. Full CI is rerunning on this commit.

Deferred follow-up: the skipped nRF fixture has genuine Pipeline9 via/pad-clearance regressions; skipping the test does not fix the board. A local layer-restriction prototype removed the SWDCLK/battery-pad short but also blocked a valid same-net pad transition, so it was rejected and is not included. A separate small repair03 starvation change passed a synthetic reproduction but did not improve the full nRF board, so it is also not included. The remaining routing fix must preserve same-net pad access while keeping foreign traces out of the battery pad. The crystal trace overlap also occurs with Pipeline7. Re-enable the skipped test after correcting the routing regression.


seveibar:
Updated this PR to published @tscircuit/capacity-autorouter@0.0.884 (from 0.0.879), synced current main (including checks 0.0.183 and the copper-pour update), and refreshed seven SVG snapshots in total.

Results:
- Differential-pair regression is fixed: declared-clearance, length-skew, endpoint, and zero-trace-error assertions pass with the published package. The PCB SVG now shows the intended length-matching meander.
- The BGA test passes, but visual inspection still shows small pad-entry bends in the phase-2/final routing SVG. Passing assertions do not establish that the reported wrinkles are fixed.
- NRF still fails with four pad-pad clearance errors, both locally and on Linux.
- Arduino imported-region rerouting still exceeds its explicit 20-second timeout on Linux (passes locally in about 14 seconds).
- Formatting, type-checking, build, and dist smoke test pass. No assertions or timeouts were relaxed in this update.

The [29-file Linux regression run](https://github.com/tscircuit/core/actions/runs/33989159971) had 27 passes and the two failures above. The [final snapshot refresh](https://github.com/tscircuit/core/actions/runs/33989637855) passed all four selected tests and committed the remaining Linux SVG baselines. The [full suite before the final snapshot commit](https://github.com/tscircuit/core/actions/runs/33989149362) found only those two non-snapshot failures plus the three snapshot mismatches now refreshed.

A new CI run is being triggered for the final snapshot head. The PR is not all green yet.

seveibar:
Follow-up for the skipped nRF regression: https://github.com/tscircuit/tscircuit-autorouter/pull/2424 fixes layer-blind port redistribution. The battery bottom-pad boundary was preventing top-layer ports from being spaced, leaving unrelated endpoints only 0.025 mm apart. With this small fix over 0.0.885, the full SRJ routes and the SWDCLK/battery-pad short is gone. Other DRC failures remain, so the test stays skipped and this PR dependency is unchanged pending upstream merge/release. Independent pad-aware A01/A03 support is in https://github.com/tscircuit/high-density-a01/pull/101; it is not bundled into the spacing fix.

PR #3665, opened 2026-09-05
fix differential pairs connected through named nets

mohan-bee:
### Motivation
- Route differential pairs wired through named nets in their selected autorouting phase.

repro https://github.com/tscircuit/core/pull/3664

### Before

image

- The USB-C breakout fails routing and its snapshot has no copper traces.

### After

image

- The same board snapshot shows routed USB data, CC, power, and ground connections.
- The USB pair routes in phase 0, with no routing or unconnected-port errors.


seveibar (commented):
this looks good but can you use autorouting phase snapshots as well? So that we can see what each phase did visually?

mohan-bee:
> this looks good but can you use autorouting phase snapshots as well? So that we can see what each phase did visually?

image

@seveibar  done https://github.com/tscircuit/core/pull/3665/changes#diff-646420d26dca575b95b9f0107a7da7f26781b15ffd4428bac3e08f5eeb16df54

seveibar (changes requested):
placeholder because this PR is dangerous, where is the TSX?

tests/repros/usb-c-breakout-net-pair.test.tsx
```diff
@@ -1,25 +1,26 @@
 import { expect, test } from "bun:test"
-import { getSimpleRouteJsonFromCircuitJson } from "lib/utils/autorouting/getSimpleRouteJsonFromCircuitJson"
+import { createAutoroutingPhaseIoStack } from "tests/fixtures/create-autorouting-phase-io-stack"
 import { getTestFixture } from "tests/fixtures/get-test-fixture"
 
 import UsbBreakoutBoard from "./usb-c-breakout/board"
```
seveibar: add the TSX to this file, you're introducing new phases that could be implicit phases, that is extremely dangerous. I searched for this board and it was a PITA to find, i don't have a link to the repro PR in your description

overall, difficult to review and dangerous, i need to see what you're doing that you're saying should create a new phase

we DO NOT create implicit phases

lib/utils/autorouting/SimpleRouteJson.ts
```diff
@@ -83,6 +83,8 @@ export type SimpleRouteConnection = {
   name: SrjConnectionName
   routingPcbGroupId?: PcbGroupId
   source_trace_id?: string
+  /** Source traces combined into a net-based routing connection. */
+  source_trace_ids?: SourceTrace["source_trace_id"][]
```
seveibar: remove? why is this needed?

this is exactly what i'm talking about when i say people need to be careful about introducing apis, api boundaries ARE A HUGE DEAL to change, this is a MAJOR API and AI WILL USE IT TO CHEAT

seveibar (approved):
GOOD

PR #3667, opened 2026-09-05
repro: annotations overlap across adjacent LED output branches

GokulPandi-M:
## Issue

When similar LED output chains are laid out next to one another, neighboring resistor and LED branches can be placed in the same narrow region. Their component bodies and generated annotations then overlap, making references, values, terminal labels, and ground connections difficult to read.

The current render reports three component-overlap warnings in the LED output cluster. This is a general repeated-branch layout problem, not a problem specific to one named diode.

Screenshot 2026-09-05 at 8 39 34 PM

## Reproduction

- Preserves the complete TrafficLightController TSX and its original connectivity.
- Adds one schematic snapshot showing the current behavior.
- Uses no manual schematic coordinates, label removal, renamed components, or extra schematic margins.

## Expected behavior

Adjacent LED output branches should be separated using their complete rendered bounds so their component bodies and annotations remain readable.

## Verification

- bun test tests/repros/repro-adjacent-led-branch-annotation-overlap.test.tsx
- bunx tsc --noEmit


imrishabh18 (approved):

GokulPandi-M:
merge please @imrishabh18

Issue #3281, opened 2026-08-18
CopperPour clearance/margin props are dropped from Circuit JSON, so KiCad refills use different rules


0hmX:
## Summary

Authored `<copperpour>` rule props are used by the copper-pour solver but are not preserved on the emitted `pcb_copper_pour` record.

Downstream editable-format exporters therefore cannot reproduce design intent. In a reviewed 4-layer board, the source requested 0.30 mm clearance/margins, but the KiCad project emitted a zone with `(clearance 0.15)`. Refill in KiCad can consequently generate different copper from the reviewed Gerbers.

This is related to, but distinct from, [#2621](https://github.com/tscircuit/core/issues/2621): that issue tracks storing fill fragments instead of the requested region; this issue tracks the missing authored rule metadata.

## Reproduction

```tsx
export default () => (
  <board width="20mm" height="20mm" layers={4}>
    {/* Add ordinary components/traces and a GND connection */}
    <copperpour
      name="GND_POUR"
      layer="inner2"
      connectsTo="net.GND"
      clearance="0.3mm"
      padMargin="0.3mm"
      traceMargin="0.3mm"
      boardEdgeMargin="0.3mm"
      cutoutMargin="0.3mm"
      useThermalReliefs
      coveredWithSolderMask
    />
  </board>
)
```

The solver input correctly receives:

```ts
pad_margin: props.padMargin ?? clearance,
trace_margin: props.traceMargin ?? clearance,
pour_margin: clearance,
board_edge_margin: props.boardEdgeMargin ?? clearance,
cutout_margin: props.cutoutMargin ?? clearance,
```

However, current main inserts only:

```ts
db.pcb_copper_pour.insert({
  shape: "brep",
  layer: props.layer,
  brep_shape: brepShape,
  source_net_id: sourceNetId,
  subcircuit_id: ...,
  covered_with_solder_mask: coveredWithSolderMask,
})
```

The resulting Circuit JSON record has no clearance, pad margin, trace margin, board-edge margin, cutout margin, thermal-relief setting, or source-pour identity.

The KiCad export from the real board therefore contained:

```kicad
(zone
  (layer "In2.Cu")
  ...
  (clearance 0.15)
)
```

despite every authored margin being 0.30 mm.

## Expected behavior

Circuit JSON should preserve enough authored copper-pour intent for editable exporters to reconstruct the same zone rules, independently of the already-computed fill geometry.

[...]

PR #3705, opened 2026-09-07
Preserve effective copper pour rules in Circuit JSON

VykosMolt:
A copper pour can use 0.30 mm clearance in the solver and then lose that value in Circuit JSON. This copies the rules from the region passed to the solver onto each emitted fill fragment: clearance, pad/trace/board-edge/cutout margins and thermal settings.

This is part of #3281. The existing fill geometry stays the same; the region/identity problem in #2621 still needs separate work.

This is a draft until [circuit-json #761](https://github.com/tscircuit/circuit-json/pull/761) is released. Both the devDependency and override need updating then. Local checks used the new schema, so a fresh checkout is still missing these types.

With that schema, the focused and related copper-pour tests pass, including zero values, thermal settings, different margins per layer and split fragments. Existing snapshots stay the same. TypeScript and the package build pass, and the combined core → schema → KiCad check produces `(clearance 0.3)`.

CI at `df93bc8` passed nine test shards. The remaining AM62L routing failure has the same 93 trace-error records on this PR and its base after removing timestamps ([base job](https://github.com/tscircuit/core/actions/runs/34153054519/job/101839111373), [PR job](https://github.com/tscircuit/core/actions/runs/34160224174/job/101860240215)). The type check is waiting on the schema release above.

AI-assisted implementation and automated verification.


VykosMolt:
Continued in #4040 (same commit, rebased); I can't reopen this one after the stale bot closed it.

PR #3712, opened 2026-09-07
fix: honor schematic pin counts consistently

### LINKED ISSUE #2871: schPinArrangement with leftPinCount/rightPinCount creates zero schematic ports

VykosMolt:
The documented `schPinArrangement` fields such as `leftPinCount` and `rightPinCount` currently produce a chip body with no schematic pins. The same counts now feed port creation, total-count inference, box layout and Circuit JSON output, with the deprecated `Size` spelling used as a fallback for each side.

Mixed spellings across sides work, explicit zero stays zero, and explicit pin mappings keep their selected pin numbers.

The original SOIC-8 example goes from 8 source ports and 0 schematic ports to 8 of each. The regression fails on the original code. All 13 affected tests pass with Bun 1.4.0, along with TypeScript, the build, formatting and the existing schematic snapshots.

Local checks used an existing dependency tree with matching declarations and the published `circuit-json@0.0.484` package. CI installed dependencies from scratch and passed type checking, build/smoke checks and nine test shards. The remaining AM62L routing failure has the same 93 trace-error records and normalized assertion diff on this PR and its exact base ([PR run](https://github.com/tscircuit/core/actions/runs/34163832175), [base job](https://github.com/tscircuit/core/actions/runs/34160285144/job/101860430264)).

Fixes #2871.

AI-assisted implementation and automated verification.


VykosMolt:
Continued in #4039 (same commit, rebased); I can't reopen this one after the stale bot closed it.

PR #3713, opened 2026-09-07
Pass trace maximum via counts to autorouter input

Abse2001:
## Summary

- add optional maxViaCount metadata to Simple Route JSON connections
- propagate source_trace.max_via_count across the core-to-autorouter boundary
- extend the trace regression through generated autorouter input

This lets crystal traces, which already receive max_via_count: 0 in core, carry that constraint into routing. A routed crystal trace that violates this limit remains an error.

Related implementation: tscircuit/tscircuit-autorouter#2446

Core already warns when a crystal is modeled as a generic `<chip />` with an X or Y reference designator, recommending the dedicated `<crystal />` component. Impedance matching is intentionally out of scope.

## Validation

- bun test tests/components/primitive-components/trace-max-via-count.test.tsx
- bunx tsc --noEmit
- bun run build
- git diff --check


seveibar (changes requested):
read your prs…

Abse2001:
Closing after design clarification. The released general autorouter does not currently consume a per-connection maxViaCount, and making zero vias a hard global-routing constraint adds disproportionate solver complexity. Crystal traces should instead be explicitly pre-routed with pcbPath; the existing max_via_count=0 check remains an error backstop. Core already warns when an X/Y-designated crystal is modeled as a generic <chip />, recommending <crystal />.

PR #3715, opened 2026-09-07
fix: propagate routing progress and cancellation through circuit wrappers

seveibar:
Fanout routing used a blocking `solve()` call, while a stopped capacity router could resume an outstanding async step and publish stale progress or completion. Fanout now advances in bounded slices; both wrappers guard their run lifecycle, yield to the event loop, and report actual step progress.

Add `renderUntilSettled({ signal })` and `cancelRendering(reason)` so cancellation reaches local routing phases, remote requests/polling, and isolated subcircuits. Aborted phases cannot publish results or start subsequent phases. Isolated circuits forward routing/solver events with `isolatedSubcircuitPath`, preserving local Circuit JSON IDs while distinguishing concurrent render contexts.

Cancellation is cooperative between solver steps. A canceled Circuit is terminal; callers create a new Circuit to restart. Remote cancellation stops client requests and polling, without deleting an already submitted server job. The README documents the API and event scope.

Validation:

- 60 autorouting utility tests passed, including async stop/restart, callback cancellation, progress, and failure regressions.
- 69 root-circuit, subcircuit, and cancellation integration tests passed; one existing test skipped.
- Separate sibling/nested isolated event-scope regression passed.
- Existing fanout, single-layer, plane-termination, sequential-phase, remote-routing, and isolated-render regressions passed.
- `bunx tsc --noEmit`, `bun run build`, formatting, and `git diff --check` passed.

### DIFF
--- a/lib/utils/autorouting/CapacityMeshAutorouter.ts
+++ b/lib/utils/autorouting/CapacityMeshAutorouter.ts
@@ -1,8 +1,8 @@
 import {
-  AutoroutingPipelineSolver,
   AssignableAutoroutingPipeline2,
   AssignableAutoroutingPipeline3,
   AutoroutingPipeline1_OriginalUnravel,
+  AutoroutingPipelineSolver,
   AutoroutingPipelineSolver3_HgPortPointPathing,
   AutoroutingPipelineSolver4,
   AutoroutingPipelineSolver5,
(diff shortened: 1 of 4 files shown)


seveibar:
Closing because this branch name was already used for an older change. I will open a replacement PR from a distinct branch containing the new implementation.

PR #3717, opened 2026-09-07
fix: omit absent fuse voltage rating labels

### LINKED ISSUE #2833: fuse without voltageRating renders a bare unit: "1A / V" on the schematic, "V" in circuit JSON

gilmarGNJ:
Closes #2833.

When `voltageRating` is absent, `parseAndConvertSiUnit` returns a null value. The fuse renderer still appended `V`, producing `1A / V` in the schematic and `display_voltage_rating: "V"` in Circuit JSON.

This omits only the absent voltage label, preserving explicit zero volts, SI-prefixed ratings, and `schShowRatings={false}`. The new visual regression covers all four cases and checks the emitted source fields.

Verification on Bun 1.4.0:

- Reproduced the new test failing before the fix (`display_voltage_rating: "V"`).
- 14 related tests across 9 files pass, including existing fuse ratings, supplier fields, RP2040 schematics and rotated text bounds.
- `tsc --noEmit` and targeted Biome formatting checks pass.
- Inspected the rendered SVG snapshots. The three existing snapshots change only fuse label text and the corresponding two rotated text bounding boxes; unrelated metadata, CSS and equivalent trace serialization were preserved.

AI-assisted contribution, with local reproduction, tests and visual verification. No changes to rating parsing or input props.


gilmarGNJ:
The red `test (5)` check is the same failure already present on this PR's exact base commit, `d696a754932bb0c15fdcb42c4577d28afb175600`:

- [Baseline job](https://github.com/tscircuit/core/actions/runs/34164145085/job/101871744068)
- [This PR's job](https://github.com/tscircuit/core/actions/runs/34168093576/job/101883034994)

Both fail `repro-am62l-lpddr4-progressive-fanout.test.tsx` at `create-am62l-lpddr4-fanout.tsx:2553`, where `pcb_trace_error` is expected to be empty but contains the same 93 trace/via overlap errors (`pcb_trace_error_0` through `_92`). This fixture contains no fuses.

I have not changed that test or updated its snapshots to suppress the failure. This is a baseline comparison, not a claim that the full CI is green.

PR #3718, opened 2026-09-08
fix: render schDisplayLabel on branched traces as inline net labels

### LINKED ISSUE #3521: Trace schDisplayLabel on branched connectivity becomes schematic_net_label

billythompsons:
## Problem

A `schDisplayLabel` on a trace that branches to 3+ ports rendered as anchored `schematic_net_label` endpoints floating detached from the wire, instead of an inline label on the wire like a normal named trace gets.

Repro (also added as a regression test):

```tsx
<trace
  name="V-PLUS-INPUT"
  schDisplayLabel="V+"
  path={[".U1 > .pin1", ".R1 > .pin1", ".R2 > .pin1"]}
/>
```

Before: two anchored `schematic_net_label` elements with "V+". After: the "V+" label sits inline on the routed wire and no anchored net labels are created.

## Root cause

Two pieces had to line up for a multi-pin net to get an inline label:

1. `createSchematicTraceSolverInputProblem` only fed nets to the solver as `netConnections` when they had a named `source_net` (or were same-sheet cross-section connections). A branched connectivity net that was only named by a trace's `schDisplayLabel` never became a `netConnection` - it stayed a set of `directConnections` plus fallback anchored net labels.
2. The solver only performs multi-pin inline-label conversion for `netConnections`, and `applyInlineNetLabelEligibility` gated inline eligibility on explicit port-to-net traces, which a display-labeled branched net doesn't have.

## Fix

- `createSchematicTraceSolverInputProblem`: feed a branched (3+ port) connectivity net with no named `source_net` as a `netConnection` when its canonical name comes from an explicit display label on a real trace. Guarded so that:
  - 2-port display-labeled nets are untouched (they already get one inline label via `directConnections`),
  - nets already fed as cross-section connections are not duplicated,
  - nets whose traces carry *conflicting* display labels keep rendering each label separately, exactly like before.
- `applyInlineNetLabelEligibility`: relax the explicit-port-net-trace gate for exactly that case (branched, no `source_net`, display-labeled). Unnamed nets stay gated, so auto-derived names don't leak onto multi-port wires.

## Verification

- New regression test: `repro-branched-schdisplaylabel-inline-net-label` asserts zero anchored net labels for the label, at least one inline `schematic_text` on the wire, and a routed net.
- Updated snapshots: `example10-schDisplayLabel`, `example11-net-symbol`, `repro128-uncessary-broken-traces` - in each, anchored labels for display-labeled branched nets become inline labels on the wire; routing is otherwise unchanged (verified structurally in the SVG DOM: same trace paths, same component layout).
- Targeted re-run of every inline-net-label / net-label-precedence / subcircuit-label test: all pass, including `repro-tb67s579ftg-inline-net-labels` (label counts unchanged), `multiple-net-labels` (conflicting labels still render separately), and `repro170` (cross-section `netConnections` not duplicated).

Fixes #3521


billythompsons:
Friendly bump on this one - happy to adjust the approach if you had something different in mind for the connectivity label handling.

billythompsons:
Merged current main in and re-verified the repro test locally (1 pass) - branch is fresh against v0.0.1890 now. Still happy to take a different direction here if you have one in mind.

billythompsons:
Follow-up after merging main: repro126's schematic snapshot legitimately changes with this fix. Its circuit has `<trace from="U1.FB" to="R1.pin2" schDisplayLabel="FB" />` on a 3-port branched net (FB pin + R1 + R2) - the exact #3521 case - so the FB label now renders inline instead of vanishing. Updated the snapshot in 26750cc8 with the diff explanation in the commit message. Verified visually: the FB label sits inline on the FB-R1 wire, no overlaps.

billythompsons:
Status update: the branch is still current with the mainline merge, the focused connectivity-label regression test passes, and CI remains green. This is ready for review; I can adjust the approach if maintainers prefer different error-label semantics.

billythompsons:

Issue #3436, opened 2026-08-25
Schematic autorouter physically joins separate same-net netlabels after a local branch trace


Abse2001:
## Reproduction

With `@tscircuit/core@0.0.1752`, `@tscircuit/cli@0.1.2002`, and `tscircuit@0.0.2413`:

```tsx
import "tscircuit"

export default () => (
  <board
    routingDisabled
    schMaxTraceDistance="10mm"
    schTraceAutoLabelEnabled={false}
  >
    <net name="GND" isGroundNet />
    <capacitor
      name="C1"
      capacitance="100nF"
      schX={-2}
      schY={1}
      schOrientation="vertical"
    />
    <capacitor
      name="C2"
      capacitance="100nF"
      schX={-1}
      schY={1}
      schOrientation="vertical"
    />
    <capacitor
      name="C3"
      capacitance="100nF"
      schX={2}
      schY={-1}
      schOrientation="vertical"
    />
    <trace from="C1.pin2" to="C2.pin2" />
    <netlabel
      net="GND"
      connectsTo="C1.pin2"
      schX={-2}
      schY={0.2}
      anchorSide="top"
    />
    <netlabel
      net="GND"
      connectsTo="C3.pin2"
      schX={2}
      schY={-1.8}
      anchorSide="top"
    />
  </board>
)
```

Run:

```sh
bunx tsci snapshot same-net-netlabels-route.circuit.tsx --update --schematic-only
```

## Expected

C1 and C2 share their explicit local trace. The C1/C2 branch and C3 branch each terminate at their own GND netlabel, with no drawn schematic wire between those two netlabel-terminated branches.

## Actual

The schematic renderer inserts an additional long wire from the C1/C2 branch to C3. Both branches still render their GND netlabels, so the added wire is redundant and can cross unrelated signals in a dense schematic.

The issue does not reproduce when each netlabel is attached to only one isolated component. Adding the explicit C1-to-C2 local trace makes the renderer merge the two same-net branches physically. PCB routing is disabled in the repro.

Abse2001:
Retested from tscircuit/ti#129 after upgrading the project runtime:

- `tscircuit 0.0.2446`
- `@tscircuit/core 0.0.1770`
- unique interface boundaries now use native `<trace ... to="net.*" />`
- repeated GND symbols use source-positioned `<netlabel>` elements

[...]

PR #3719, opened 2026-09-08
repro: same-net netlabel branches with a local trace stay separate

billythompsons:
Adds a regression test for #3436: two netlabel-terminated branches on the same net plus an explicit local trace must not be physically joined by an extra schematic wire.

Verified against current main: the render produces exactly three schematic traces (the two short netlabel stubs and the C1-C2 branch) and both GND netlabels - the behavior the issue expects. The original bug was reported against 0.0.1752 and no longer reproduces, so this locks in the correct behavior and the issue can probably be closed.

Refs #3436


billythompsons:
Friendly bump - this is the regression test for #3436. Let me know if you'd like the assertions structured differently.

billythompsons:
Also merged current main and re-ran the regression test locally (1 pass, 3 expect calls). Ready whenever you get to it.

billythompsons:
Status update: the branch remains current, the targeted reproduction for the trace-connection behavior passes, and CI is green. This is ready for review; I can revise the test or implementation if a different behavior is intended.

billythompsons:

PR #3730, opened 2026-09-08
fix: preserve legacy schematic crossings after start contacts

### LINKED ISSUE #3729: Legacy schematic start contact suppresses later crossing markers
The legacy schematic crossing helper skips all later crossings when another perpendicular edge meets the start of the input edge.

At `1fc6c2c400d8ae82875eeabde599dc62f5206883`, this minimal input to `createSchematicTraceCrossingSegments` returns the original edge with **zero** `is_crossing` segments:

```ts
const edges = [{from: {x: 0, y: 0}, to: {x: 10, y: 0}}];
const otherEdges = [0, 4, 7].map(x => ({
  from: {x, y: -1},
  to: {x, y: 1},
}));
```

Expected: preserve the existing no-crossing-at-start behavior at x=0, but still emit crossing segments at x=4 and x=7. Removing only the x=0 edge makes those two crossings appear.

The start contact becomes the nearest candidate. The later coincident-point guard then continues the outer input-edge loop, so the two remaining candidates are never considered. Moving that guard into candidate collection fixes the suppression without changing crossing length or near-end handling.

**Scope:** this is the helper used by the legacy per-trace renderer. The current default `_featureMspSchematicTraceRouting = true` bypasses that renderer. I have not demonstrated an effect in default rendering, and this report does not claim one.

A local regression fails before the change (expected two crossings, got zero) and passes afterward for horizontal/vertical wires in both directions, both candidate orders, preserved endpoints, segment continuity and a start-only contact. A labeled SVG snapshot verifies the resulting wire crossings. All 15 trace tests pass with a 30-second timeout; one existing solver test exceeded the default five-second timeout.

Found and validated with AI assistance. A focused fix is ready for review.

yathanshgaba:
A perpendicular contact at the start of a legacy schematic edge wins nearest-crossing selection, then skips the whole edge. Later crossings disappear. Filter that start contact before selecting the nearest crossing, preserving the existing no-hop-at-start behavior.

This affects the legacy per-trace crossing helper; the current default schematic renderer bypasses it. No default-renderer impact is claimed.

Validation:
- New regression fails on the base (zero crossings instead of two) and passes with the fix.
- Covers both axes and directions, candidate order, start-only contact, endpoints and segment continuity.
- All 15 trace tests pass with `--timeout 30000`. One existing solver test takes about 8.6 seconds and exceeds the default 5-second timeout.
- Labeled schematic SVG rendered and inspected.

AI-assisted implementation and verification, with independent review. No dependency or public API changes.

TypeScript (`bunx tsc --noEmit`), package build (`bun run build`), targeted Biome formatting and `git diff --check` pass.

Fixes #3729.

CI note: test shard 5 fails the existing `repro-am62l-lpddr4-progressive-fanout.test.tsx` DDR routing assertion. The [exact base revision also fails that shard](https://github.com/tscircuit/core/actions/runs/34187827275/job/101939622618). Its complete failure block and [this PR's failure](https://github.com/tscircuit/core/actions/runs/34194044114/job/101957812985) are byte-identical after removing timestamps: 1,730 lines containing 93 unique `pcb_trace_error` entries (IDs 0–92). The excerpt includes a repeated GitHub annotation, which raises the occurrence count to 101. Excerpt SHA-256 `3a66035e2b8107e2e912dd96cb97964a2f28180961f2fa7e0527ae1c52b6224b`. Type checking, the build smoke test and formatting pass.

CI has finished: nine of ten test shards pass, including the new regression in shard 4. Shard 5 has only the verified pre-existing failure described above.


yathanshgaba:
Following up on the stale notice: this regression fix is still awaiting maintainer review. The legacy crossing helper currently lets a perpendicular start contact suppress later crossing markers; the added regression covers both axes/directions and preserves endpoints and continuity. This is explicitly scoped to the legacy helper, not the default schematic renderer.

The PR and base revisions remain unchanged from the recorded checks. The new regression passed in CI shard 4; the remaining DDR failure was also present on the exact base revision, as documented in the PR. Could a maintainer review the fix or indicate whether this legacy path is still accepted for maintenance? I'm available to address specific review feedback.

PR #3733, opened 2026-09-08
fix: parse scientific notation in simulation graph settings

### LINKED ISSUE #3732: Scientific notation is truncated in simulation graph scale and offset settings

```ts
```

yathanshgaba:
Fixes #3732.

String graph settings such as `graphCurrentPerDiv="5e-4A"` currently emit `5 A/div` because the parser stops at the exponent. This change includes an optional signed `e`/`E` exponent in the numeric capture before applying the existing SI prefix. The same value now emits `0.0005 A/div`.

The regression covers initial voltage/current probe records and post-engine graph records, including offset/division calculations with different exponents. A controlled simulation engine isolates graph metadata handling; the SVG snapshot shows the resulting voltage and current scales. Additional parser cases cover signed and uppercase exponents, leading decimals, SI prefixes, ordinary numeric values, and overflow.

Validation:
- Confirmed red integration regression before the parser change: voltage scale `5` instead of `0.5`, current scale `5` instead of `0.0005`, offset `0.2` instead of `2` divisions.
- 11 related simulation/probe/parser tests passed, including existing ngspice graph override tests (58 assertions).
- Final snapshot regression passed after shortening fixture labels; SVG rendered and visually inspected.
- `tsc --noEmit`, ESM/declaration build, formatting and `git diff --check` passed.

The change preserves the existing unit-suffix and partial-input parsing behavior; broader invalid-input validation is outside this fix.

CI completed: 9 of 10 test shards passed, along with typecheck, build smoke test and formatting. Both new regressions explicitly passed in the CI logs (parser in shard 8, integration in shard 9).

Shard 5 fails only the existing `repro-am62l-lpddr4-progressive-fanout` regression. I compared its assertion with the exact base commit's [shard 5 log](https://github.com/tscircuit/core/actions/runs/34187827275/job/101939622618): after removing timestamps, the 1,598-line assertion containing 93 distinct trace errors is byte-for-byte identical. SHA-256: `0d06e7f28d28719d9edd96f066a275369f8a59aebb100306746df95563d3677a`. [This PR's failing shard](https://github.com/tscircuit/core/actions/runs/34196210221/job/101964383409). No DRC or fixture changes are included in this PR.


yathanshgaba:
Following up on the stale notice: this parser fix is still awaiting maintainer review. It preserves scientific notation in graph settings, including the regression where `graphCurrentPerDiv="5e-4A"` was interpreted as 5 A/div instead of 0.0005 A/div. The tests cover parsing plus initial settings and post-engine overrides; the controlled-engine fixture checks graph metadata rather than simulation accuracy.

The PR and base revisions remain unchanged from the recorded checks. Both new regressions passed in CI; the remaining DDR failure matches the exact base revision, as documented in the PR. Could a maintainer review this change? I'm available to address specific feedback.

PR #3739, opened 2026-09-08
fix: parse scientific notation in simulation graph values

### LINKED ISSUE #3732: Scientific notation is truncated in simulation graph scale and offset settings

Humaniacul:
Closes #3732

## Problem

`parseSimulationGraphValue` ends its numeric capture before `e`/`E`, so the exponent is dropped and the remaining characters are treated as a unit. A string graph setting therefore disagrees with the equivalent numeric prop:

```ts
parseSimulationGraphValue("1e-3V") // was 1;   expected 0.001
parseSimulationGraphValue("5e-4A") // was 5;   expected 0.0005
parseSimulationGraphValue("5e2mV") // was 5;   expected 0.5
```

This reaches `volts_per_div` / `amps_per_div` and `display_center_offset_divs` for `graphVoltagePerDiv`, `graphCurrentPerDiv` and `graphVerticalOffset`, via both `VoltageProbe`/`Ammeter` and `getGraphDisplayOverrides`. With `graphVerticalOffset="+1E-3A"` and `graphCurrentPerDiv="5e-4A"` the emitted offset was `0.2` divisions instead of `2`.

## Fix

Consume an optional signed exponent as part of the numeric value, before the SI prefix multiplier is applied:

```diff
-/^([+-]?(?:\d+(?:\.\d*)?|\.\d+))\s*([pnumµkKMG]?)/
+/^([+-]?(?:\d+(?:\.\d*)?|\.\d+)(?:[eE][+-]?\d+)?)\s*([pnumµkKMG]?)/
```

The exponent group requires at least one digit, so a string whose unit merely starts with `e` is not misread as an exponent and keeps its existing behaviour — `"1eV"` still parses as `1`, `"5e"` as `5`. `"5e2mV"` parses as `5e2` then applies `m`, giving `0.5`.

Values that overflow to `Infinity` (`"1e999V"`) now return `undefined` through the pre-existing `Number.isFinite` guard, rather than silently parsing as `1`.

## Tests

- `tests/utils/simulation/parse-simulation-graph-value.test.ts` — unit coverage for the three cases in the issue, exponent + SI prefix combinations, equivalence with the numeric prop, the `"1eV"` / `"5e"` non-exponent cases, and the existing decimal / signed / SI-prefixed / invalid inputs.
- `tests/features/spice-analysis/graph-independent-axes-scientific-notation-overrides.test.tsx` — circuit-level regression against a controlled spice engine, mirroring `graph-independent-axes-numeric-overrides.test.tsx` so scientific-notation strings must emit the same scales and offset divisions as the equivalent numeric props.

Both fail on `main` and pass with the fix; on `main` the integration test reproduces the reported `display_center_offset_divs: 0.2`. `bunx tsc --noEmit`, `biome check` and `bun test tests/features/spice-analysis tests/utils` (127 pass, 0 fail) are clean locally.

### DIFF
--- a/lib/utils/simulation/parseSimulationGraphValue.ts
+++ b/lib/utils/simulation/parseSimulationGraphValue.ts
@@ -18,8 +18,11 @@ export const parseSimulationGraphValue = (
   if (typeof value === "number") return value
 
   const trimmed = value.trim()
+  // The exponent is part of the numeric value, so it must be consumed before
+  // the SI prefix. It requires at least one digit so unit strings that merely
+  // start with "e" (e.g. "1eV") keep falling through to the prefix group.
   const match = trimmed.match(
-    /^([+-]?(?:\d+(?:\.\d*)?|\.\d+))\s*([pnumµkKMG]?)/,
+    /^([+-]?(?:\d+(?:\.\d*)?|\.\d+)(?:[eE][+-]?\d+)?)\s*([pnumµkKMG]?)/,
   )
   if (!match) return undefined
 
--- a/tests/features/spice-analysis/graph-independent-axes-scientific-notation-overrides.test.tsx
+++ b/tests/features/spice-analysis/graph-independent-axes-scientific-notation-overrides.test.tsx
@@ -1,0 +1,107 @@
+import { expect, test } from "bun:test"
+import type { SpiceEngine } from "@tscircuit/props"
+import { getTestFixture } from "tests/fixtures/get-test-fixture"
+import "tests/fixtures/simulation-matcher"
+import { isGraphScopeTrace } from "./isGraphScopeTrace"
+
+const timestamps_ms = Array.from({ length: 4 }, (_, index) => index)
+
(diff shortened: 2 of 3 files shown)


Humaniacul:
The two red shards here are pre-existing on `main`, not caused by this change.

The same run on `main` at `ef4919c3` — the commit this branch is based on — fails the same two shards ([run 34211500333](https://github.com/tscircuit/core/actions/runs/34211500333)), and every `Bun Test` run on `main` today has failed the same way.

The failing tests are both long-running autorouting cases:

- `test (4)` — `repro116: arduino uno circuit json can reroute a center imported region` (74.6s)
- `test (5)` — `routes nine DDR buses including clock, DQS0, DQS1, DMI0, DMI1, and reset with the regular fanout solver` (111.0s)

Neither touches `parseSimulationGraphValue` or the graph display override path. The other 8 shards, `type-check`, `format-check` and `smoke-test` are green.

PR #3745, opened 2026-09-08
fix: update AM62L decoupling layout for fanout solver

ShiboSoftwareDev:
## Summary
- retain @tscircuit/fanout-solver 0.0.66 and relocate the AM62L staged decoupling copper around its compact fanout geometry
- add name-keyed direct-decoupling placement overrides so affected capacitors remain stable and auditable independently of array ordering
- give every fanout via a unique coordinate with at least 0.05 mm copper-edge clearance, then refresh the PCB snapshot

## Root cause
The 0.0.60+ fanout solver geometry is valid, but it occupies several positions that this progressive fixture reserved implicitly for decoupling capacitors and vias rendered in later phases. Once those later elements were materialized, Bun Test node 5 reported 93 trace collisions. This updates the fixture layout to coexist with the current solver instead of rolling it back or weakening DRC assertions.

## Validation
- bunx biome check tests/fixtures/create-am62l-lpddr4-fanout.tsx
- bun test tests/repros/repro-am62l-lpddr4-progressive-fanout.test.tsx
- 11,483 assertions passed on the rebased v0.0.1866 main tree


ShiboSoftwareDev:
Superseded by #3772. The replacement removes the solver-version/component-name override map and updates the fixture's existing source-of-truth placement table directly.

PR #3775, opened 2026-09-08
fix(simulation): parse scientific notation in graph scale/offset string settings

### LINKED ISSUE #3732: Scientific notation is truncated in simulation graph scale and offset settings

billythompsons:
Fixes #3732

## Problem

`parseSimulationGraphValue` silently truncated scientific notation in string graph settings. The numeric capture group stopped before `e`/`E`, so `graphVoltagePerDiv="1e-3V"` emitted `volts_per_div: 1` while the equivalent numeric prop `graphVoltagePerDiv={1e-3}` emitted `0.001`.

This affects `graphVoltagePerDiv`, `graphCurrentPerDiv`, and `graphVerticalOffset` in voltage/current probe records and graph display overrides.

## Fix

Extended the numeric capture with an optional exponent group:

```diff
- /^([+-]?(?:\d+(?:\.\d*)?|\.\d+))\s*([pnumµkKMG]?)/
+ /^([+-]?(?:\d+(?:\.\d*)?|\.\d+)(?:[eE][+-]?\d+)?)\s*([pnumµkKMG]?)/
```

Single shared parser, so all call sites (VoltageProbe, Ammeter, graph display overrides) are fixed at once.

## Tests

New `tests/utils/parse-simulation-graph-value.test.ts`:
- scientific notation: `"1e-3V"`, `"1E-3"`, `"-1e-3V"`, `"2.5e2mA"` (exponent combined with SI prefix)
- parity: string `"1e-3V"` now equals numeric `1e-3`
- regression: SI prefixes (`1mV`, `500uA`), plain numbers, invalid input, `undefined` all unchanged


billythompsons:
CI note: the two failing shards are both long-running autorouter tests timing out (~85-97s), not related to this change:
- `test (5)`: "routes nine DDR buses... regular fanout solver"
- `test (7)`: "repro116: arduino uno circuit json can reroute a center imported region"

The same two failures also appear on the other open PRs for #3732 (e.g. #3739 fails on both), so they look like pre-existing autorouter flakes on current main rather than anything from this PR. All simulation/netlist-path tests pass.

billythompsons:
Still active. The fix for the sci-notation parser in #3732 is unchanged and ready for review; as noted above, the two failing snapshot jobs fail identically on sibling PRs, so they look like a main-branch flake rather than something from this change.

billythompsons:
Status update: the regression coverage remains in place and all current checks are green. This is still ready for review; I can rebase or make any requested changes promptly.

billythompsons:

PR #3786, opened 2026-09-09
repro: scientific notation truncated in simulation graph scale/offset settings

dwdcth:
Refs #3732

`parseSimulationGraphValue` stops its numeric capture before the `e`/`E` of a signed exponent, so string graph settings silently drop the exponent: `"1e-3V"` parses as `1` instead of `0.001`, `"5e-4A"` as `5`, `"+1E-3A"` as `1`.

Adds regression repros pinning the current behavior (same approach as #3782 / #3781):

- parser-level golden assertions for the truncated scientific-notation inputs, with the expected post-fix values noted in comments, plus checks that decimal / SI-prefixed / numeric inputs keep working
- circuit-level assertions that `<voltageprobe graphVoltagePerDiv="1e-3V">` and `<ammeter graphCurrentPerDiv="5e-4A" graphVerticalOffset="+1E-3A">` emit `volts_per_div: 1`, `amps_per_div: 5`, `display_center_offset_divs: 0.2` — these should become `0.001`, `0.0005` and `2` when the parsing is fixed

Both tests pass on current main by pinning the buggy values; they are meant to fail and be updated by whoever fixes the parser.

### DIFF
--- a/tests/repros/repro3732-scientific-notation-truncated-in-graph-settings.test.tsx
+++ b/tests/repros/repro3732-scientific-notation-truncated-in-graph-settings.test.tsx
@@ -1,0 +1,24 @@
+import { expect, test } from "bun:test"
+import { parseSimulationGraphValue } from "lib/utils/simulation/parseSimulationGraphValue"
+
+// https://github.com/tscircuit/core/issues/3732
+// The numeric capture in parseSimulationGraphValue stops before the e/E of a
+// signed exponent, so string graph settings like graphVoltagePerDiv="1e-3V"
+// silently drop the exponent (and any SI prefix after it) instead of becoming
+// 0.001. Numeric props are unaffected.
+
+test("repro3732: scientific notation truncated in simulation graph settings", () => {
+  // Current (buggy) values parsed from scientific-notation strings.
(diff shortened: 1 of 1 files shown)


dwdcth:
The failed `test (3)` job looks unrelated to this change: the failure is in the `spice-model-chip-sends-subcircuit-to-engine` snapshot comparison, while this PR only adds a standalone test file under `tests/repros/` with no source changes. The new test does not run any simulation (assertions read the soup right after the initial render), and it passes locally along with the rest of the suite.

Would someone with permissions mind re-running the failed job? Happy to dig in further if it fails again after a re-run.

dwdcth:
Narrowed this to the parser-level repro only. The earlier version also rendered a powered circuit with probes, and that integration test appears to have perturbed shared simulation-engine state inside the CI shard: the `spice-model-chip-sends-subcircuit-to-engine` snapshot started failing exactly when this file joined the node-3 shard (it passes on main and in every other shard). Dropping the integration test keeps the repro focused and leaves the engine untouched — the parser golden assertions still pin the exact truncated values from the issue.

PR #3792, opened 2026-09-09
Enable configurable PCB via stitching with board props

imrishabh18:
Add board-level opt-in for copper-pour via stitching using @tscircuit/props 0.0.651 (https://github.com/tscircuit/props/pull/843).

```tsx
<board enableViaStitching viaStitchPitch="1mm">
  {/* components and copper pours */}
</board>
```

Stitching runs only when the owning board enables it. Pass the parsed `viaStitchPitch` in millimeters to the solver; omitting pitch preserves its default. The internal `_featurePcbViaStitching = false` switch still disables stitching globally.

Enable stitching in the nRF52810 tracker and four nRF52810-based implicit-pour examples. AM62L boards omit the prop and retain their original unstitched snapshots. The examples verify that stitching vias survive cleanup, belong to poured nets, and span top and bottom.

Keep the nRF52810 regression enabled with `beta_pipeline9`. It tracks the three remaining known trace-via clearance violations at BT1, C1, and L2 after main's duplicate-via fix; extra or changed violations fail. Routing, floating-island, RF-keepout, and GND/VBAT-contact assertions remain active. Its existing raster tolerances remain 0.2% per layer and 0.3% combined.

Refresh compact-beacon and LED-controller snapshots for main's LED fabrication symbols and removal of duplicate routed vias. No snapshot tolerance changes or skipped tests were introduced for these examples.

Validation:
- All four implicit-pour example tests pass without update mode (40 assertions).
- nRF52810 passes with the three exact clearance reports; its snapshots remain unchanged.
- All six copper-pour tests pass, including zero solver starts when enableViaStitching is omitted/false and verification that larger pitch produces fewer vias.
- Eight AM62L snapshots passed after removing stitching opt-in.
- Type checking and formatting pass after syncing current main. Full CI passes on d546f709: all ten Bun test jobs, type checking, formatting, dependency checks, and the distribution smoke test. CI run: https://github.com/tscircuit/core/actions/runs/34705363969.

Pass `viaStitchPitch` directly to the solver and pin the solver commit from https://github.com/tscircuit/via-stitch-solver/pull/8, which removes the old pitch name throughout. The core pitch-control test and typecheck pass for this integration; its new CI run is pending. Solver PR #8 passes all CI checks.


tests/repros/repro-nrf52810-without-copper-pours/repro-nrf52810-without-copper-pours.test.tsx
```diff
@@ -5,8 +5,7 @@ import Nrf52810Circuit from "./nrf52810-circuit"
 
 // Reproduces https://tscircuit.com/seveibar/nrf52810#files without explicit
 // <copperpour> elements so the implicit copper pour phase owns their creation.
-// TODO: Re-enable after fixing Pipeline9's via/pad-clearance regressions.
```
seveibar, line 8: 🤔 are you sure this isnt due to via stitching

seveibar (commented):
I think we should enable this with a board prop, people may want to customize the stitching anyway so we could introduce enableViaStitching and viaStitchPitch?

PR #3802, opened 2026-09-09
Adopt two-pad LED orientation analysis and invalidate cached unknown frames (jlcpcb fabricating polatiry issue)

imrishabh18:
Adopt the published `@tscircuit/circuit-json-util@0.0.113` fix from https://github.com/tscircuit/circuit-json-util/pull/164 in the development dependency and override, preserving the wildcard peer declaration required by this repository. This lets part-orientation analysis populate the local and supplier pin-1 frames for two-pad LEDs/diodes instead of returning unknown.

Bump the supplier-orientation cache namespace from v2 to v3. Previously cached null LED frames otherwise survive the dependency upgrade and suppress the new analysis. A regression test starts with a v2 null entry, checks emitted local/supplier frames for opposite two-pad footprints, and verifies the new cache is reused across renders. Includes a labeled PCB snapshot and updates the existing cache-key expectation.

Validation: five orientation/polarity/cache tests pass; the new regression fails with the old cache namespace; TypeScript and repository-wide Biome formatting pass. The utility is published on npm as 0.0.113.

Release coordination: merge/release this before propagating the new core version through eval and the main tscircuit/browser bundles. The utility-only bumps in those repositories enable fresh analysis, but the core release is needed to invalidate existing platform caches.

Downstream adoption PRs:
- eval: https://github.com/tscircuit/eval/pull/4467
- CLI: https://github.com/tscircuit/cli/pull/4667
- main package: https://github.com/tscircuit/tscircuit/pull/4809
- browser preview: https://github.com/tscircuit/runframe/pull/5047


seveibar (approved):
@imrishabh18 i think we should ADD fabrication notes to any led based on the actual nets, also we could add the diode symbol. Basically we force alignment between the image and what is in code

PR #3861, opened 2026-09-10
test: reproduce silent cross-subcircuit net-to-net connection

0hmX:
https://github.com/user-attachments/assets/a4d0ac91-e983-4fc4-82ca-e18560ea2987

A parent trace from `.LEFT net.BUS` to `.RIGHT net.BUS` silently leaves the two routed child subcircuits physically disconnected. The intended follow-up fix is to throw an error for this case.

Adds only a four-resistor repro and its labeled PCB snapshot. The test has a single PCB snapshot assertion capturing the current disconnected result; it does not assert that net-to-net bridging must be routed. No router or checker implementation changes are included, and error throwing is not implemented in this PR.

Validation: the targeted snapshot test passes (1 test, 1 assertion).

```sh
bun test tests/repros/repro-subcircuit-net-to-net-missing-bridge.test.tsx
```


tests/repros/repro-subcircuit-net-to-net-missing-bridge.test.tsx
```diff
@@ -0,0 +1,64 @@
...
+          resistance="1k"
+          footprint="0402"
+          pcbX={2}
+          schX={1.5}
+        />
+        <net name="BUS" />
+        <trace from=".R1 .pin2" to="net.BUS" />
+        <trace from=".R2 .pin1" to="net.BUS" />
+      </subcircuit>
+      <subcircuit name="RIGHT" pcbX={10} schX={6} autorouter="auto">
```
imrishabh18, line 32: Add `exposedNets` to the subcircuit prop and pass the `BUS`, I think that will work

0hmX: If exposedNets={["BUS"]} is required for this connection, referencing that net from outside the subcircuit should produce a clear validation error. Currently, the connection is accepted but the PCB remains disconnected without an error.

Could we report subcircuit_net_not_exposed, with a message like: “Cannot connect to .LEFT net.BUS: BUS is private to LEFT. Add BUS to the subcircuit’s exposedNets.”

This repro intentionally leaves out that prop to capture the missing diagnostic.

The agent spent time diagnosing this as a routing bug before understanding the net-exposure behavior. A clear error explaining that the net must be exposed would help avoid that confusion.

imrishabh18: yeah, makes sense. Also the `exposedNets ` is not documented anywhere could also be the reason AI couldn't find it

PR #3877, opened 2026-09-11
Reproduce decoupler plane fanout blocked by retained ground

ShiboSoftwareDev:
## Reproduction

Two 0603 decoupling capacitors share 3V3 and GND. C1 first receives a native ground breakout; a subsequent native fanout phase connects both ground pads to the unbroken inner1 ground plane.

The second phase fails with “only 0 of 2 connections could escape”, even though it should be allowed to contact its own retained ground copper and its declared ground plane.

This is a regular TSX circuit using default Pipeline 9, native breakout/fanout, and `toMatchPcbSnapshot`. No custom autorouter, mocked solver, supplied copper, edited SVG, disabled PCB DRC, or fabricated SRJ is used. The test records the current failure so this repro-only PR passes; the stacked fix will require zero native errors and update this same PCB snapshot.

## Validation

- `bun test tests/repros/repro-decoupler-plane-retained-ground.test.tsx --timeout 30000`
- New test formatted with Core-compatible Biome; `git diff --check`.

The source correction belongs in `tscircuit/fanout-solver`; its implementation PR and the stacked Core integration fix will be cross-linked here. This repro is not a fabrication-readiness claim for the larger Linux board.


ShiboSoftwareDev:
Stack is ready: #3878 is based on this repro branch and changes the same ordinary PCB SVG from failed/unrouted to 4 traces, 3 through-vias, and zero native errors. The actual solver correction is tscircuit/fanout-solver#221. Merge this repro first; merge/release the solver fix before the integration PR.

Issue #221, opened 2024-11-01
Bug: `pcbRotation` not impacting output `cad_component` rotation on resistor


seveibar:
![Image](https://github.com/user-attachments/assets/1eda91b1-05f1-467b-aa6d-5485f2a114b1)

https://snippets.tscircuit.com/editor?&snippet_type=package#data:application/gzip;base64,H4sIAOceJGcAA52U226cMBCG73mKka9YKbCwJ2gUoqpsUlVK0hVNT1cVAYeidrFlzKoV4t0znCHsKlHu7PE/3/zjMcR7zoSEHLKUejS8oSEU8CjYHsh7mQbxPKUHGj/4Qhc01P7SkCjDlPvdF2u1Mczd1vvmnc6MtPVBk0xbHjR3YS+tlWGNOR5N41QycVZuXJ/7QYy7MVAEWSznAROUKAr9VyUHLEkl3P53f8ccHFC5YDw9R2bi7+k5pFLESQTFDJxLyBVo9Lvvnonq5/5VUh6QWafzNrWqdacSb0POkC6qgJ8EWIOYtxh7ZExyLCYxYKyMBcGiPceacKwpZ2EYf14k2ROSfcTRixy36ay7apW4dWvDtMDnhm2sMY6rUtfyjex6jGsaHOLqDt+Eu7naNvOpn6RKyggOBhWCykwkoOIS4CISLOOX1Rp35fiaNQAPHn44+aoYBjwmfRmzxCHvjJBGpDu7unNIQqV+WA9iuzoWJfjo2+CnI8LPX++b4PLXsg3PO1f4imovVnHMA/A4MZ289K4jqRhmWqPMn05u6uuRwB4TTbtBjrO0Ja7KKr3LAQTfAmdp1xYkNOobHx8NkiprpyqjybpyWRgBubfR0cCieAbvedXI25ENbE/N5J7dsDC7TZfCD2j1s2hu8vpDAZINCvelTmutUmue0tbHiELtsRYm3I9321eJXUvHNl8rxTvopaNZXsy7z2GmFMoTFJB4HdsFAAA=

imrishabh18:
@seveibar The rotation currently rotates the copper pads of the board, and not the whole 3d-model. Take a look here

[...]

PR #3878, opened 2026-09-11
Fix decoupler plane fanout and update the PCB snapshot

ShiboSoftwareDev:
## Stack

Base / merge first: https://github.com/tscircuit/core/pull/3877

Root implementation: https://github.com/tscircuit/fanout-solver/pull/221

This child PR pins that exact fanout-solver commit and updates the same native PCB SVG introduced by the repro PR. The TSX circuit and its constraints are unchanged.

## Before → after

- Before: ground fanout fails (0/2), with no committed routed traces.
- After: 4 traces, 3 through-vias, zero native errors.
- The regression now requires zero errors, the expected routing counts, and complete through-via spans.

The circuit is two real 0603 decoupling capacitors connected to common 3V3 and ground nets, an ordinary default-Pipeline-9 breakout, and a subsequent native fanout phase to an unbroken inner1 ground plane. No manual copper, custom router, mocked solver, injected SRJ, or postprocessed SVG.

The implementation allows plane drops to contact their declared ground plane and retained same-net copper while retaining foreign-net, physical-pad, and via/drill checks. The integration test uses the normal Core renderer and `toMatchPcbSnapshot`.

## Validation

- Native repro + existing breakout, fanout-bus and unbroken-pour tests: 4 passed / 22 assertions.
- Fixed snapshot regenerated normally and visually inspected; repeat run passes without snapshot updating.
- Source implementation: focused routing/clearance and differential-pair tests pass; full and strict-library TypeScript checks pass.
- New Core test formatted with compatible Biome; `git diff --check` passes.

The dependency is pinned to immutable source for review/CI. It can be changed to the corresponding npm version once fanout-solver #221 is merged and released. Do not merge this integration fix ahead of that implementation.

This proves this real circuit is fixed; it does not claim the complete Linux board is fabrication-ready.


ShiboSoftwareDev:
Solver stack corrected: tscircuit/fanout-solver#222 is now the repro base for tscircuit/fanout-solver#221. Updated dependency pin: 4a54bd18a7e12deb6b592f2556aa003994af5335. Rebased solver tree is identical to the previously green implementation; the native Core repro still passes without changing its solved PCB snapshot.

PR #3882, opened 2026-09-11
repro: automatic layout produces long traces between nearby passives

GokulPandi-M:
## Summary

Adds a faithful repro for excessively long schematic traces produced by automatic layout.

In the current snapshot:

- the shared VCC connection between C4 and C7 takes a page-scale detour
- the shared PREAMP connection between R5 and R6 takes a tall detour even though the resistors are nearby

The surrounding branches are retained because they are part of the automatic packing context that triggers both detours. The repro does not set `schX`/`schY`, add manual trace routes, or include a fix.

## Verification

- `bun test tests/repros/repro186-passive-pair-long-traces.test.tsx`
- `bunx biome check tests/repros/repro186-passive-pair-long-traces.test.tsx`
- `bunx tsc --noEmit`

## Snapshot

automatic layout produces long passive-pair traces


mohan-bee:
this seems to be a trace solver issue,

PR #3886, opened 2026-09-11
Fix nondeterministic PCB IDs from asynchronous footprints

rushabhcodes:
Asynchronous footprint responses can change PCB primitive insertion order. If one footprint finishes and another is still pending, a render pass can allocate pad and silkscreen IDs for the completed footprint first. Reversing the response order then produces different Circuit JSON IDs and SVG bytes for the same board. This caused intermittent AP3429A snapshot failures in tscircuit/sparkfun-boards, including the snapshot job on PR https://github.com/tscircuit/sparkfun-boards/pull/328.

Add `PcbFootprintStringRender` and `FetchPartFootprint` as async dependencies of `PcbPrimitiveRender`, `PcbPortRender`, and `PcbPortAttachment`. All footprint requests still run concurrently. Once they complete, these phases traverse the circuit tree in its existing order, assigning stable primitive IDs and attaching ports only after their pads exist. The existing renderer already waits for footprints before layout; these missing dependencies move the necessary wait ahead of primitive and port creation.

Add a regression test that explicitly resolves two footprint promises in opposite orders and renders between completions. It verifies that all four pads have port links, both runs produce identical pad records and byte-identical PCB SVGs, and the result matches the included visual snapshot. The test reproduced the ID-order mismatch before the fix.

Validation on current upstream main:
- 32 footprint, URL-loading, parts-engine, and render-lifecycle tests passed.
- TypeScript check and package build passed.
- Formatting and `git diff --check` passed.

This fixes the race in core; it does not change the CLI's snapshot comparison rules or replace runtime footprints with local copies.


rushabhcodes:
@vreview

vreview:
### Video review

**Complete**

[▶ Watch the video review](https://0hmx.com/artifacts/video-review-823941374-issue_comment-5635244778-10505dcb/)

Reviewing commit `10505dcb`.

<sub>Job: 823941374-issue_comment-5635244778</sub>

seveibar (commented):
Doesnt this prevent partial rendering of the pcb before async footprints are loaded?

PR #3889, opened 2026-09-11
fix: async-loaded footprint pads missing pcb_port_id

### LINKED ISSUE #3887: Local .kicad_mod footprint creates pads with null pcb_port_id and crashes autorouting
## Problem

A resistor using `footprint="./two-pad.kicad_mod"` produces visible SMT pads, but both imported pads have `pcb_port_id: null`. Routing to that resistor fails with an undefined `connectedPort.x`. The same connection routes successfully with a built-in footprint or JSX converted from the exact same KiCad file.

## Reproduction

Tested on Linux x64 with Bun 1.4.2, `@tscircuit/cli@0.1.2057`, `tscircuit@0.0.2516` (resolving `@tscircuit/core@0.0.1892`), and TypeScript 5.9.3. These reproductions use only public tscircuit packages and synthetic components.

In an empty directory:

```sh
bun init -y
bun add --exact @tscircuit/cli@0.1.2057 tscircuit@0.0.2516 typescript@5.9.3
```

Use this `tsconfig.json`:

```json
{"compilerOptions":{"jsx":"react-jsx","moduleResolution":"Bundler","module":"ESNext","target":"ESNext","skipLibCheck":true}}
```

`two-pad.kicad_mod`:

```scheme
(footprint "TwoPad"
 (version 20240108)
 (generator "pcbnew")
 (layer "F.Cu")
 (attr smd)
 (fp_text reference "REF**" (at 0 -2) (layer "F.SilkS") (effects (font (size 1 1) (thickness 0.15))))
 (pad "1" smd rect (at -0.8 0) (size 0.9 1) (layers "F.Cu" "F.Paste" "F.Mask"))
 (pad "2" smd rect (at 0.8 0) (size 0.9 1) (layers "F.Cu" "F.Paste" "F.Mask"))
)
```

`async-footprint.tsx`:

```tsx
export default () => (
 <board width={12} height={8} >
  <resistor name="R1" resistance="1k" footprint="./two-pad.kicad_mod" pcbX={-3} schX={-2} />
  <resistor name="R2" resistance="1k" footprint="0603" pcbX={3} schX={2} />
  <trace from="R1.pin2" to="R2.pin1" />
 </board>
)
```

```sh
bunx tsci build async-footprint.tsx
```

Inspect `dist/async-footprint/circuit.json`.

## Actual result

- R1's two imported `pcb_smtpad` entries have `port_hints: ["1"]` / `["2"]`, but both have `pcb_port_id: null`.
- The built-in R2 footprint's pads have valid PCB port IDs.
- One source trace is declared; **zero `pcb_trace` entries** are emitted.
- Circuit JSON contains `pcb_trace_missing_error`: `Trace [R1.pin2 to R2.pin1] is not connected (it has no PCB trace)`.
- Build output includes:

```text
Async effect error in PcbTraceRender "autorouting":
TypeError: undefined is not an object (evaluating 'connectedPort.x')
    at getSimpleRouteJsonFromCircuitJson
    at _runLocalAutorouting
```

The CLI currently exits 0 despite this error; that exit-status behavior is tracked in https://github.com/tscircuit/cli/issues/4707.

## Controls / workaround

1. Replace only R1's footprint with `"0603"`: **one routed PCB trace, no circuit errors**.
2. Convert the same synthetic footprint:

```sh
bunx tsci convert two-pad.kicad_mod --name TwoPad --output TwoPad.tsx
```

Import `TwoPad` and use `footprint={TwoPad({ name: "R1" }).props.footprint}` on R1: **one routed PCB trace, no circuit errors**.

## Expected

Asynchronously loaded footprint pads should acquire the same source/PCB port associations as the equivalent inline footprint before routing begins. If association fails, report a useful component/pad diagnostic instead of dereferencing an undefined connected port.

Related but different: #1704 concerned imported footprints inside a group. This reproduction has no group and fails with a direct footprint path on a board child.

billythompsons:
Fixes #3887.

## Problem
A component using an async-loaded `footprint="./file.kicad_mod"` produced visible SMT pads whose `pcb_port_id` stayed `null`, so autorouting crashed dereferencing `connectedPort.x` and the trace was never emitted.

## Root cause
`Port.doInitialPcbPortRender` early-returns when the port has no matched pcb primitives. With async footprint loading, `PcbPortRender` runs before the `.kicad_mod` finishes loading, so the port renders nothing. When the footprint pads are later added and matched during `PortMatching`, nothing re-triggers the port's pcb render, so `pcb_port_id` remains `null` and the pad's `PcbPortAttachment` writes `null`.

## Fix
`Port.registerMatch` now marks `PcbPortRender` dirty when a pcb primitive matches after that phase already initialized. The existing `updatePcbPortRender` retry path (which is a no-op when `pcb_port_id` is already set) then creates the pcb port, and pad attachment picks up the id in the same render cycle.

## Test
New regression test `footprint-kicad-mod-async-pads-port-attachment.test.tsx` reproduces the issue exactly (two-pad `.kicad_mod` loaded through the platform `footprintFileParserMap`, trace to a builtin 0603) and asserts both imported pads get a `pcb_port_id` with zero circuit errors. Fails before the fix, passes after. `tests/footprint` (27 files) all pass.

### DIFF
--- a/lib/components/primitive-components/Port/Port.ts
+++ b/lib/components/primitive-components/Port/Port.ts
@@ -301,6 +301,16 @@ export class Port extends PrimitiveComponent<typeof portProps> {
    */
   registerMatch(component: PrimitiveComponent) {
     this.matchedComponents.push(component)
+    // If this port already rendered without any matched pcb primitives
+    // (e.g. the primitives were added later by an async footprint load),
+    // mark PcbPortRender dirty so updatePcbPortRender retries now that a
+    // match exists.
+    if (
+      component.isPcbPrimitive &&
+      this.renderPhaseStates.PcbPortRender?.initialized
+    ) {
+      this._markDirty("PcbPortRender")
+    }
   }
   getNameAndAliases() {
     const { _parsedProps: props } = this
--- a/tests/footprint/footprint-kicad-mod-async-pads-port-attachment.test.tsx
+++ b/tests/footprint/footprint-kicad-mod-async-pads-port-attachment.test.tsx
@@ -52,7 +52,7 @@ it("assigns pcb_port_id to pads from an async-loaded .kicad_mod footprint (issue
   const r1 = circuit.selectOne("resistor.R1")
   const r1PcbComponentId = r1?.pcb_component_id
   const r1Pads = circuitJson.filter(
-    (e: any) =>
+    (e): e is Extract<typeof e, { type: "pcb_smtpad" }> =>
       e.type === "pcb_smtpad" && e.pcb_component_id === r1PcbComponentId,
   )
 
--- a/tests/footprint/footprint-kicad-mod-async-pads-port-attachment.test.tsx
+++ b/tests/footprint/footprint-kicad-mod-async-pads-port-attachment.test.tsx
@@ -1,0 +1,64 @@
+import { expect, it } from "bun:test"
+import { getTestFixture } from "tests/fixtures/get-test-fixture"
+import { KicadFootprintToCircuitJsonConverter } from "kicad-to-circuit-json"
+
+const twoPadKicadMod = `(footprint "TwoPad"
+ (version 20240108)
+ (generator "pcbnew")
+ (layer "F.Cu")
+ (attr smd)
+ (fp_text reference "REF**" (at 0 -2) (layer "F.SilkS") (effects (font (size 1 1) (thickness 0.15))))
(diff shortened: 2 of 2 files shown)


billythompsons:
Still active and ready for review. CI is fully green and the branch is current with main. The included repro test covers the exact async .kicad_mod two-pad case from #3887 (fails before the fix, passes after).

billythompsons:
Status update: the async kicad_mod pad regression fix and its focused test remain green in CI. This is ready for review; I can rebase or adjust the parsing path if needed.

billythompsons:

PR #3891, opened 2026-09-11
Fix AM62L phased fanout handoff

ShiboSoftwareDev:
## Root fix

This is stacked on #3890 so the PCB snapshot shows the failing and fixed Core behavior directly. The merged fanout-solver 0.0.73 update lives in the base repro PR; this PR contains only the Core root fix and its passing expectations.

- replaces stale phase traces by `pcb_trace_id` when carrying previous fanout output into the next independent phase instead of concatenating duplicate copies
- materializes one physical `pcb_via` when same-net routed branches share the same transition, while keeping electrically unknown vias distinct
- compares physical via fields directly with an explicit board-world millimetre tolerance instead of serializing coordinates
- enables the existing shared-route-via regression as a normal passing test

## Real circuit coverage

The AM62L fixture keeps the 60 real TSX capacitors, 120 explicit through vias, and real component-to-via trace connections from #3890. Its decoupling drops carry their authored `pcbPath` directly; the fixture router only joins true two-endpoint SRJ signal connections and does not synthesize PDN geometry.

The fixed global handoff contains only unique `pcb_trace_id` entries, and the circuit completes with no autorouting, trace, or via-clearance errors. The passing SVG labels that result explicitly. No component positions or SVG geometry were hand-edited.

The changed geometry boundaries document their right-handed board-world frame, axes, units, and point semantics. Test comparisons remain typed geometric values rather than rounded coordinate strings.

## Verification

- `bun test tests/repros/repro-am62l-direct-decoupling-fanout.test.tsx` (1 pass, 12,845 assertions)
- `bun test tests/repros/repro-shared-route-via-duplicates.test.tsx`
- `bun test tests/repros/repro-am62l-lpddr4-bottom-decoupling-packing.test.tsx tests/repros/repro-am62l-lpddr4-paired-fanout-layer-e2e.test.tsx`
- `bunx tsc --noEmit --pretty false`
- `bunx biome check` on every changed TypeScript file
- `git diff --check`


lib/components/primitive-components/Group/Group.ts
```diff
@@ -133,6 +133,59 @@ const getDistanceToPoint = (
...
+/**
+ * Returns whether two vias represent the same physical plated hole.
+ *
+ * `x` and `y` are point coordinates in the right-handed board-world frame,
+ * measured in millimeters (+X right, +Y top, +Z above the board). PCB layers
+ * describe the Z-axis span. These are positions rather than direction vectors,
+ * and no transform is applied.
+ */
+function pcbViasSharePhysicalHole(
+  firstPcbVia: PcbViaPhysicalIdentity,
```
seveibar: Very bad file for this

lib/components/primitive-components/Group/Group.ts
```diff
@@ -2248,7 +2256,15 @@ export class Group<Props extends z.ZodType<any, any, any> = typeof groupProps>
               subcircuit_id: this.subcircuit_id!,
               pcb_group_id: this.pcb_group_id ?? undefined,
               subcircuit_connectivity_map_key: subcircuitConnectivityMapKey,
-            })
+            }
+            if (
+              materializedPcbVias.some((materializedPcbVia) =>
```
seveibar, line 2261: what is a materialized via?

seveibar (approved):

PR #3910, opened 2026-09-12
Skip outer-layer unbroken copper pours as routing obstacles

maci0:
Replaces stale-bot-closed #2601.

Unbroken top/bottom copper pours are conforming fills (filled after routing, flowing around traces). They are not copper at routing time. Emitting them as hard obstacles makes a 2-layer board with a full GND pour unroutable, and also makes the autorouter trace-route the pour net.

Inner-layer unbroken pours stay obstacles so escape vias still have a target, and those nets still route.

Fixes the same issue as #2601.


maci0:
Rebased onto current main (`b21fe203`) and force-pushed. No conflicts.

Retested after rebase: `simple-route-json-unbroken-copper-pour-obstacles` plus the full `tests/utils/autorouting/` suite and `example37-copper-pour-solver-emit` / `example39-unbroken-copper-pour-escape-vias` — 47 passing, 0 failing.

PR #3735, opened 2026-09-08
fix: retain small cutouts in routing obstacles

yathanshgaba:
Fixes #3734.

A 2 mm × 0.2 mm polygon slot and a 0.2 mm-diameter circular cutout currently disappear from the router input: the helpers' 0.6 mm sampling band puts the first midpoint outside each shape. Cap the effective band to the polygon height or circle diameter, then use it consistently for sampling and returned rectangle height. Zero-size shapes remain empty; shapes at least as tall as the configured band retain their existing sampling behavior.

The regression renders actual polygon and circular cutouts and asserts their emitted SRJ obstacles, dimensions, centers, four board layers and empty connectivity. Its PCB snapshot shows the authored shapes; the SRJ assertions establish obstacle preservation. Separate helper cases cover translated thin shapes, boundary sizes, an ordinary larger shape, and empty/zero-size inputs.

Validation:
- Red regressions demonstrated omitted polygon and circular obstacles before their corresponding fixes. The combined integration test failed with two obstacles instead of three before the circle correction.
- 51 related obstacle, autorouting and polygon tests passed (205 assertions).
- Typecheck, ESM/declaration build, formatting and whitespace checks passed.
- Updated snapshot rendered and inspected; independent review found no actionable defect.

This remains the existing scanline approximation. The polygon helper also serves polygon pads and copper-pour boundaries; the patch does not redesign larger-shape final bands or claim a demonstrated routed-trace failure.

CI on `6e1d35016e682a1ae0a7c9459f76a41d5ad76bc5` completed: nine of ten test shards passed, along with typecheck, formatting, build smoke test and Vercel preview. All three new regressions explicitly passed: small-circle helper in shard 5, polygon helper and combined cutout-to-SRJ fixture in shard 6.

The remaining failure is the existing `repro-am62l-lpddr4-progressive-fanout` test. I compared [this PR's shard 5](https://github.com/tscircuit/core/actions/runs/34198755423/job/101972448383) with the [exact base commit's shard 5](https://github.com/tscircuit/core/actions/runs/34187827275/job/101939622618). Removing timestamps yields a byte-identical 1,598-line assertion containing 93 distinct trace errors; SHA-256 `0d06e7f28d28719d9edd96f066a275369f8a59aebb100306746df95563d3677a`.

[...]

PR #3911, opened 2026-09-12
Inflate cutout obstacles by copper-to-edge clearance

maci0:
Replaces stale-bot-closed #2612.

A cutout becomes Edge.Cuts. The autorouter hugs obstacle edges, so a slot emitted at its physical size lets traces fail copper-to-edge DRC. Inflate rect/circle/polygon cutout obstacles by 0.3 mm per side.

Updates the cutout routing snapshot because the detour is now farther from the slot.


maci0:
Rebased onto current main (`b21fe203`, includes #3936's refreshed snapshots) and force-pushed.

- Conflict was only `tests/components/primitive-components/__snapshots__/cutout-routing-pcb.snap.svg`; resolved by taking main's refreshed snapshot and regenerating for this change (0.52% route diff from the inflated cutout obstacle).
- Retested: `cutout-routing`, `cutout-simple-route-json`, `cutout-obstacles`, `cutout`, and `cutout-obstacle-edge-clearance` all pass; `tsc --noEmit` clean; Biome clean.
- Note for review: #3735 ("retain small cutouts in routing obstacles") touches the same obstacle-building path, so these two may want to land together or in a known order.

maci0:
Rebased again onto `1621a01a` (main moved while CI ran). The only conflict was the `cutout-routing-pcb` snapshot again; took main's refreshed version and regenerated for this change. All four cutout tests pass locally.

PR #3915, opened 2026-09-12
Emit per-side via tenting fields

seveibar:
PCB via primitives and imported routed vias now emit `tented_on_top` and `tented_on_bottom` instead of `is_tented`, following https://github.com/tscircuit/circuit-json/pull/788.

Adds optional `tentedOnTop` and `tentedOnBottom` to `PcbVia`, retaining deprecated `isTented` as a fallback. Imports use circuit-json's compatibility transform, so legacy JSON still works and explicit per-side values take precedence, including `false`. Tenting sides follow the same footprint flip as via layers. Updates the circuit-json development dependency and override to 0.0.487, and @tscircuit/props to the latest release, 0.0.653. Peer dependency wildcards remain unchanged.

Validation: 16 focused regression tests pass with @tscircuit/props 0.0.653. TypeScript, ESM/declaration build, formatting, and diff checks pass. Coverage includes direct and imported primitives on both layers at all four quarter-turn rotations, optional/legacy/explicit boolean combinations, and imported routed vias. The new tenting checks assert emitted JSON because PCB SVG snapshots do not expose these flags.

### DIFF
--- a/lib/components/primitive-components/PcbVia.ts
+++ b/lib/components/primitive-components/PcbVia.ts
@@ -11,7 +11,10 @@ import { PrimitiveComponent } from "../base-components/PrimitiveComponent"
 export const pcbViaProps = viaProps
   .extend({
     layers: z.array(layer_ref).optional(),
+    /** @deprecated Use tentedOnTop and tentedOnBottom instead. */
     isTented: z.boolean().optional(),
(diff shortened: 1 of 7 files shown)


seveibar:
CC @techmannih

PR #3919, opened 2026-09-12
Add repro for routed trace overlapping fixed GND label

rushabhcodes:
A manually positioned GND label in the SparkFun VEML7700 schematic is crossed by an automatically routed wire on the same net. The fixed symbol connects to U1.pin3, while the separate U1-to-D1 ground route passes through the GND text. The expected behavior is to preserve the chosen label position and route around its symbol/text bounds.

Add a minimal reproduction containing one generic chip, one capacitor, one LED, and the fixed GND label, plus a schematic SVG snapshot. The fixture removes the original board's connectors, pull-ups, jumper circuitry, footprints, and external assets. PCB rendering is disabled, and the portrait snapshot keeps the entire ground route visible.

This PR records the current faulty output for investigation; it does not change routing behavior. In the generated Circuit JSON, the routed vertical segment is at x = -1.17, only 0.07 mm from the fixed label's x = -1.1 anchor, and crosses its text area. Core currently excludes explicitly positioned labels from automatic label placement, while the solver input's text obstacles are collected from component-associated schematic_text rather than these schematic_net_label records.

Reproduction:

```sh
bun test tests/repros/repro-fixed-gnd-label-trace-overlap.test.tsx
```

Validation:
- Based on freshly fetched upstream main at eacf4d0e (core 0.0.1902).
- Generated and visually inspected the schematic snapshot, confirming the GND text overlap.
- Re-ran the repro test successfully without updating snapshots.
- Biome check and git diff whitespace check pass.


rushabhcodes:
Moved the reproduction to https://github.com/tscircuit/schematic-trace-solver/pull/1197. The replacement includes captured solver input, a standalone test, and a schematic snapshot. It explicitly documents that the fixed GND label is absent from the captured solver input and is rendered as context for the overlap.

PR #3923, opened 2026-09-12
Fix false connection errors at rotated rectangular pads

rushabhcodes:
A PCB route can reach the correct rotated pad but be assigned to a neighboring pin's source trace, causing false missing-connection errors. This affected the CD74HC4067 MUX breakout while fixing [SparkFun boards PR #329](https://github.com/tscircuit/sparkfun-boards/pull/329).

When a routed endpoint has no explicit PCB port ID, `getSourceTraceIdForRoutedTrace` identifies its pad by checking the endpoint's position. The rectangular-pad check ignored `ccw_rotation`, so closely spaced pads rotated by 90° or 270° could incorrectly contain a neighboring pad's endpoint.

Transform the endpoint from board coordinates into the pad's local coordinates before comparing it with the pad's width and height. This makes the source-trace assignment follow the actual pad geometry while retaining the existing allowance for trace width.

Validation:
- The regression directly asserts the assigned `source_trace_id` for 0°, 90°, 180°, and 270° on both PCB layers. It fails before the fix and passes afterward.
- Both the new regression and the existing opposite-layer pad attribution regression pass.
- TypeScript checking passes.
- The full SparkFun MUX renders with zero errors when this fix is combined with the separate power-net corrections in PR #329.


seveibar (changes requested):
extremely confusing to use an unrotated local frame. The rotated rect should be represented by a polygon for that check

PR #3925, opened 2026-09-13
Fix double scaling in PCB text size estimates

Irvinwop:
Silkscreen and copper text size estimates apply `fontSize` twice. For example, five characters at `fontSize={0.5}` report a 1.25 × 0.25 mm box; applying the existing estimate once gives 2.5 × 0.5 mm. This also affects footprint bounds calculated from copper text.

Return the dimensions already scaled to millimeters from both `getPcbSize()` implementations. A regression test exercises the exported component classes at 0.25, 0.5, 1 and 2 mm and checks that width and height scale linearly.

Validation:

- Regression before fix: expected width 1.25 mm, received 0.3125 mm at 0.25 mm font size.
- After fix: 14 focused tests pass, including existing text rendering, multiline, knockout, inherited PCB style and `pcbSx` tests.
- Typecheck and formatting of all three changed files pass.
- Production ESM/declaration build and the built-package smoke test pass.
- All ten CI test shards pass on Bun 1.4.0: **1,564 pass, 42 skip, 0 fail**, across 1,349 files. CI typecheck, built-package smoke test, dependency check and Vercel deployment also pass.

Implemented and tested with Codex (GPT-6); no independent human review is claimed.


Irvinwop:
@rushabhcodes can I get a review on this?

rushabhcodes:
why is there no change in snapshots

Irvinwop:
The existing snapshots don't change because this patch doesn't change the rendered `pcb_silkscreen_text` / `pcb_copper_text` elements. `doInitialPcbPrimitiveRender()` still writes the same `text`, `font_size`, anchor, layer, etc.

The bug is in the pre-layout size estimate: `getPcbSize()` was multiplying `fontSize` twice. That estimate feeds `_getPcbLocalBoundsBeforeLayout()` in `PrimitiveComponent.ts` and the pre-render overlap helpers, so it affects bounds/layout decisions without changing the text element that the current PCB snapshots render.

That's why I added `pcb-text-size-scaling.test.ts`: it exercises the affected contract directly at 0.25, 0.5, 1, and 2 mm and fails on the old implementation (for example, 0.25 mm is estimated as 0.0625 mm high instead of 0.25 mm).

Issue #3903, opened 2026-09-12
Add global board viaTenting prop with per-via overrides, mask rendering and exports


raykholo:
Versions: tscircuit 0.0.2516, @tscircuit/core 0.0.1892, @tscircuit/props 0.0.649, circuit-json 0.0.484, circuit-to-svg 0.0.413, circuit-json-to-gltf 0.0.118, CLI 0.1.2057, Bun 1.4.2, Linux. Native export comparison uses KiCad 10.0.2.

## Request

Please expose via tenting in TSX and carry it consistently into mask-aware SVG/3D rendering and manufacturing exports. Our two-layer board uses local ground vias next to each SMD ground pin and a continuous bottom ground plane. We want both faces of these vias covered by solder mask while connector holes remain exposed.

Primary usage: one global board-level default (proposed API, not currently implemented):

```tsx
<board viaTenting="both">
  {/* All manual and autorouted vias inherit tenting on both faces. */}
  <via name="GND_U1" pcbX={-1.85} pcbY={0}
    holeDiameter="0.3mm" outerDiameter="0.6mm"
    fromLayer="top" toLayer="bottom" connectsTo="net.GND" />
</board>
```

The global setting is required for this workflow: users should not have to repeat it on every via. Suggested values: `"both" | "top" | "bottom" | "none"`. Inheritance should reach nested components/subcircuits and all generated routing vias belonging to this PCB; it should not spill into a separate child PCB.

Per-via overrides for exceptions:

```tsx
<via name="GND_U1" pcbX={-1.85} pcbY={0}
  holeDiameter="0.3mm" outerDiameter="0.6mm"
  fromLayer="top" toLayer="bottom" connectsTo="net.GND"
  tented />
```

`tented` would mean both sides. For asymmetric mask requirements, consider `tented={{ top: true, bottom: false }}`. The board default `<board viaTenting="both">` must also apply to autorouter-generated vias, with explicit per-via overrides. Preserve existing behavior when no setting is supplied.

## Current observations

- The installed ViaProps/viaProps schema has no tenting property.
- Circuit JSON already supports optional `pcb_via.is_tented`; reuse it for the both-sides case rather than introducing a competing boolean. Per-side support would need a backward-compatible schema extension.
- `circuit-to-svg` renders otherwise identical vias with `is_tented: true` and `false` identically even with `showSolderMask: true`. Its `createSvgObjectsFromPcbVia` always emits exposed copper and drill circles.
[...]

PR #3926, opened 2026-09-13
Support board default via tenting with per-via overrides

techmannih:
`<board defaultViaTenting>` stores per-side defaults on `pcb_board`. Vias leave tenting undefined unless explicitly overridden, so `<via tented={false}>` stays exposed while unspecified vias inherit their board's settings.

Preserves imported board defaults and route-via overrides, including `false`, and flips explicit sides with the footprint. Boards include their subcircuit ID so renderers resolve the correct defaults in panels. Updates `circuit-to-svg` to `^0.0.424`.

Adds two focused tests: a labelled panel snapshot for opposite board defaults and explicit via overrides, and a stitching snapshot verifying generated vias inherit the board default. Existing route-import coverage checks explicit tenting alongside via dimensions; two existing board JSON snapshots include the new fields.

Validation: both new tests and the three existing via/import regression tests pass. Related to #3903.


lib/components/primitive-components/CopperPour/CopperPour_doInitialPcbViaStitchRender.ts
```diff
@@ -53,9 +55,17 @@ const renderViaStitchingForCopperPours = (copperPour: CopperPour) => {
...
+    boardComponent?._parsedProps.defaultViaTenting,
+  )
   for (const pcbVia of solver.getOutput().pcbVias) {
-    const { pcb_via_id: _solverViaId, ...pcbViaInput } = pcbVia
-    const insertedVia = db.pcb_via.insert(pcbViaInput)
+    const { pcb_via_id: _solverViaId, ...pcbViaInput } = pcb_via.parse(pcbVia)
+    const insertedVia = db.pcb_via.insert({
+      ...pcbViaInput,
+      tented_on_top: boardTenting.tented_on_top ?? pcbViaInput.tented_on_top,
+      tented_on_bottom:
```
seveibar, line 66: these properties are optional, we can inherit from board (i.e. leave undefined) unless the via specifically has the prop defined. This makes

seveibar: it's fine for now

seveibar: actually nvm, we need it

seveibar (changes requested):
undefined unless overridden, we may need to adjust the pcb_board spec to have the default behavior

if we merged this as-is, pcb_trace.route vias would be unspecified or unable to be customized. That needs to be in the spec for handling

If not in a pcb_board, then pcb_trace.route would need to be able to have the specification for it

imrishabh18 (approved):
Take a review from @seveibar as well

seveibar (approved):

PR #3932, opened 2026-09-13
repro: same-net VREF branches form a redundant parallel route

GokulPandi-M:
## Summary

- Captures two near-parallel branches of the same VREF net forming a redundant narrow loop.
- Adds a focused geometry assertion and the full A4 schematic snapshot.

Screenshot 2026-09-16 at 12 30 50 AM

## Reproduced behavior

The source connects each passive to VREF once, but the generated schematic contains two vertical VREF segments roughly 0.0245 schematic units apart over about 1.97 units. The resulting double line is redundant but electrically remains on the same net.


imrishabh18 (commented):
I think you might be using 2 traces for this, or defined the connection multiple times. Can you verify this (look at the circuit json maybe)

GokulPandi-M:
> I think you might be using 2 traces for this, or defined the connection multiple times. Can you verify this (look at the circuit json maybe)

i checked the Circuit JSON, the connections are not defined twice, so i think the double trace is being generated by the schematic router iinw @imrishabh18

imrishabh18 (approved):

Abse2001 (approved):

PR #3935, opened 2026-09-13
fix: re-validate missing-trace warnings after async footprint load (#4442)

### LINKED ISSUE #4442: source_pin_missing_trace_warning emitted for ports connected by source_trace in the same build
`tsci build` with tscircuit 0.0.2306 on:

```tsx
import React from "react"
import "tscircuit"

export default () => (
  <board width="30mm" height="20mm">
    <switch name="SW1" type="spdt" footprint="kicad:Button_Switch_SMD/SW_SPDT_PCM12" noConnect={["pin3"]} />
    <trace from=".SW1 > .pin1" to=".SW1 > .pin2" />
  </board>
)
```

Built `circuit.json` contains warnings for ports that the same artifact records as trace-connected:

```
WARNING on pin1(source_port_0) -> connected via trace: true
WARNING on pin2(source_port_2) -> connected via trace: true
```

Expected: `source_pin_missing_trace_warning` only for ports absent from `source_trace.connected_source_port_ids`. The check appears to run before trace connections register; warnings are never re-validated.

marcos452652258-gif:
## Summary

Fixes #4442.

When a component loads its footprint asynchronously (via URL), the ports and traces may be recreated after the initial render. This fix ensures that stale `source_pin_missing_trace_warning` entries are cleared before re-running the design rule checks, so warnings reflect the final connectivity state instead of an intermediate pre-load state.

Includes:
- Clear stale warnings in `NormalComponent_doInitialSourceDesignRuleChecks`
- Re-run `SourceTraceRender` and `SourceDesignRuleChecks` after async footprint load
- Test: verify warnings are re-emitted for legitimately unconnected pins after async load

Closes tscircuit/tscircuit#4442.


marcos452652258-gif:

PR #3946, opened 2026-09-14
feat: support board routeRemaining while retaining unrouted DRC

seveibar:
`<board routeRemaining={false}>` now skips implicit routing of unassigned connections while retaining explicit autorouting phases and breakouts. Omitted or `true` preserves existing routing behavior. The setting is inherited by nested subcircuits.

Source connectivity and routing DRC remain active: connections left unrouted produce `pcb_port_not_connected_error` entries. Unnumbered connection-targeted phases filter their input, and reroute-only phases do not synthesize an initial catch-all route when remaining routing is disabled.

Updates `@tscircuit/props` to `^0.0.660` (tscircuit/props#853). Regression coverage includes nine scenarios: disabled, enabled, omitted, no phases, unnumbered phases, nested subcircuits, targeted phases, trace phase indexes, and breakouts, with one annotated 3×3 PCB snapshot grid showing routed counts, connectivity, pin numbers, and actual DRC messages, plus assertions identifying the unconnected pins.

Validation: 34 targeted routing and DRC tests passed; TypeScript checking, formatting of changed files, and package build passed.


seveibar:
@ShiboSoftwareDev

seveibar:
@ShiboSoftwareDev fwiw this is going to require extensive testing w/ AI to see if it's effective etc. i think the implementation is not the hardest part haha

i will update the skill etc. after this gets merged

PR #3960, opened 2026-09-15
Pin canonical net connectivity fix

ShiboSoftwareDev:
Core currently uses connectivity-map 0.0.30, where a merged canonical connectivity-net ID is not recognized as connected to its member traces. This pins [circuit-json-to-connectivity-map#46](https://github.com/tscircuit/circuit-json-to-connectivity-map/pull/46) at commit `b17b50f5794b992dfcfa120d07cca7344fbb8052` so the T113 canonical-net fix can be integration-tested in core before release. The pinned commit includes its generated package artifacts and preserves the exported class's structural type, so the immutable Git dependency resolves without install scripts or downstream type incompatibilities.

Validation:

- `bunx tsc --noEmit`
- `bun run build`
- `bunx biome format .`
- upstream connectivity-map: 19 tests pass
- GitHub CI: all checks passed on the previous pin; the refreshed pin retains identical runtime behavior and removes a declaration-only compatibility regression
- core `bun test`: 1,566 pass, 41 skip, 16 fail across 1,623 tests
- controlled run on unchanged `origin/main` reproduced all 16 failures: nine host/test-timeout failures and seven existing snapshot mismatches (four via-stitch snapshots at 0.08–0.14%, two AM62 snapshots at 0.00%, and one Arduino snapshot at 1.16%)


ShiboSoftwareDev:
Superseded by tscircuit/tscircuit-autorouter#2581. The clean fix stays inside Pipeline9 and keeps circuit-json-to-connectivity-map at its released version, so core no longer needs this Git commit pin. Closing the integration pin to avoid carrying the broader connectivity semantic change.

PR #3963, opened 2026-09-15
fix: report empty connections targets instead of crashing the render

### LINKED ISSUE #2865: Empty string in `connections` crashes the render with a raw css-what parser error

Itachi3355:
Fixes #2865

## Problem

An empty or whitespace-only value in `connections` reached `selectOne` as a selector, where `css-what` threw:

```
error: Expected name, found .
```

Two problems, as described in the issue:

1. The exception aborted the entire render, so one bad entry produced no circuit JSON at all — sibling connections, other components and every downstream phase never ran.
2. The message named neither the component nor the pin, and pointed at `css-what` internals. The natural reading is that the *valid* selector is malformed.

## Fix

Empty targets are skipped when collecting nets (`_getNetsFromConnectionsProp`) and when creating traces (`_createTracesFromConnectionsProp`), so the rest of the board renders. The misconfiguration is recorded as a `source_component_misconfigured_error` naming the component and pin:

```
<capacitor#13 name=".C1" /> has an empty connections target for pin "pin1". Provide a selector such as ".R1 > .pin1" or "net.VCC", or remove the entry.
```

The error is emitted in `doInitialSourceComponentPropertyValidation` rather than at trace-creation time, because `CreateTracesFromProps` runs before `SourceRender` and `source_component_id` does not exist yet.

Scope, following the issue author's notes:

- Only empty and whitespace-only targets are treated as misconfigured. A non-empty but unresolvable selector (`".R9 > .pin1"` where R9 doesn't exist) already has its own downstream handling and is untouched.
- The check covers both the string and array forms of a `connections` entry.
- Kept in core rather than rejecting `""` at the props-schema level in `tscircuit/props`, which would be a cross-repo change with a wider blast radius.

## Verification

- 2 new tests in `tests/components/normal-components/empty-connections-target.test.tsx`. Both fail on clean `main` with the exact reported `Expected name, found .` crash, and pass with the fix.
- Ran the 310 test files that reference `connections`: 308 pass / 4 fail. The same 4 fail on clean `main` with this change stashed — all autorouting/breakout/fanout, unrelated to this path. No failure is unique to this change.
- `bunx tsc --noEmit` exits 0; biome format clean.

One unrelated flake noticed while establishing the baseline: `tests/features/autoroutingphase-connection-phase-index.test.tsx` failed 2 of 3 runs on clean `main` with this change stashed.


Itachi3355:
Merged current `main` (33 commits) and re-verified — the two regression tests still pass, and there were no upstream changes to `NormalComponent.ts`, so no conflicts. CI is green.

Not stale from my side; this is waiting on maintainer review. Happy to rebase, split, or adjust the error message/wording if you would prefer a different shape.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

PR #3972, opened 2026-09-16
fix: preserve trace-name validation when routing is disabled

techmannih:
## Fix

Traces with the same name on separate electrical nets now retain their existing name-conflict diagnostic when PCB routing is disabled. Previously `Group.doInitialPcbDesignRuleChecks()` returned before validating names for platform/board `routingDisabled` and disabled legacy autorouters.

Run the existing Group name validation before the routing guards. Keep the legacy-router guard next to standalone routing checks; their existing helper already handles `routingDisabled`. The production diff remains seven lines in `Group.ts`, preserving name scoping and the exception for mutually connected same-named traces.

## Regression tests

Stacked on #3971, which contains every explicit scenario and its fixture. This layer changes the three disabled-case assertions to require the existing diagnostic and updates their snapshots. All setup, snapshot options, routing-enabled coverage, valid-name controls, and router/cache tests come from the repro layer unchanged.

Only the error-text overlay changes in the three snapshots; PCB geometry and ratsnests are identical. Legacy missing-copper diagnostics remain present. No new error message or error type is introduced.

Related to https://github.com/tscircuit/tscircuit/issues/4948. Native stack: `main` → #3971 → this fix.

## Validation

- The desired diagnostic assertions fail in all three disabled cases before the production fix.
- Focused repro, group, name-conflict, legacy-router, platform/board routing-disabled, nested-routing, and standalone DRC coverage: **37 pass, 1 existing skip, 0 fail** across 38 test files, with a 30-second test timeout.
- `tsc --noEmit` and Biome checks pass.
- Updated PCB snapshots rendered and visually inspected; an SVG comparison confirms only the error-text overlay changed.


tests/repros/repro4948-routing-disabled-duplicate-names.test.tsx
```diff
@@ -1,15 +1,14 @@
...
+test("routingDisabled preserves duplicate component name diagnostics", async () => {
+  const diagnosticCounts = []
 
-beforeAll(async () => {
   for (const mode of [
     "routing_enabled",
     "platform_disabled",
     "board_disabled",
+    "legacy_disabled",
   ]) {
```
imrishabh18: Wrong pattern, make the tests explicit for each of the case. Very confusing to review

imrishabh18 (changes requested):

tests/repros/__snapshots__/repro4948-trace-name-board-disabled-pcb.snap.svg
imrishabh18, line 1: How is this beneficial? If the board has routingDisabled then why to confuse the AI with this error when It's not needed to route? I don't think this is a valid error from the AI

imrishabh18 (approved):

techmannih:
@vreview

vreview:
### Video review

**Complete**

[▶ Watch the video review](https://0hmx.com/artifacts/video-review-823941374-issue_comment-5714062801-54549447/)

Reviewing commit `54549447`.

<sub>Job: 823941374-issue_comment-5714062801</sub>

nazasnow:
Hi @techmannih, thank you for the review and feedback! I will update the PR to address these points shortly.

nazasnow:
Hi @techmannih, thank you for the review and feedback! I will update the PR to address these points shortly.

PR #3984, opened 2026-09-16
Fix schematic outline color for strict SVG consumers

### LINKED ISSUE #3982: SCHEMATIC_COMPONENT_OUTLINE_COLOR "rgba(132, 0, 0)" is not portable SVG color; strict SVG renderers reject schematic output
`lib/utils/constants.ts` defines

```ts
export const SCHEMATIC_COMPONENT_OUTLINE_COLOR = "rgba(132, 0, 0)"
```

and it is the default color for `SchematicLine`, `SchematicArc`, `SchematicRect`, `SchematicCircle`, the connector inner symbol, and port stems drawn with `schStemLength`. It reaches circuit JSON and `circuit-to-svg` output as `rgba(132, 0, 0)`.

Three-argument `rgba()` is valid in CSS Color Level 4, so browsers render it. It is not valid in SVG 1.1 / CSS2 color syntax, which strict SVG consumers implement. CairoSVG, for example, fails on the generated schematic SVG.

The usual workaround in a library, drawing the stem yourself with a valid color instead of using `schStemLength`, has a cost. The extra line enlarges the symbol's content bounds, which changes how `userCoordinateToResizedSymbolTransformMat` maps the symbol into its declared width/height, and that moves every port. We did exactly that for pin/contact symbols in @tsci/adom-inc.library. The shifted ports then came out disconnected in native KiCad schematic exports (via circuit-json-to-kicad): 4 of 23 contacts lost their net. So the color can't safely be fixed downstream.

**Suggested fix:** `rgb(132, 0, 0)`. It's the same color and valid in every CSS and SVG level.

Checked against main today (`lib/utils/constants.ts` line 44). Installed: @tscircuit/core 0.0.1914, tscircuit 0.0.2562.

billythompsons:
Fixes #3982.

Changes the default schematic component outline color from three-argument `rgba(132, 0, 0)` to the equivalent `rgb(132, 0, 0)`. The new spelling is accepted by SVG 1.1/CSS2 consumers while preserving the rendered color in browsers.

Also extends the existing `SchematicLine` test to assert the portable color reaches circuit JSON.

Testing:
- `git diff --check`
- Focused Bun test could not be run locally because Bun is not installed in this execution environment; CI will run the repository suite.


billythompsons:
Status update: the portable SVG color change and focused circuit-JSON regression remain green across the current CI suite. This is ready for review and I can rebase or adjust the test if needed.

PR #3988, opened 2026-09-17
repro: inner-layer via routes lose source trace attribution

techmannih:
Inner-layer routes between manually placed through-vias lose their source trace attribution when the autorouter supplies endpoint PCB port IDs only in `connectsTo`.

The deterministic fixture renders separate GND and SIGNAL routes on `inner1` through the normal autorouter API. Their logical via ports are on `top`, so geometry matching cannot recover the source IDs. This repro explicitly asserts the current bug: both emitted routes have an undefined `source_trace_id`. The stacked fix in #3991 changes that expectation to each route's exact source trace ID.

The snapshot labels are derived from the emitted source IDs, making the missing association visible without changing the copper geometry.

Repro: both source traces are unassigned

Validation: the repro and opposite-layer attribution tests pass. Running the fixed branch's correct-ID assertion against unfixed core fails with expected `source_trace_0`, received `undefined`. No skipped or expected-failure tests are introduced. Synced with main, which includes the upstream empty-board check fix in `@tscircuit/checks@0.0.198`.


techmannih:
@vreview

vreview:
### Video review

**Complete**

[▶ Watch the video review](https://0hmx.com/artifacts/video-review-823941374-issue_comment-5708586545-d618a4e6/)

Reviewing commit `d618a4e6`.

<sub>Job: 823941374-issue_comment-5708586545</sub>

PR #3991, opened 2026-09-17
fix: preserve inner-layer via route net identities

techmannih:
## Motivation

Preserve correct net attribution when routing between through-vias on inner PCB layers.

## Issue

The autorouter provides endpoint PCB port IDs in `connectsTo`, but core ignores them during source-trace attribution.

When via ports are registered on `top` and the route runs on `inner1`, geometry matching cannot resolve the endpoints, leaving `source_trace_id` undefined.

## Fix

Resolve endpoint ports from `connectsTo` before falling back to geometry matching. Existing route-point port IDs retain precedence.

The regression test verifies that GND and SIGNAL routes retain their correct source IDs. Updated snapshots explain the routing scenario and show attribution changing from `UNASSIGNED` to the correct net names.

The regression now requires each emitted route's exact source trace ID. Snapshot labels derive from those IDs:

| Repro | Fixed |
| --- | --- |
| Missing source traces | Correct GND and SIGNAL source traces |

Validation:

- The correct-ID assertion fails on unfixed core and passes here.
- 12 targeted tests across 11 files pass, covering via/net connectivity, opposite-layer pads, shared vias, subcircuits, breakout attribution, and the previously failing schematic-image test.
- Typecheck, package build, distribution smoke test, Biome formatting, and `git diff --check` pass.
- Synced with main's upstream empty-board check fix in `@tscircuit/checks@0.0.198`.


seveibar (changes requested):
idk what is happening with this implementation CC @Abse2001 might be able to help you figure it out but this makes no sense

seveibar (changes requested):
you should probably revert the repro until we confirm there's an actual bug- i definitely don't know wtf is happening and the test is not clear- revert the repro unless you can confirm there's an actual bug

techmannih:
@vreview

vreview:
### Video review

**Failed — the review could not finish. Tag the bot in a new comment to retry. The operator can inspect the job logs.**

Reviewing commit `06d24d00`.

<sub>Job: 823941374-issue_comment-5708668147</sub>

techmannih:
@vreview

vreview:
### Video review

**Failed — the review could not finish. Tag the bot in a new comment to retry. The operator can inspect the job logs.**

Reviewing commit `06d24d00`.

<sub>Job: 823941374-issue_comment-5709105705</sub>

vreview:
ran out of codex credit

PR #4010, opened 2026-09-18
fix(schematic): make SCHEMATIC_COMPONENT_OUTLINE_COLOR portable SVG color (#3982)

### LINKED ISSUE #3982: SCHEMATIC_COMPONENT_OUTLINE_COLOR "rgba(132, 0, 0)" is not portable SVG color; strict SVG renderers reject schematic output

PINYOPATTANAWASANPORN:
### Summary of Changes
- Replaces 3-argument `rgba(132, 0, 0)` with standard `rgb(132, 0, 0)` in `SCHEMATIC_COMPONENT_OUTLINE_COLOR`.

### Root Cause / Technical Context
Fixes #3982. In `lib/utils/constants.ts`, `SCHEMATIC_COMPONENT_OUTLINE_COLOR` was defined as `"rgba(132, 0, 0)"`.
While 3-argument `rgba()` is supported in modern browsers under CSS Color Level 4, it is invalid under SVG 1.1 / CSS2 color specifications.
Strict SVG consumers and renderers (such as CairoSVG or downstream conversion pipelines) fail when parsing generated schematic SVGs containing this color string.

Replacing it with `"rgb(132, 0, 0)"` produces the exact same color value while remaining compliant with all SVG and CSS specifications.

### Verification & Testing
- Verified `SCHEMATIC_COMPONENT_OUTLINE_COLOR` constant exports valid SVG color string `rgb(132, 0, 0)`.
- No visual or bounding box regressions introduced.


PINYOPATTANAWASANPORN:
Friendly ping on this PR. All CI checks (	est (1..10), 	ype-check, ormat-check, smoke-test) are green and passing. The change remains a minimal, surgical 1-line standard SVG color compliance fix for #3982. Ready for review when convenient.

PR #4018, opened 2026-09-19
fix(hole): preserve name property in pcb_hole circuit-json (#4415)

### LINKED ISSUE #4415: Holes can't carry a name or a courtyard through circuit-json
A mounting hole cannot carry two things a PCB design needs it to carry: a name, and a courtyard. Both gaps are in circuit-json rather than in any one exporter, so every consumer downstream has to either invent the missing data or drop it.

## 1. `<hole name="H1">` is accepted and silently dropped

`holeProps` in `@tscircuit/props` declares `name?: string` for all four hole shapes, and `<hole name="H1" .../>` renders without a warning. But `Hole.ts` never reads `props.name`, and `pcb_hole` has no `name` field to put it in, so the name disappears between the TSX and the circuit-json.

```tsx
<board width="50mm" height="50mm">
  <hole pcbX={-15} pcbY={0} diameter="3.2mm" name="H9" />
</board>
```

```json
{"type":"pcb_hole","pcb_hole_id":"pcb_hole_0","pcb_component_id":null,"hole_shape":"circle","hole_diameter":3.2,"x":-15,"y":0}
```

This matters because KiCad requires a reference designator on every footprint. Its Specctra DSN export — the one the freerouting plugin runs — refuses the whole board when it finds one missing, with `Footprint with value of '' has an empty reference designator`. Since the design cannot supply a name, the exporter has to make one up; tscircuit/circuit-json-to-kicad#416 does exactly that, handing out `H1`, `H2`… and skipping designators real components have claimed. That is the right stopgap, but the designator should be authorable, and it should be the same one in every export rather than one each exporter invents independently.

## 2. A courtyard cannot be attached to a hole

`pcb_courtyard_circle` (and `_rect`, `_outline`, `_polygon`, `_pill`) require `pcb_component_id`. A board-level `<hole>` has no `pcb_component`, so there is no way to say "this courtyard belongs to this hole".

`<courtyardcircle>` does exist as a primitive and can be placed at board level, but it inserts against its own schema:

```tsx
<courtyardcircle pcbX={-15} pcbY={0} radius={3.45} layer="top" />
```

```json
{"type":"pcb_courtyard_circle","pcb_courtyard_circle_id":"pcb_courtyard_circle_0","pcb_component_id":null,"layer":"top","center":{"x":-15,"y":0},"radius":3.45}
```

That `null` has two visible consequences today.

`circuit-json-to-kicad` drops the element entirely — it only ever looks courtyards up per `pcb_component_id` — so authoring one produces nothing in the `.kicad_pcb`.

`@tscircuit/checks` groups courtyards by `pcb_component_id` before comparing them, so every board-level courtyard lands in one bucket keyed `null`. Two board-level courtyards that overlap each other are never compared, and one that overlaps a component reports as:

```
pcb_courtyard_overlap_error  Courtyard of null overlaps with courtyard of U1
```

The practical effect is that mounting holes are invisible to courtyard DRC, and there is no way to fix that from the design. KiCad's own `MountingHole` footprints ring the hole with an `F.CrtYd` circle at `hole diameter + 0.25mm` (`MountingHole_3.2mm_M3` uses 3.45, `MountingHole_2.1mm` uses 2.35), so a KiCad-native board has this and a tscircuit-exported one does not.

An exporter could synthesize that circle, but it shouldn't: it would mean deciding that a Ø3.2 hole is an M3 screw with a head, and a cable pass-through has exactly the same geometry. Courtyards are design intent and belong in the design.

## Proposal

Small and additive, nothing existing breaks:

- add `name?: string` to the `pcb_hole` shape schemas in circuit-json, and pass `props.name` through in `Hole.ts`
- give courtyard elements a way to reference a hole — either an optional `pcb_hole_id`, or relaxing `pcb_component_id` to optional and letting a courtyard stand alone, which would at least stop the current schema violation
- consumers then honour whichever lands: `circuit-json-to-kicad` prefers `hole.name` for the reference designator and puts a hole-targeted courtyard inside that hole's footprint (which is what makes it DRC-active in KiCad — board-level graphics are not footprint courtyards), and `@tscircuit/checks` groups by the hole instead of collapsing into `null`

## Alternative considered

Making a board-level `<hole>` a real component with a `source_component` and `pcb_component` would fix the name, the courtyard and the BOM question through one mechanism. It looks considerably heavier: `<hole>` has to stay a primitive inside `<footprint>` — that pattern is used throughout the USB-C footprints and covered by `hole-in-footprint.test.tsx` — so it would need to behave differently by context, and a new source ftype ripples into refdes assignment, BOM, netlist, the schematic, `createComponentsFromCircuitJson`, `inflateStandalonePcbPrimitives` and autorouter obstacles. Happy to go that way instead if that is the preferred direction — it is a data-model call rather than a bug fix, which is why this is an issue rather than a PR.

## Related

- https://github.com/tscircuit/circuit-json-to-kicad/pull/416 — gives standalone holes and pads an invented reference designator, the stopgap for gap 1
- https://github.com/tscircuit/circuit-json-to-kicad/issues/227 and https://github.com/tscircuit/circuit-json-to-kicad/issues/51 — the same missing-footprint-identity failure on other paths
- https://github.com/tscircuit/tscircuit/issues/3301 — Specctra DSN export producing an unusable file, tscircuit's own DSN path

marcos452652258-gif:
## Summary

Fixes tscircuit/tscircuit#4415.

`<hole name="H1" diameter="3.2mm" />` accepts `name` in `@tscircuit/props`, but `Hole.ts` was not passing `props.name` to `db.pcb_hole.insert()`, and `createComponentsFromCircuitJson.ts` was not passing `elm.name` back when inflating hole primitives.

This caused mounting hole reference designators to disappear between JSX and circuit JSON, resulting in downstream export issues (e.g. KiCad DSN exporter rejecting empty reference designators).

### Changes
- Pass `name: props.name` to `db.pcb_hole.insert()` across all hole shapes (`circle`, `rect`, `pill`, `rotated_pill`, `oval`) in `Hole.ts`
- Pass `name: elm.name` when creating `Hole` instances in `createComponentsFromCircuitJson.ts`
- Add unit test in `tests/components/primitive-components/hole-name.test.tsx` verifying that `name` is preserved in `pcb_hole` circuit JSON


marcos452652258-gif:

PR #4024, opened 2026-09-19
fix(pcb-pack): propagate pack-applied rotation to pcb_plated_hole and pcb_hole

### LINKED ISSUE #4021: pack-applied rotation does not propagate to pcb_plated_hole.ccw_rotation (hand-written pcbRotation does)
## Summary

When `pack` rotates a component by 90°, its SMT pads and ports transform correctly, but its `pcb_plated_hole` elements keep `ccw_rotation: 0`. Their oriented extents are therefore wrong, and a pad that clears the hole in the real footprint is reported as overlapping it (`pcb_pad_pad_clearance_error`, `clearance: 0mm`). Rotating the **same** footprint by the **same** 90° with a hand-written `pcbRotation={90}` propagates correctly and produces no error.

Reproduced on `@tscircuit/core` 0.0.1792 + `@tscircuit/eval` 0.0.1297, and unchanged on `@tscircuit/core` 0.0.1933 + `@tscircuit/eval` 0.0.1416.

## Evidence: same footprint, same angle, three paths

The footprint is a 16-pin USB-C receptacle (LCSC C2765186-class, easyeda-generated) with 4 pill-shaped plated holes (`outer 1.2 × 2.0 mm`, `hole 0.8 × 1.6 mm`) and 12 SMT pads.

| path | `pcb_component.rotation` | plated holes' `ccw_rotation` | clearance errors |
|---|---|---|---|
| hand-written `pcbRotation={0}`, no pack | 0 | `[0, 0, 0, 0]` | 0 |
| hand-written `pcbRotation={90}`, no pack | 90 | `[90, 90, 90, 90]` | 0 |
| **pack-applied rotation** (5-block board, `<board pack gap="1mm">`) | 90 | **`[0, 0, 0, 0]`** | **2** |

Reported by the pack path:

```
pcb_pad_pad_clearance_error — Pads pcb_port[.J1 > .A1B12] and pcb_plated_hole[#pcb_plated_hole_2] are too close (clearance: 0mm, minimum: 0.1mm)
pcb_pad_pad_clearance_error — Pads pcb_port[.J1 > .B1A12] and pcb_plated_hole[#pcb_plated_hole_1] are too close (clearance: 0mm, minimum: 0.1mm)
```

(0.0.1416 reports the same pair as `clearance: -0.159mm`.) All four plated holes are owned by `J1` itself (checked via `pcb_component_id`), so this is not another part being packed into it.

## Why the geometry genuinely collides in the pack case

From the emitted circuit JSON, component at `rotation: 90`:

- pad `A1B12` (`shape: "polygon"`), AABB `x ∈ [-4.249, -2.949]`, `y ∈ [-9.069, -8.469]`
- plated hole at `(-2.831, -9.894)`, `outer 1.2 × 2.0`, `ccw_rotation: 0` → AABB `x ∈ [-3.431, -2.231]`, `y ∈ [-10.894, -8.894]`
- both axes intersect (x by 0.48 mm, y by 0.18 mm)

Had the hole been rotated with its component (`2.0 × 1.2`), its AABB would be `y ∈ [-10.494, -9.294]`, which no longer meets the pad. So the DRC reports faithfully; the defect is upstream of it, in how pack applies rotation to plated holes.

## Impact

Not specific to one footprint. Pack rotates 11–12 components on every board size we tried (40×30, 46×36, 50×40, 56×46 mm; gap 0.8–2 mm), so any part that receives ±90° and carries a non-circular plated hole has the latent defect; it only surfaces when a pad sits close enough to a hole. 180° is unaffected.

## What does not work around it

- a different USB-C part: pack rotates that one to −90° with the same result
- upgrading to 0.0.1933 / 0.0.1416: unchanged
- `pcbRotation={0}` on the component: pack overrides it (patched and control outputs identical)
- a larger board or a different `gap`: all combinations above still rotate the part 90°
- making the plated hole's outer square: the check still fires, since the needed fix is the axis swap, not a larger outline

## Notes for reproducing

A pack-free minimal case does **not** reproduce, because the hand-written rotation path is correct. The defect needs pack to be the one rotating the part, which in our experience takes a board with enough components (ours has 5 blocks, ~20 parts). Two traps cost us several false negatives:

1. The pad/hole clearance check only runs on pads that have a `pcb_port`, i.e. pins that are actually connected. A part placed with no `connections` shows no error even when its geometry overlaps.
2. A fixture board without `pack` raises a layout error that suppresses the clearance errors behind it; with only 2–3 parts pack may not rotate anything.

Happy to share the full board source if useful.

meanusarcanus:
## Description
Fixes #4021

### Root Cause
When `calculate-packing` calculates layout packing and rotates components (e.g. 90°), `applyPackOutput.ts` called `transformPCBElements()`, which transforms coordinates `(x, y)` and component rotation, but did not update the angular rotation fields (`ccw_rotation`, `hole_ccw_rotation`, `rect_ccw_rotation`) on `pcb_plated_hole` and `pcb_hole`. As a result, non-circular plated holes (such as pill and oval holes) remained at `ccw_rotation: 0`, causing their oriented bounding extents to misalign with rotated SMT pads and triggering false-positive `pcb_pad_pad_clearance_error` DRC violations.

### Solution
- Introduced `updatePcbElementsRotation(elements, rotationDegrees)` in `applyPackOutput.ts`.
- Updates `ccw_rotation`, `hole_ccw_rotation`, and `rect_ccw_rotation` on `pcb_plated_hole` and `pcb_hole` across cluster members, individual components, and packed groups.
- Added comprehensive unit tests in `tests/features/pcb-pack-layout/pack-plated-hole-rotation.test.tsx` verifying rotation propagation for pill plated holes and rotated pill holes with rect pads.

@algora-pbc /claim #4021


meanusarcanus:
### 💰 Payout Details
- **PayPal**: `tednadres@yahoo.com`

meanusarcanus:
Still active and ready for review.

PR #4037, opened 2026-09-20
fix(pcb_trace): preserve matching source_trace_id for multi-layer routed traces

### LINKED ISSUE #3019: pcb_trace.source_trace_id can point at an unrelated source_trace, contradicting its own pcb_trace_id
## Summary

On a dense autorouted board a small number of `pcb_trace` elements come back with a
`source_trace_id` belonging to an entirely different connection. The element
contradicts itself: `pcb_trace_id` is formed as `<source_trace_id>_<n>`, and the id
embedded there disagrees with the `source_trace_id` field on the same object.

## Observed

2 of 161 routed traces, on tscircuit `0.0.2237`:

| `pcb_trace_id` | id embedded in it | `source_trace_id` field |
| --- | --- | --- |
| `source_trace_88_0` | `source_trace_88` | `source_trace_79` |
| `source_trace_87_0` | `source_trace_87` | `source_trace_72` |

The route geometry matches the id **embedded in `pcb_trace_id`** — the two endpoints
land exactly on the pads of `source_trace_88` / `source_trace_87` — so it is the
`source_trace_id` *field* that is wrong. The offset is not constant (−9 and −15), so
this is not a simple off-by-N.

Both affected connections have the same shape: a two-pin part on one side of the board
wired to a two-pin part mounted on the opposite side, so the route needs a layer change.
Every other trace in the same design satisfies the invariant.

## Knock-on: vias inherit the wrong net

`pcb_via.subcircuit_connectivity_map_key` is resolved through that bad
`source_trace_id`, so the vias sitting on those two routes are tagged with the *other*
connection's connectivity net:

| via | parent `pcb_trace` | via's key | parent trace's real net |
| --- | --- | --- | --- |
| A | `source_trace_88_0` | `…_net34` | `…_net43` |
| B | `source_trace_87_0` | `…_net30` | `…_net42` |

`…_net34` and `…_net30` are precisely the nets of the mis-referenced source traces, which
is what ties the two symptoms to one root cause.

## Why it matters

Anything that derives net identity from `source_trace_id` — clearance/DRC checking,
netlist export, trace highlighting — attributes copper to the wrong net. A trace-to-trace
clearance checker written the obvious way reports phantom shorts between a trace and its
own pads (I saw ~500 false positives from these two traces alone, because every segment
of the mislabelled route is compared against the pads it legitimately terminates on), and
correspondingly can miss real ones.

`connectsTo` appears to be trustworthy: on all 161 traces every port in `connectsTo`
resolves to a single `subcircuit_connectivity_map_key`. It is currently the only reliable
way to get a routed trace's net.

## Invariant check

```js
for (const t of db.filter((e) => e.type === "pcb_trace")) {
  const embedded = t.pcb_trace_id.replace(/_\d+$/, "")
  if (embedded.startsWith("source_trace") && t.source_trace_id && embedded !== t.source_trace_id) {
    console.log("mismatch", t.pcb_trace_id, "->", t.source_trace_id)
  }
}
```

Same check for vias:

```js
const netOf = (t) => portNet.get(t.connectsTo?.[0]) // via source_port.subcircuit_connectivity_map_key
for (const v of db.filter((e) => e.type === "pcb_via")) {
  const parent = db.find((e) => e.pcb_trace_id === v.pcb_trace_id && e.type === "pcb_trace")
  if (parent && v.subcircuit_connectivity_map_key !== netOf(parent)) console.log("via net mismatch", v.pcb_via_id)
}
```

## Notes

Deterministic — the same input produces byte-identical routes across runs, and the
mismatch reproduces every time. I was not able to shrink it to a small fixture: it does
not appear on a small synthetic board (15 traces, 2 layers), and a scaled-up fixture
failed to route at all so produced no traces to check. Happy to run further diagnostics
on the design where it does reproduce.

meanusarcanus:
## Summary

Fixes #3019

Resolves an issue where `pcb_trace.source_trace_id` could be assigned to an unrelated source trace on multi-layer or dense autorouted boards.

### Root Cause
1. `getEndpointSourcePortIdsFromGeometry` previously inspected `[route[0], route[route.length - 1]]`. When routes transition layers or start/end with non-wire route types, or when via endpoints are encountered, geometry matching failed to extract both port endpoints.
2. When fallback net candidate selection was triggered, `candidates[0]` was selected arbitrarily even when `pcb_trace.pcb_trace_id` already embedded the exact originating `source_trace_id` (e.g. `source_trace_88_0` -> `source_trace_88`).

### Solution
- Use `getWireRouteEndpoints(trace)` inside `getEndpointSourcePortIdsFromGeometry` so valid wire endpoints are properly checked against pads/ports.
- If `pcb_trace_id` embeds a known `source_trace_id`, prioritize or disambiguate candidate source traces with that ID in `getSourceTraceIdForRoutedTrace`.
- Add test verifying `pcb_trace.source_trace_id` consistency with embedded ID and correct via connectivity map key attribution.

@algora-pbc /claim #3019

Payout PayPal: \`tednadres@yahoo.com\`


meanusarcanus:
Bounty / Payout details:

PayPal: tednadres@yahoo.com
Author: meanusarcanus@gmail.com

meanusarcanus:
Still active and waiting for review.

